PR #1, opened 2017-12-01
Use custom user agent string at all times

laurenyu:
The previous implementation allowed for our user agent string to be overridden if the user supplied a boto3 session, SageMaker client, or SageMaker runtime client.

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -17,11 +17,12 @@ import re
 
 import os
 import pkg_resources
+import platform
 import sys
 import time
 
 import boto3
-import botocore.session
+import botocore
 import json
 import six
 from botocore.exceptions import ClientError
@@ -29,8 +30,11 @@ from botocore.exceptions import ClientError
 from sagemaker.utils import name_from_image
 import sagemaker.logs
 
-SDK_VERSION = pkg_resources.require('sagemaker')[0].version
 
+SDK_VERSION = pkg_resources.require('sagemaker')[0].version
+OS_NAME = platform.system()
+OS_VERSION = platform.release()
+PYTHON_VERSION = '{}.{}.{}'.format(sys.version_info.major, sys.version_info.minor, sys.version_info.micro)
 
 logging.basicConfig()
 LOGGER = logging.getLogger('sagemaker')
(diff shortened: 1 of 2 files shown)


src/sagemaker/session.py
```diff
@@ -17,20 +17,24 @@
...
 import six
 from botocore.exceptions import ClientError
 
 from sagemaker.utils import name_from_image
 import sagemaker.logs
 
-SDK_VERSION = pkg_resources.require('sagemaker')[0].version
 
+SDK_VERSION = pkg_resources.require('sagemaker')[0].version
+OS_NAME = platform.system()
```
owenataws: If no OS name can be resolved, ``platform.system()`` will return an empty string. Can this be something like:

``OS_NAME = platform.system() or 'UnresolvedOS'``

Same for ``OS_VERSION``

src/sagemaker/session.py
```diff
@@ -71,30 +75,20 @@ def __init__(self, boto_session=None, sagemaker_client=None, sagemaker_runtime_c
...
+        user_agent_string = 'AWS-SageMaker-Python-SDK/{} Python/{} {}/{} Boto3/{} Botocore/{}'\
+            .format(SDK_VERSION, PYTHON_VERSION, OS_NAME, OS_VERSION, boto3.__version__, botocore.__version__)
 
-        # Some information is appended automatically, so this creates a user agent string that looks like
-        # 'AWS-SageMaker-Python-SDK/1.0/1.7.45 Python/3.6.2 Darwin/15.6.0 Boto3/1.4.7 Botocore/1.7.45'
-        botocore_session.user_agent_name = 'AWS-SageMaker-Python-SDK/{}'.format(SDK_VERSION)
-        botocore_session.user_agent_extra = 'Boto3/{} Botocore/{}'.format(boto3.__version__,
-                                                                          botocore_session.user_agent_version)
+        self.sagemaker_client = sagemaker_client or self.boto_session.client('sagemaker')
+        self.sagemaker_client._client_config.user_agent = user_agent_string
```
owenataws: Is there a reason we don't preserve the original user agent string as well. Why not just append the SageMaker string to the end?

laurenyu: The default user agent string is 'Boto3/1.4.8 Python/3.6.3 Darwin/15.6.0 Botocore/1.8.6', which is why I just added the Boto and Botocore versions to the end. It needs to start with 'AWS-SageMaker' and be followed by 'language/version OS/version' for parsing. Appending the original user agent string here is a good call, though.

src/sagemaker/session.py
laurenyu: done. Also moved everything to a new module.

PR #9, opened 2017-12-06
Improving endpoint deletion

mvsusp:
- created ```timeout_and_delete_endpoint_by_name``` context manager to easily delete an endpoint in the end of the integ test
- set boto3 and botocore logging to INFO for integ tests so is easier to read log results and less noisier
-  added teamcity-messages module so we can se better tests stats


tests/integ/test_cifar.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jesterhazy: rename file to test_tf_cifar.py

PR #10, opened 2017-12-06
describing training job call every 30 seconds

mvsusp:
This change is required until we start using CloudWatch events to determine the training job status. It is allows the streaming to keep happening every 5 seconds but describing training job calls only every 30 seconds.

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -605,7 +605,7 @@ class Session(object):
         # Notes:
         # - The JOB_COMPLETE state forces us to do an extra pause and read any items that got to Cloudwatch after
         #   the job was marked complete.
-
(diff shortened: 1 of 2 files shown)


src/sagemaker/session.py
```diff
@@ -645,8 +645,12 @@ def logs_for_job(self, job_name, wait=False, poll=5):  # noqa: C901 - suppress c
 
             if state == LogState.JOB_COMPLETE:
                 state = LogState.COMPLETE
+            elif seconds_since_last_describe_job_call < 30:
+                seconds_since_last_describe_job_call += max(poll, 1)
```
jesterhazy: why not use actual seconds (from time.time()) instead of this pseudo-time counter?

PR #11, opened 2017-12-07
increase endpoint creation timeout to 20 minutes

mvsusp:


iquintero:
On the message:
increased -> increase 

so that it is in imperative mood 👍

laurenyu:
I thought previously we had decided that a timeout of 15 minutes should be plenty. Did something change?

jesterhazy:
@laurenyu it's a balance. we want to set the timeout so that tests that are going to fail anyway don't jam up the build queue, but we don't want to fail tests that will succeed because our timeouts are too low. i think @mvsusp has found that the current timeouts are causing too many artificial failures.

laurenyu:
Of course, I understand tradeoff. I just want to make sure we're not unnecessarily increasing the timeout - e.g. do _all_ of the tests need to have increased timeouts? was 20 minutes based on running the tests without a timeout or just a small increase while we continue experimenting? I didn't see any explanation in this PR, so I wanted to ask if there was a previous conversation I had missed.

mvsusp:
@laurenyu I increased only the timeouts for endpoints creation. My (small) sample of experiments demonstrated that one of the possible issues with 15 minutes timeouts is when the **endpoint delete operation** happens while the **endpoint creation operation** is still happening. That results in a flaky error and an endpoint that was not deleted.

Now that we have statistics in the integration tests, we can check which tests are slower and refactor them as necessary.

laurenyu:
cool, thanks for the explanation!

PR #30, opened 2017-12-24
Added: print out billable seconds after training completes

djarpin:
The time it takes to run `.fit()` includes provisioning hardware, downloading the container, downloading data, running the algorithm, and saving outputs.  Customer's are not charged for provisioning hardware, which can be a substantial portion of training.  To provide better transparency here, it would be nice to print billable seconds after the "===== Job Complete =====" message.

I'll provide additional context offline.


laurenyu:
two of the unit tests failed with this error:

```
        if wait:
            self._check_job_status(job_name, description)
            if dot:
                print()
            print('===== Job Complete =====')
            # Customers are not billed for hardware provisioning, so billable time is less than total time
>           billable_time = (description['TrainingEndTime'] - description['TrainingStartTime']) * instance_count
E           KeyError: 'TrainingEndTime'
```

PR #44, opened 2018-01-16
Update TensorFlow examples following API change

nkconnor:
It appears at some point the function definition for TensorFlow was changed and any code using these previous examples will break (the SageMaker environment appears to auto-update.. so assumably that code "is broken").

The current function signature def is

` def __init__(self, training_steps=None, evaluation_steps=None, checkpoint_path=None, py_version="py2", **kwargs):`


owen-t:
Thanks for your submission!

PR #45, opened 2018-01-16
Support multi-part uploads

jbencook:
For large datasets, the current `Session.upload_data` method fails. This PR switches the call to `Object.upload_file` which can do multi-part uploads. Also updated the unit tests.


.gitignore
```diff
@@ -20,3 +20,4 @@ examples/tensorflow/distributed_mnist/data
 doc/_build
 **/.DS_Store
 venv/
+*~
```
owen-t: Can this make it into the commit message as well.

owen-t:
Thanks for your submission!

.gitignore
jbencook: Yeah I accidentally added an emacs file on our fork. You mean you want a new commit in this PR with a message about that change?

owen-t: It's a good addition, happy for it to be part of this PR. If we leave it in this PR, can the commit message just reference the addition. Something like

"Upload files to S3 using multipart uploads.

Emacs temporary files covered in .gitignore"

jbencook: Yeah I'm happy to add it. Any idea if amending my commit message and force pushing to our fork will do it?

owen-t: Yeah, should be fine

jbencook: Ok that worked

PR #55, opened 2018-01-23
Add image_name option in MxNet constructor

tianweiz07:
Customers can specify the container image for training MxNet


src/sagemaker/mxnet/estimator.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, entry_point, source_dir=None, hyperparameters=None, py_versio
         """
         super(MXNet, self).__init__(entry_point, source_dir, hyperparameters, **kwargs)
         self.py_version = py_version
+        self.image_name = image_name
```
owen-t: Please add an API doc comment for this argument.

Please rename to 'image' from 'image_name'

tianweiz07: Add API doc comment and rename image name

Issue #4, opened 2017-12-04
Asynchronous fit


defreez:
My first use for SageMaker involved async submission of KMeans jobs. EstimatorBase.fit raises NotImplemented('Asynchronous fit not available') if the wait parameter is set to false.

Is fit() necessarily synchronous, or is it just not implemented _yet_?

winstonaws:
Hey, thanks for trying out SageMaker!

We're going to support async fit, it just didn't make the cut for launch. This is on our backlog, and we'll give you an update when we have an ETA.

defreez:
That's great to hear. Thanks!

iquintero:
This is already implemented in master. However its not released to pypi yet. I will reopen this issue to track it and close it once this feature lands in pypi. Anyone interested in using it now can install the sdk from source.

iquintero:
This was just released in v1.1.0 to PyPI

PR #59, opened 2018-01-25
Add support for async fit()

iquintero:
when calling fit(wait=False) it will return immediately. The training
job will carry on even if the process exits. by using attach() the
estimator can be retrieved by providing the training job name.

This fixes: https://github.com/aws/sagemaker-python-sdk/issues/4


iquintero:
Note:

I have done a big refactor of how attach() works. This is the main change here, everything else is mostly tests.

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -64,6 +64,20 @@ def data_location(self, data_location):
             data_location = data_location + '/'
         self._data_location = data_location
 
+    @classmethod
+    def _from_training_job(cls, init_params, hyperparameters, image, sagemaker_session):
```
mvsusp: Can you add Docstrings to this method?

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -64,6 +64,20 @@ def data_location(self, data_location):
...
+
+        # The hyperparam names may not be the same as the class attribute that holds them,
+        # for instance: local_lloyd_init_method is called local_init_method. We need to map these
+        # and pass the correct name to the constructor.
+
+        for attribute, value in cls.__dict__.items():
+            if isinstance(value, hp):
```
mvsusp: Accessing __dict__ to find a class attribute shows that ```_from_training_job``` should not be a class method but an instance method.

src/sagemaker/estimator.py
```diff
@@ -152,8 +152,46 @@ def fit(self, inputs, wait=True, logs=True, job_name=None):
...
+    @classmethod
+    def attach(cls, training_job_name, sagemaker_session=None):
+        """Attach to an existing training job.
+
+        Create an Estimator bound to an existing training job. After attaching, if
+        the training job has a Complete status, it can be ``deploy()`` ed to create
+        a SageMaker Endpoint and return a ``Predictor``.
+
+        If the training job is in progress, attach will block and display log messages
+        from the training job, until the training job completes.
```
mvsusp: Can you add some code examples here?

src/sagemaker/estimator.py
```diff
@@ -152,8 +152,46 @@ def fit(self, inputs, wait=True, logs=True, job_name=None):
...
+
+        Args:
+            training_job_name (str): The name of the training job to attach to.
+            sagemaker_session (sagemaker.session.Session): Session object which manages interactions with
+                Amazon SageMaker APIs and any other AWS services needed. If not specified, the estimator creates one
+                using the default AWS configuration chain.
+            **kwargs: Additional kwargs passed to the :class:`~sagemaker.estimator.Estimator` constructor.
```
mvsusp: kwargs is not in the method signature

src/sagemaker/estimator.py
```diff
@@ -152,8 +152,46 @@ def fit(self, inputs, wait=True, logs=True, job_name=None):
...
+
+        Returns:
+            sagemaker.estimator.Framework: ``Estimator`` with the attached training job.
+        """
+        sagemaker_session = sagemaker_session or Session()
+
+        if training_job_name is not None:
```
mvsusp:
```
if not training_job_name
```

src/sagemaker/estimator.py
```diff
@@ -528,56 +566,24 @@ def hyperparameters(self):
...
+        framework_init_params['source_dir'] = json.loads(hyperparameters.get(DIR_PARAM_NAME))
+        framework_init_params['enable_cloudwatch_metrics'] = json.loads(
+            hyperparameters.get(CLOUDWATCH_METRICS_PARAM_NAME))
+        framework_init_params['container_log_level'] = json.loads(
+            hyperparameters.get(CONTAINER_LOG_LEVEL_PARAM_NAME))
 
         # drop json and remove other SageMaker specific additions
-        hyperparameters = {entry: json.loads(hp[entry]) for entry in hp}
-        framework_init_params['hyperparameters'] = hyperparameters
+        hp_map = {entry: json.loads(hyperparameters[entry]) for entry in hyperparameters}
```
mvsusp: hp is a map already. Maybe rename to deseriealized_hps or something in this lines?
My following up question is why is this method responsible to deserialize the hps?

tests/integ/test_factorization_machines.py
```diff
@@ -53,3 +54,45 @@ def test_factorization_machines():
...
+
+        # training labels must be 'float32'
+        fm.fit(fm.record_set(train_set[0][:200], train_set[1][:200].astype('float32')), wait=False)
+        training_job_name = fm.latest_training_job.name
+
+        print("Detached from training job. Will re-attach in 20 seconds")
+        time.sleep(20)
+        print("attaching now...")
+
+    with timeout_and_delete_endpoint_by_name(endpoint_name, sagemaker_session, minutes=35):
```
mvsusp: 35 minutes is too long. Can we make a test under 15 minutes?

tests/integ/test_kmeans.py
```diff
@@ -60,3 +61,49 @@ def test_kmeans():
...
+        kmeans.center_factor = 1
+
+        kmeans.fit(kmeans.record_set(train_set[0][:100]), wait=False)
+        training_job_name = kmeans.latest_training_job.name
+
+        print("Detached from training job. Will re-attach in 20 seconds")
+        time.sleep(20)
+        print("attaching now...")
+
+    with timeout_and_delete_endpoint_by_name(endpoint_name, sagemaker_session, minutes=35):
```
mvsusp: Too long

tests/integ/test_factorization_machines.py
iquintero: The test itself is taking the same time as the synchronous fit one, the difference is this timeout is accounting for fit() + deploy(). If you notice the timeout is consistent with the synchronous tests that we have.

The difference is the call to

with timeout_and_delete_endpoint_by_name() is spending a lot of time on attach()  this time is usually spent in fit() in the synchronous tests. In this case the fit() returns right away so it doesn't account for much of the runtime. So the test takes the same amount of time to complete.

src/sagemaker/amazon/amazon_estimator.py
iquintero: will do

src/sagemaker/estimator.py
iquintero: will get rid of it. It used to be and I forgot to get rid of it here.

src/sagemaker/estimator.py
iquintero: will do

src/sagemaker/estimator.py
iquintero: will add an example.

src/sagemaker/estimator.py
iquintero: I will rename it :D
- this was already done by attach(). I don't think its worth to move this deserialization to its own method.

src/sagemaker/amazon/amazon_estimator.py
owen-t: I think this is fine here, hyperparameters are implemented as descriptor objects.

src/sagemaker/estimator.py
```diff
@@ -152,8 +152,46 @@ def fit(self, inputs, wait=True, logs=True, job_name=None):
         self.latest_training_job = _TrainingJob.start_new(self, inputs)
         if wait:
             self.latest_training_job.wait(logs=logs)
+
+    @classmethod
+    def _from_training_job(cls, init_params, hyperparameters, image, sagemaker_session):
```
owen-t: Add documentation here, it's important. You're introducing a protocol that subclasses need to follow, so this should be documented.

src/sagemaker/estimator.py
```diff
@@ -152,8 +152,46 @@ def fit(self, inputs, wait=True, logs=True, job_name=None):
...
+        """
+        sagemaker_session = sagemaker_session or Session()
+
+        if training_job_name is not None:
+            job_details = sagemaker_session.sagemaker_client.describe_training_job(TrainingJobName=training_job_name)
+            init_params, hp, image = cls._prepare_estimator_params_from_job_description(job_details)
+
         else:
-            raise NotImplemented('Asynchronous fit not available')
+            raise ValueError('must specify training_job name')
```
owen-t: IMO, this isn't necessary, just let the underlying call fail.

PR #65, opened 2018-01-31
Make sure that an arn including 'role' won't pass

matthieudelaro:
Currently, an ARN such as the following `'arn:aws:iam::424242424242:user/matthieu_role'` is accepted as a valid ARN by the get_execution_role() function, despite the fact that a test has been implemented:
```python
    if 'role' in arn:
        return arn
    message = 'The current AWS identity is not a role: {}, therefore it cannot be used as a SageMaker execution role'
    raise ValueError(message.format(arn))
```

Unfortunately, the test passes if ever the arn includes the term 'role'.

This commit improves this test by testing that 'role' is between a : and a /, in order to match arn like `arn:aws:iam::424242424242:role/sagemaker_role_matthieu`


mvsusp:
Thanks for contributing @matthieudelaro.
Can you add a unit test on https://github.com/aws/sagemaker-python-sdk/blob/157d8670977243f7f77327175d40364c885482b3/tests/unit/test_session.py#L31
with the missing use?

laurenyu:
continued this PR in #570

matthieudelaro:
@laurenyu, thanks a lot for all you work that made the merging possible!

PR #68, opened 2018-02-01
Fix Estimator role expansion

### LINKED ISSUE #43: creating estimator with role name fails if role is default sagemaker service role
if you pass a role name (instead of full ARN) to estimator constructor, and you are using default role created by SageMaker, training job will fail with "unable to assume the role" ClientError:

`sagemaker-mxnet-py2-cpu-2018-01-14-15-35-12-859: Failed Reason: ClientError: SageMaker was unable to assume the role 'arn:aws:iam::<ACCOUNT>:role/AmazonSageMaker-ExecutionRole-20180114T073202'`

error message shows the sdk converted role name to ARN without the 'service-role' prefix. running the same job with corrected full ARN works:

`arn:aws:iam::<ACCOUNT>:role/service-role/AmazonSageMaker-ExecutionRole-20180114T075260`

i encountered this error starting a training job, but I assume it would impact endpoint creation in same way.

iquintero:
Instead of manually constructing the role ARN, use the IAM boto client
to do it. This properly expands service-roles and regular roles.

This fixes #43

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -522,8 +522,8 @@ class Session(object):
     def expand_role(self, role):
         """Expand an IAM role name into an ARN.
 
-        If the role is already in the form of an ARN, then the role is simply returned. Otherwise, the role
-        is formatted as an ARN, using the current account as the IAM role's AWS account.
+        If the role is already in the form of an ARN, then the role is simply returned. Otherwise we retrieve the full
+        ARN and return it.
 
         Args:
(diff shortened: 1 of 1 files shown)


iquintero:
It doesn't really make sense to have a unit test for this, as I would have to mock the boto call to IAM and at that point there is nothing else left to test. Also the ARN is not being manually crafted anymore so we should be able to trust that boto is doing the right thing here.

mvsusp:
Agreed.

PR #69, opened 2018-02-02
Fix description of an argument of sagemaker.session.train

shotarok:
Fix description of an argument of `sagemaker.session.train`. `input_config` should be an array which has channel objects.   The description of `input_config` is taken from https://botocore.readthedocs.io/en/latest/reference/services/sagemaker.html#SageMaker.Client.create_training_job.

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -195,7 +195,9 @@ class Session(object):
                     a directory in the Docker container.
                 * 'Pipe' - Amazon SageMaker streams data directly from S3 to the container via a Unix-named pipe.
 
-            input_config (list): An array of Channel objects. Each channel is a named input source.
+            input_config (list): An array of Channel objects. Each channel is a named input source. Please refer to
+                 the format details described:
+                 https://botocore.readthedocs.io/en/latest/reference/services/sagemaker.html#SageMaker.Client.create_training_job
             role (str): An AWS IAM role (either name or full ARN). The Amazon SageMaker training jobs and APIs
                 that create Amazon SageMaker endpoints use this role to access training data and model artifacts.
                 You must grant sufficient permissions to this role.
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -195,7 +195,7 @@ class Session(object):
                     a directory in the Docker container.
                 * 'Pipe' - Amazon SageMaker streams data directly from S3 to the container via a Unix-named pipe.
 
-            input_config (list): An array of Channel objects. Each channel is a named input source. Please refer to
+            input_config (list): A list of Channel objects. Each channel is a named input source. Please refer to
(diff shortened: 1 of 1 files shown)


src/sagemaker/session.py
```diff
@@ -195,17 +195,7 @@ def train(self, image, input_mode, input_config, role, job_name, output_config,
...
-
-                * (str) - the S3 location where training data is saved.
-                * (dict[str, str] or dict[str, sagemaker.session.s3_input]) - If using multiple channels for
-                    training data, you can specify a dict mapping channel names
-                    to strings or :func:`~sagemaker.session.s3_input` objects.
-                * (sagemaker.session.s3_input) - channel configuration for S3 data sources that can provide
-                    additional information about the training dataset. See :func:`sagemaker.session.s3_input`
-                    for full details.
-
+            input_config (list): An array of Channel objects. Each channel is a named input source.
```
laurenyu: Could you add a link to the botocore docs?

shotarok: There are no links to `botocore` docs in this module. How about putting a link to `Channel` [API document](https://docs.aws.amazon.com/sagemaker/latest/dg/API_Channel.html) like [session.py#L744](https://github.com/aws/sagemaker-python-sdk/blob/b45d79c/src/sagemaker/session.py#L744)?

laurenyu: I meant the [link](https://botocore.readthedocs.io/en/latest/reference/services/sagemaker.html#SageMaker.Client.create_training_job) you referenced in your PR description - since there's more information there about `InputDataConfig`, I think it'd be nice to make it clear where to find more information. A link to the Channel API doc would be a nice addition as well

shotarok: I see. I added the link to the description in 3671efe.

src/sagemaker/session.py
```diff
@@ -195,17 +195,9 @@ def train(self, image, input_mode, input_config, role, job_name, output_config,
...
-
-                * (str) - the S3 location where training data is saved.
-                * (dict[str, str] or dict[str, sagemaker.session.s3_input]) - If using multiple channels for
-                    training data, you can specify a dict mapping channel names
-                    to strings or :func:`~sagemaker.session.s3_input` objects.
-                * (sagemaker.session.s3_input) - channel configuration for S3 data sources that can provide
-                    additional information about the training dataset. See :func:`sagemaker.session.s3_input`
-                    for full details.
-
+            input_config (list): An array of Channel objects. Each channel is a named input source. Please refer to
```
winstonaws: Should be "A list of Channel objects". "Array" in python actually refers to a different type: https://docs.python.org/3/library/array.html

shotarok: @winstonaws Thanks. I fixed it in a7178c3.

PR #73, opened 2018-02-07
Add ntm algorithm with doc, unit tests, integ tests

yangaws:
Since ntm does similar job to lda, the implementation basically follows LDA. All codes include 4 parts:

1, NTM, NTMModel, NTMPredictor implementation
2, Unit tests
3, Integ tests
4, Doc


src/sagemaker/amazon/ntm.py
```diff
@@ -0,0 +1,147 @@
...
+        self.weight_decay = weight_decay
+        self.learning_rate = learning_rate
+
+    def create_model(self):
+        """Return a :class:`~sagemaker.amazon.NTMModel` referencing the latest
+        s3 model data produced by this Estimator."""
+
+        return NTMModel(self.model_data, self.role, sagemaker_session=self.sagemaker_session)
+
+    def fit(self, records, mini_batch_size, **kwargs):
```
lukmis: According to the doc (https://docs.aws.amazon.com/sagemaker/latest/dg/ntm_hyperparameters.html) mini_batch_size is not required. This function should not be necessary.

lukmis: We could do validation for mini_batch_size if provided.

tests/integ/test_ntm.py
```diff
@@ -0,0 +1,77 @@
...
+
+        predict_input = np.random.rand(1, feature_num)
+        result = predictor.predict(predict_input)
+
+        assert len(result) == 1
+        for record in result:
+            assert record.label["topic_mixture"] is not None
+
+
+def _prepare_record_set_from_local_files(dir_path, destination, num_records, feature_dim, sagemaker_session):
```
lukmis: Could you move to a separate location as it is reused by both NTM and LDA?

tests/integ/test_ntm.py
```diff
@@ -0,0 +1,77 @@
...
+    with timeout_and_delete_endpoint_by_name(endpoint_name, sagemaker_session, minutes=20):
+        model = NTMModel(ntm.model_data, role='SageMakerRole', sagemaker_session=sagemaker_session)
+        predictor = model.deploy(1, 'ml.c4.xlarge', endpoint_name=endpoint_name)
+
+        predict_input = np.random.rand(1, feature_num)
+        result = predictor.predict(predict_input)
+
+        assert len(result) == 1
+        for record in result:
+            assert record.label["topic_mixture"] is not None
```
lukmis: According to: https://docs.aws.amazon.com/sagemaker/latest/dg/ntm-in-formats.html it is "topic_weights"

tests/integ/test_ntm.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # all records must be same
+        feature_num = int(all_records[0].features['values'].float32_tensor.shape[0])
+
+        ntm = NTM(role='SageMakerRole', train_instance_count=1, train_instance_type='ml.c4.xlarge', num_topics=10,
+                  sagemaker_session=sagemaker_session, base_job_name='test-ntm')
+
+        record_set = _prepare_record_set_from_local_files(data_path, ntm.data_location,
+                                                          len(all_records), feature_num, sagemaker_session)
+        ntm.fit(record_set, 100)
```
lukmis: Probably we can skip 2nd parameter here.

tests/unit/test_ntm.py
```diff
@@ -0,0 +1,288 @@
...
+    with pytest.raises(TypeError):
+        NTM(encoder_layers=0, sagemaker_session=sagemaker_session, **ALL_REQ_ARGS)
+
+
+def test_epochs_validation_fail_type(sagemaker_session):
+    with pytest.raises(ValueError):
+        NTM(epochs='other', sagemaker_session=sagemaker_session, **ALL_REQ_ARGS)
+
+
+def test_epochs_validation_fail_value(sagemaker_session):
```
lukmis: Since the validation checks both min and max it would be great if we had both conditions checked for these HPs.

tests/unit/test_ntm.py
```diff
@@ -0,0 +1,288 @@
...
+        'VolumeSizeInGB': 30
+    },
+    'stop_condition': {'MaxRuntimeInSeconds': 86400}
+}
+
+FEATURE_DIM = 10
+MINI_BATCH_SIZE = 200
+HYPERPARAMS = {'num_topics': NUM_TOPICS, 'feature_dim': FEATURE_DIM, 'mini_batch_size': MINI_BATCH_SIZE}
+STRINGIFIED_HYPERPARAMS = dict([(x, str(y)) for x, y in HYPERPARAMS.items()])
+HP_TRAIN_CALL = dict(BASE_TRAIN_CALL)
```
lukmis: If this is not being used anywhere please remove.

README.rst
```diff
@@ -39,7 +39,7 @@ You can install from source by cloning this repository and issuing a pip install
 
     git clone https://github.com/aws/sagemaker-python-sdk.git
     python setup.py sdist
-    pip install dist/sagemaker-1.0.3.tar.gz
```
lukmis: If you want to bump the version here, please update setup.py and CHANGELOG

src/sagemaker/amazon/ntm.py
yangaws: Thanks.

Now remove old validator and validate range of mini_batch_size instead.

tests/unit/test_ntm.py
yangaws: Removed unnecessary parameters.

README.rst
yangaws: Fixed! Thanks

tests/integ/test_ntm.py
yangaws: Moved the method to new file that imported by both lda and ntm.

tests/integ/test_ntm.py
yangaws: Changed it to None. I think we still need to pass a None there even if we don't want to pass any values.

tests/unit/test_ntm.py
yangaws: Updated.

Now all hyper-parameters with a range will be validated for both lower and upper limit.

tests/unit/test_ntm.py
```diff
@@ -286,11 +286,11 @@ def test_call_fit_wrong_type_mini_batch_size(sagemaker_session):
     data = RecordSet("s3://{}/{}".format(BUCKET_NAME, PREFIX), num_records=1, feature_dim=FEATURE_DIM,
                      channel='train')
 
-    with pytest.raises(ValueError):
+    with pytest.raises((TypeError, ValueError)):
```
yangaws: Some comments here: 

The reason I use tuple (TypeError, ValueError) is because, different error is raised in different python versions. Python 2 raises ValueError and Python 3 raises TypeError.

PR #76, opened 2018-02-14
Support tensorflow-1.5.0 and mxnet-1.0.0

nadiaya:


README.rst
```diff
@@ -682,19 +682,28 @@ When training and deploying training scripts, SageMaker runs your Python script
...
 
-Your MXNet training script will be run on version 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the train_instance_type, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
-
-Each Docker container has the following dependencies installed:
+Your MXNet training script will be run on version 1.0.0 (by default) or 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the train_instance_type, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
```
laurenyu: I know this isn't related to your change, but I'd make things like ``train_instance_type`` monospace (enclosed in backticks)

README.rst
```diff
@@ -682,19 +682,28 @@ When training and deploying training scripts, SageMaker runs your Python script
...
 -  CUDA 9.0
 -  numpy 1.12
 
+The MXNet Docker container version 1.1 has the following dependencies installed:
+-  Python 2.7 or Python 3.5, depending on the ``py_version`` argument on
+   the MXNet constructor.
+-  MXNet 1.0.0, built for either GPU or CPU, depending on the instance
+   type for training or deploying.
+-  CUDA 9.0
+-  numpy 1.13.3
```
laurenyu: this seems like it'll become unwieldy as new versions are added, especially since half the list is the same for each (1.0 vs 1.1). Is there a way to maybe put this into some kind of table-like display? (open to other suggestions, too, of course)

same thought below for TF

src/sagemaker/tensorflow/defaults.py
```diff
@@ -0,0 +1,13 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+DOCKER_TAG = "1.1"
```
laurenyu: use single quotes - that's been preferred throughout most of this repo

src/sagemaker/mxnet/defaults.py
```diff
@@ -0,0 +1,13 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+DOCKER_TAG = "1.1"
```
laurenyu: use single quotes - that's been preferred throughout most of this repo

src/sagemaker/tensorflow/estimator.py
```diff
@@ -107,7 +107,8 @@ class TensorFlow(Framework):
 
     __framework_name__ = 'tensorflow'
 
-    def __init__(self, training_steps=None, evaluation_steps=None, checkpoint_path=None, py_version="py2", **kwargs):
+    def __init__(self, training_steps=None, evaluation_steps=None, checkpoint_path=None, py_version="py2",
```
lukmis: Please also update function _prepare_init_params_from_job_description in both tf and mxnet - it is used during 'attach' and we may will need to pass tag there.

setup.py
```diff
@@ -11,7 +11,7 @@ def read(fname):
 
 
 setup(name="sagemaker",
-      version="1.0.4",
+      version="1.1.0",
```
lukmis: Please update CHANGELOG when bumping the version.

src/sagemaker/tensorflow/estimator.py
mvsusp: I was thinking the same @lukmis ! I am afraid that docker tag support will not work with attachment.

mvsusp:
We need to tests both versions in the integration tests. 
We need to test attachment as well.

tests/unit/test_estimator.py
```diff
@@ -325,15 +325,27 @@ def test_init_with_source_dir_s3(strftime, sagemaker_session):
...
 def test_framework_name_from_other():
-    framework, py_ver = framework_name_from_image('123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-myown-py2-gpu:1')
+    framework, py_ver, image_ver = \
+        framework_name_from_image('123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-myown-py2-gpu:1.0')
     assert framework is None
     assert py_ver is None
+    assert image_ver is None
+
+
+def test_framework_name_from_image_wrong_version():
```
lukmis: Are these tests suppose to test parsing of the image name? If so, what is wrong with 1.5 version? I think 'framework_name_from_image' shall just return 1.5.

src/sagemaker/fw_utils.py
```diff
@@ -114,19 +114,19 @@ def framework_name_from_image(image_name):
...
+    sagemaker_pattern = re.compile('^(\d+)(\.)dkr(\.)ecr(\.)(.+)(\.)amazonaws.com(/)(.*:.*)$')
     sagemaker_match = sagemaker_pattern.match(image_name)
     if sagemaker_match is None:
-        return None, None
+        return None, None, None
     else:
-        # extract framework and python version
-        name_pattern = re.compile('^sagemaker-(tensorflow|mxnet)-(py2|py3)-(cpu|gpu)$')
+        # extract framework, python version and image version
+        name_pattern = re.compile('^sagemaker-(tensorflow|mxnet)-(py2|py3)-(cpu|gpu):(1.[0..1])$')
```
lukmis: This utility function shall just return the tag and not make assumptions what is allowed. This decision can be made by a caller.

iquintero: I agree, maybe change the last regex group to be: ([0-9.]+) or something along those lines.

Issue #52, opened 2018-01-21
Installing Dependencies


cgarciae:
Hi! I am using Deep Mind's Sonnet library to make my architecture code reusable. Sonnet is not included in the `py2` container and I have not found a way to pass additional requirements to SageMaker. Could you implement a way to pass a `requirements.txt` file to the `Tensorflow` constructor so SageMaker installs the requirements in the Docker container before training? Floydhub lets you define the `floyd_requirements.txt` file that does exactly this and works very well.

cgarciae:
Excellent! Looking forward to it.

For anyone interested, I've resorted to installing dependencies from python, just add these lines before all other imports:
```python
from subprocess import call
call("pip install dm-sonnet".split(" "))
```

nadiaya:
This feature has been released.

YashalShakti:
Hi,

Will this feature be added to the MxNet container as well?

Thanks!

Jacob-Stevens-Haas:
Hey, not sure if I'm doing it wrong, or if "feature released" is only python2 (which was the OP's version).

I open terminal `pip install geopy`.  Asked some AWS folks at an event and they said to install from within notebook, because the instance of Python is specific to the notebook.  That doesn't work in normal notebooks, but, hey look, it still doesn't work.   @cgarciae's trick does.  Is that supposed to be how we install them?

dshahrokhian:
Is this feature only working with tensorflow?

PR #80, opened 2018-02-16
Support user-supplied requirements.txt for TensorFlow

laurenyu:
This change allows the path to a 'requirements.txt' file to be specified for training and hosting. The file has to be in the same directory as the other files being bundled to S3. This should address #52.


src/sagemaker/tensorflow/estimator.py
```diff
@@ -116,13 +117,16 @@ def __init__(self, training_steps=None, evaluation_steps=None, checkpoint_path=N
             checkpoint_path (str): Identifies S3 location where checkpoint data during model training can be
                 saved (default: None). For distributed model training, this parameter is required.
             py_version (str): Python version you want to use for executing your model training code (default: 'py2').
+            requirements_file (str): Path to a ``requirements.txt`` file (default: ''). The path should be relative to
```
iquintero: I believe the path should be within the source_dir. By saying relative a user could do:

requirements_file="../requirements_file.txt"

but it wouldn't have been uploaded. Then it will fail.

laurenyu: good point - clarified the wording

tests/integ/test_tf.py
```diff
@@ -87,6 +87,33 @@ def test_tf_async(sagemaker_session):
         print('predict result: {}'.format(result))
 
 
+def test_tf_with_requirements(sagemaker_session):
```
mvsusp: This test should assert that the library BeautifulSoup4 is installed in the container.
You can do that by pip freezing and saving it as a file in opt/ml/model.
You can check later that the model has the file.

That being said, consider if this test is really necessary.

tests/unit/test_tf_estimator.py
```diff
@@ -174,6 +178,7 @@ def test_tf(time, strftime, sagemaker_session):
     assert {'Environment':
             {'SAGEMAKER_SUBMIT_DIRECTORY': 's3://{}/{}/sourcedir.tar.gz'.format(BUCKET_NAME, JOB_NAME),
              'SAGEMAKER_PROGRAM': 'dummy_script.py',
+             'SAGEMAKER_REQUIREMENTS': 'requirements.txt',
```
mvsusp: What happens with the requirements file is in a different folder than the user_script?
What happens if the user passes relative or absolute paths?
What happens if the user passes an invalid path?

We need unit test cases for that.

tests/integ/test_tf.py
laurenyu: what are your reasons for this test being unnecessary?

mvsusp: We are testing that the user supplied requeriments are working in the container side.
Not all the containers features are tested in the client, only the main ones, to avoid long tests covering edge cases here. 
One possible option, is to change our current integration tests to cover your scenario as well, but I would suggest you to import a library that is required for training.
Another option is to add an additional integration test that adds value besides just the requeriment itself. A Keras model that loads a hdf5 file would be a great example.

mvsusp: We are testing that the user supplied requeriments are working in the container side.
Not all the containers features are tested in the client, only the main ones, to avoid long tests covering edge cases here. 
One possible option, is to change our current integration tests to cover your scenario as well, but I would suggest you to import a library that is required for training.
Another option is to add an additional integration test that adds value besides just the requeriment itself. A Keras model that loads a hdf5 file would be a great example.

tests/unit/test_tf_estimator.py
laurenyu: we can add validation (with corresponding unit tests) for that. I think it'd be sufficient to check that the path is not relative and that the file exists using `os.path.exists()`

mvsusp: Yeah, I think that will be sufficient. Thanks.

laurenyu:
I've now removed the integ test I had written, added validation for the requirements file argument with corresponding unit tests, and documented the additional argument for the TF estimator in the README.

chengshunxia:
can a similar implementation added to other framework (pytorch/mxnet)? would it better to add this in   EstimatorBase ??

laurenyu:
@chengshunxia this has already been implemented for other frameworks:
* https://sagemaker.readthedocs.io/en/stable/frameworks/mxnet/using_mxnet.html#use-third-party-libraries
* https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#using-third-party-libraries
* https://sagemaker.readthedocs.io/en/stable/frameworks/sklearn/using_sklearn.html#using-third-party-libraries

chengshunxia:
so we don't need add a parameter in the python code, just add a requirements.txt under source directory , is that right ?

PR #82, opened 2018-02-23
Support TensorFlow-1.5.0 and MXNet-1.0.0 

nadiaya:
Add support for TensorFlow-1.5.0 and MXNet-1.0.0

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,12 +2,18 @@
 CHANGELOG
 =========
 
+1.1.0
+=====
+
+* feature: Estimators: add support for TensorFlow-1.5.0
+* feature: Estimators: add support for MXNet-1.0.0
+
 1.0.4
 =====
 
 * feature: Estimators: add support for Amazon Neural Topic Model(NTM) algorithm
-* feature: Documentation: Fix description of an argument of sagemaker.session.train
-* feature: Documentation: Add FM and LDA to the documentation
+* feature: Documentation: fix description of an argument of sagemaker.session.train
+* feature: Documentation: add FM and LDA to the documentation
 * feature: Estimators: add support for async fit
 * bug-fix: Estimators: fix estimator role expansion
 
@@ -15,16 +21,16 @@ CHANGELOG
 =====
 
 * feature: Estimators: add support for Amazon LDA algorithm
-* feature: Hyperparameters: Add data_type to hyperparameters
-* feature: Documentation: Update TensorFlow examples following API change
-* feature: Session: Support multi-part uploads
+* feature: Hyperparameters: add data_type to hyperparameters
+* feature: Documentation: update TensorFlow examples following API change
+* feature: Session: support multi-part uploads
 
 
 1.0.2
 =====
 
 * feature: Estimators: add support for Amazon FactorizationMachines algorithm
-* feature: Session: Correctly handle TooManyBuckets error_code in default_bucket method
+* feature: Session: correctly handle TooManyBuckets error_code in default_bucket method
 * feature: Tests: add training failure tests for TF and MXNet
 * feature: Documentation: show how to make predictions against existing endpoint
 * feature: Estimators: implement write_spmatrix_to_sparse_tensor to support any scipy.sparse matrix
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ You can install from source by cloning this repository and issuing a pip install
 
     git clone https://github.com/aws/sagemaker-python-sdk.git
     python setup.py sdist
-    pip install dist/sagemaker-1.0.4.tar.gz
+    pip install dist/sagemaker-1.1.0.tar.gz
 
 Supported Python versions
 ~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -114,6 +114,8 @@ MXNet SageMaker Estimators
 
 With MXNet Estimators, you can train and host MXNet models on Amazon SageMaker.
 
+Supported versions of MXNet: ``1.0.0``, ``0.12.1``.
+
 Training with MXNet
 ~~~~~~~~~~~~~~~~~~~
 
@@ -185,7 +187,7 @@ If you want to run your training script locally via the Python interpreter, look
 Using MXNet and numpy
 ^^^^^^^^^^^^^^^^^^^^^
 
-You can import both ``mxnet`` and ``numpy`` in your training script. When your script runs in SageMaker, it will run with access to MXNet version 0.12 and numpy version 1.12.0. For more information on the environment your script runs in, please see `SageMaker MXNet Containers <#sagemaker-mxnet-containers>`__.
+You can import both ``mxnet`` and ``numpy`` in your training script. When your script runs in SageMaker, it will run with access to MXNet version 1.0.0 and numpy version 1.13.3 by default. For more information on the environment your script runs in, please see `SageMaker MXNet Containers <#sagemaker-mxnet-containers>`__.
 
 Running an MXNet training script in SageMaker
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -682,18 +684,23 @@ When training and deploying training scripts, SageMaker runs your Python script 
 
 SageMaker runs MXNet Estimator scripts in either Python 2.7 or Python 3.5. You can select the Python version by passing a ``py_version`` keyword arg to the MXNet Estimator constructor. Setting this to ``py2`` (the default) will cause your training script to be run on Python 2.7. Setting this to ``py3`` will cause your training script to be run on Python 3.5. This Python version applies to both the Training Job, created by fit, and the Endpoint, created by deploy.
 
-Your MXNet training script will be run on version 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the train_instance_type, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
+Your MXNet training script will be run on version 1.0.0 (by default) or 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the ``train_instance_type``, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
 
-Each Docker container has the following dependencies installed:
+The Docker images have the following dependencies installed:
 
--  Python 2.7 or Python 3.5, depending on the ``py_version`` argument on
-   the MXNet constructor.
--  MXNet 0.12, built for either GPU or CPU, depending on the instance
-   type for training or deploying.
--  CUDA 9.0
--  numpy 1.12
++-------------------------+--------------+-------------+
+| Dependencies            | MXNet 0.12.1 | MXNet 1.0.0 |
++-------------------------+--------------+-------------+
+| Python                  |   2.7 or 3.5 |   2.7 or 3.5|
++-------------------------+--------------+-------------+
+| CUDA                    |          9.0 |         9.0 |
++-------------------------+--------------+-------------+
+| numpy                   |       1.13.3 |      1.13.3 |
++-------------------------+--------------+-------------+
 
 The Docker images extend Ubuntu 16.04.
+
+You can select version of MXNet by passing a ``framework_version`` keyword arg to the MXNet Estimator constructor. Currently supported versions are ``1.0.0`` and ``0.12.1``. You can also set ``framework_version`` to ``1.0 (default)`` or ``0.12`` which will cause your training script to be run on the latest supported MXNet 1.0 or 0.12 versions respectively.
 
 TensorFlow SageMaker Estimators
 -------------------------------
@@ -701,6 +708,8 @@ TensorFlow SageMaker Estimators
 TensorFlow SageMaker Estimators allow you to run your own TensorFlow
 training algorithms on SageMaker Learner, and to host your own TensorFlow
 models on SageMaker Hosting.
+
+Supported versions of TensorFlow: ``1.4.1``, ``1.5.0``.
 
 Training with TensorFlow
 ~~~~~~~~~~~~~~~~~~~~~~~~
@@ -735,7 +744,7 @@ Preparing the TensorFlow training script
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Your TensorFlow training script must be a **Python 2.7** source file. The current supported TensorFlow
-version is **1.4.0**. This training script **must contain** the following functions:
+versions are **1.5.0 (default)** and **1.4.1**. This training script **must contain** the following functions:
 
 - ``model_fn``: defines the model that will be trained.
 - ``train_input_fn``: preprocess and load training data.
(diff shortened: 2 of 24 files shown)


tests/unit/test_estimator.py
```diff
@@ -325,15 +325,18 @@ def test_init_with_source_dir_s3(strftime, sagemaker_session):
 
 
 def test_framework_name_from_framework_image():
-    framework, py_ver = framework_name_from_image('123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-mxnet-py2-gpu:1')
+    image_name = '123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-mxnet-py2-gpu:2.5.6-gpu-py2'
+    framework, py_ver, tag = framework_name_from_image(image_name)
     assert framework == 'mxnet'
     assert py_ver == 'py2'
+    assert tag == '2.5.6-gpu-py2'
```
mvsusp: 👍 🥇 I love this tags.

src/sagemaker/fw_utils.py
```diff
@@ -28,26 +28,28 @@
 """
 
 
-def create_image_uri(region, framework, instance_type, py_version='py2', tag='1.0', account='520713654638'):
+def create_image_uri(region, framework, instance_type, framework_version, py_version='py2', account='520713654638'):
```
mvsusp: I've notice that `create_image_uri` does not have unit tests. Can you write unit tests for this function?

tests/unit/test_mxnet.py
```diff
@@ -95,31 +105,33 @@ def test_mxnet(strftime, sagemaker_session):
...
             'ModelDataUrl': 's3://m/m.tar.gz'} == model.prepare_container_def(GPU)
 
     assert 'cpu' in model.prepare_container_def(CPU)['Image']
     predictor = mx.deploy(1, GPU)
     assert isinstance(predictor, MXNetPredictor)
 
 
-def test_model(sagemaker_session):
-    model = MXNetModel("s3://some/data.tar.gz", role=ROLE, entry_point=SCRIPT_PATH, sagemaker_session=sagemaker_session)
+def test_model(sagemaker_session, mxnet_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -105,32 +115,32 @@ def _build_tf(sagemaker_session, train_instance_type=None, checkpoint_path=None,
...
-    tf = _build_tf(sagemaker_session, train_instance_type='ml.p2.2xlarge')
+    tf = _build_tf(sagemaker_session, tf_version, train_instance_type='ml.p2.2xlarge')
 
-    assert tf.train_image() == FULL_GPU_IMAGE_URI
+    assert tf.train_image() == _get_full_gpu_image_uri(tf_version)
 
 
-def test_tf_deploy_model_server_workers(sagemaker_session):
-    tf = _build_tf(sagemaker_session)
+def test_tf_deploy_model_server_workers(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -139,8 +149,8 @@ def test_tf_deploy_model_server_workers(sagemaker_session):
         MODEL_SERVER_WORKERS_PARAM_NAME.upper()]
 
 
-def test_tf_deploy_model_server_workers_unset(sagemaker_session):
-    tf = _build_tf(sagemaker_session)
+def test_tf_deploy_model_server_workers_unset(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -192,19 +202,21 @@ def test_tf(time, strftime, sagemaker_session):
 @patch('subprocess.Popen')
 @patch('subprocess.call')
 @patch('os.access', return_value=False)
-def test_run_tensorboard_locally_without_tensorboard_binary(time, strftime, popen, call, access, sagemaker_session):
+def test_run_tensorboard_locally_without_tensorboard_binary(time, strftime, popen, call, access,
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -192,19 +202,21 @@ def test_tf(time, strftime, sagemaker_session):
...
+                    framework_version=tf_version)
 
     with pytest.raises(EnvironmentError) as error:
         tf.fit(inputs='s3://mybucket/train', run_tensorboard_locally=True)
     assert str(error.value) == 'TensorBoard is not installed in the system. Please install TensorBoard using the ' \
                                'following command: \n pip install tensorboard'
 
 
-def test_model(sagemaker_session):
+def test_model(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -214,9 +226,11 @@ def test_model(sagemaker_session):
 @patch('subprocess.Popen')
 @patch('subprocess.call')
 @patch('os.access', side_effect=[False, True])
-def test_run_tensorboard_locally_without_awscli_binary(time, strftime, popen, call, access, sagemaker_session):
+def test_run_tensorboard_locally_without_awscli_binary(time, strftime, popen, call, access,
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -231,9 +245,10 @@ def test_run_tensorboard_locally_without_awscli_binary(time, strftime, popen, ca
 @patch('time.strftime', return_value=TIMESTAMP)
 @patch('time.time', return_value=TIME)
 @pytest.mark.skip(reason="this test fails sometimes and it needs further investigation")
-def test_run_tensorboard_locally(time, strftime, popen, call, access, sagemaker_session):
+def test_run_tensorboard_locally(time, strftime, popen, call, access, sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -251,9 +266,11 @@ def test_run_tensorboard_locally(time, strftime, popen, call, access, sagemaker_
 @patch('time.strftime', return_value=TIMESTAMP)
 @patch('time.time', return_value=TIME)
 @pytest.mark.skip(reason="this test fails sometimes and it needs further investigation")
-def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access, socket, sagemaker_session):
+def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access, socket,
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -266,51 +283,103 @@ def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access
                           stderr=-1, stdout=-1)
 
 
-def test_tf_checkpoint_not_set(sagemaker_session):
+def test_tf_checkpoint_not_set(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -266,51 +283,103 @@ def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access
...
+    tf = _build_tf(sagemaker_session, tf_version, checkpoint_path=None, base_job_name=job_name,
                    output_path="s3://{}/".format(sagemaker_session.default_bucket()))
     tf.fit(inputs=s3_input('s3://mybucket/train'), job_name=job_name)
 
     expected_result = '"s3://{}/{}/checkpoints"'.format(sagemaker_session.default_bucket(), job_name)
     assert tf.hyperparameters()['checkpoint_path'] == expected_result
 
 
-def test_tf_training_and_evaluation_steps_not_set(sagemaker_session):
+def test_tf_training_and_evaluation_steps_not_set(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -266,51 +283,103 @@ def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access
...
 
-    tf = _build_tf(sagemaker_session, training_steps=None, evalutation_steps=None, output_path=output_path)
+    tf = _build_tf(sagemaker_session, tf_version, training_steps=None, evalutation_steps=None, output_path=output_path)
     tf.fit(inputs=s3_input('s3://mybucket/train'))
     assert tf.hyperparameters()['training_steps'] == 'null'
     assert tf.hyperparameters()['evaluation_steps'] == 'null'
 
 
-def test_tf_training_and_evaluation_steps(sagemaker_session):
+def test_tf_training_and_evaluation_steps(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

tests/unit/test_tf_estimator.py
```diff
@@ -266,51 +283,103 @@ def test_run_tensorboard_locally_port_in_use(time, strftime, popen, call, access
...
-    tf = _build_tf(sagemaker_session, training_steps=123, evalutation_steps=456, output_path=output_path)
+    tf = _build_tf(sagemaker_session, tf_version, training_steps=123, evalutation_steps=456, output_path=output_path)
     tf.fit(inputs=s3_input('s3://mybucket/train'))
     assert tf.hyperparameters()['training_steps'] == '123'
     assert tf.hyperparameters()['evaluation_steps'] == '456'
 
 
-def test_tf_checkpoint_set(sagemaker_session):
-    tf = _build_tf(sagemaker_session, checkpoint_path='s3://my_checkpoint_bucket')
+def test_tf_checkpoint_set(sagemaker_session, tf_version):
```
mvsusp: This test does not need to run for both versions.

src/sagemaker/mxnet/defaults.py
```diff
@@ -0,0 +1,13 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+MXNET_VERSION = '1.0'
```
winstonaws: Maybe "DEFAULT_MXNET_VERSION"?

src/sagemaker/tensorflow/defaults.py
```diff
@@ -0,0 +1,13 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+TF_VERSION = '1.5'
```
winstonaws: "DEFAULT_TF_VERSION"?

src/sagemaker/tensorflow/estimator.py
```diff
@@ -185,9 +189,13 @@ def _prepare_init_params_from_job_description(cls, job_details):
             if value is not None:
                 init_params[argument] = value
 
-        framework, py_version = framework_name_from_image(init_params.pop('image'))
+        framework, py_version, tag = framework_name_from_image(init_params.pop('image'))
         init_params['py_version'] = py_version
 
+        # For backward compatibility map deprecated container tag to a framework version
```
winstonaws: container tag -> image tag

src/sagemaker/mxnet/estimator.py
```diff
@@ -78,8 +82,8 @@ def create_model(self, model_server_workers=None):
         return MXNetModel(self.model_data, self.role, self.entry_point, source_dir=self.source_dir,
                           enable_cloudwatch_metrics=self.enable_cloudwatch_metrics, name=self._current_job_name,
                           container_log_level=self.container_log_level, code_location=self.code_location,
-                          py_version=self.py_version, model_server_workers=model_server_workers,
-                          sagemaker_session=self.sagemaker_session)
+                          py_version=self.py_version, framework_version=self.framework_version,
```
winstonaws: We probably want an extra assertion in the appropriate unit test that we pass along the framework version correctly to the created model (for both TF and MXNet).

src/sagemaker/mxnet/estimator.py
```diff
@@ -47,10 +48,12 @@ def __init__(self, entry_point, source_dir=None, hyperparameters=None, py_versio
                 to convert them before training.
             py_version (str): Python version you want to use for executing your model training code (default: 'py2').
                               One of 'py2' or 'py3'.
+            framework_version (str): MXNet version you want to use for executing your model training code.
```
winstonaws: Can you add info to this docstring to tell users how they can figure out what valid values are? (Same for TF as well.)

CHANGELOG.rst
```diff
@@ -2,29 +2,35 @@
 CHANGELOG
 =========
 
+1.1.0
+=====
+
+* feature: Estimators: add support for TensorFlow-1.5.0
```
lukmis: Please update after master branch merge.

src/sagemaker/tensorflow/estimator.py
```diff
@@ -185,9 +189,13 @@ def _prepare_init_params_from_job_description(cls, job_details):
...
+        # otherwise extract framework version from the tag itself.
```
lukmis: Could we have an utility function that would return version from the tag? Otherwise we hardcode this logic (element [0] of the tag)  in 2 different places (for both frameworks).

tests/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import pytest
+
+
+@pytest.fixture(scope='module', params=["1.4", "1.4.1", "1.5", "1.5.0"])
```
lukmis: It would be great to cover the default case, e.g. pass special word like 'default' and in the test do not pass version at all.

tests/unit/test_estimator.py
```diff
@@ -325,15 +325,18 @@ def test_init_with_source_dir_s3(strftime, sagemaker_session):
...
 
 
 def test_framework_name_from_other():
-    framework, py_ver = framework_name_from_image('123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-myown-py2-gpu:1')
+    framework, py_ver, tag = framework_name_from_image('123.dkr.ecr.us-west-2.amazonaws.com/sagemaker-myown-py2-gpu:1')
     assert framework is None
     assert py_ver is None
+    assert tag is None
```
lukmis: Why would we get 'None' here? The tag will be '1', right?

src/sagemaker/mxnet/estimator.py
```diff
@@ -21,7 +21,8 @@ class MXNet(Framework):
 
     __framework_name__ = "mxnet"
 
-    def __init__(self, entry_point, source_dir=None, hyperparameters=None, py_version='py2', **kwargs):
+    def __init__(self, entry_point, source_dir=None, hyperparameters=None, py_version='py2',
+                 framework_version=MXNET_VERSION, **kwargs):
```
lukmis: Something to consider: shall the framework_version be part of the Framework class and not be repeated in each of the child classes?

README.rst
```diff
@@ -682,19 +682,24 @@ When training and deploying training scripts, SageMaker runs your Python script
...
 
-Your MXNet training script will be run on version 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the train_instance_type, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
+Your MXNet training script will be run on version 1.0.0 (by default) or 0.12 of MXNet, built for either GPU or CPU use. The decision to use the GPU or CPU version of MXNet is made by the ``train_instance_type``, set on the MXNet constructor. If you choose a GPU instance type, your training job will be run on a GPU version of MXNet. If you choose a CPU instance type, your training job will be run on a CPU version of MXNet. Similarly, when you call deploy, specifying a GPU or CPU deploy_instance_type, will control which MXNet build your Endpoint runs.
```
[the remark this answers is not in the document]
nadiaya: But we still would need to make an sdk change to change the default.
I believe it adds value to say what version we support by default.

src/sagemaker/mxnet/defaults.py
nadiaya: defaults.DEFAULT_MXNET_VERSION ?

src/sagemaker/tensorflow/defaults.py
nadiaya: defaults.DEFAULT_TF_VERSION ?

src/sagemaker/mxnet/estimator.py
nadiaya: Maybe in the future, we decided not to do it as part of this change.

tests/unit/test_estimator.py
nadiaya: Not the way framework_name_from_image works. In this case it:
```
        name_pattern = re.compile('^sagemaker-(tensorflow|mxnet)-(py2|py3)-(cpu|gpu):(.*)$')
        name_match = name_pattern.match(sagemaker_match.group(8))
```
name_match is None and everything is None as a result.
I don't see the value in changing this behavior since we can't really use the tag if we don't have the framework, similar to python version.

src/sagemaker/tensorflow/estimator.py
nadiaya: i really thought about it, but the version is specific to mxnet and tensorflow and it's too little to generalize.
Also we should be able to remove it at some point in the future.

CHANGELOG.rst
nadiaya: Will do.

src/sagemaker/fw_utils.py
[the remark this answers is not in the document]
nadiaya: I don't have an explanation on why it was defaulted at all.

src/sagemaker/fw_utils.py
```diff
@@ -28,26 +28,28 @@
...
         py_version (str): Python version. (default: 'py2')
-        tag (str): ECR image tag, which denotes the image version. (default: '1.0')
         account (str): AWS account that contains the image. (default: '520713654638')
 
     Returns:
         str: The appropriate image URI based on the given parameters.
     """
-    device_version = 'cpu'
+    device_type = 'cpu'
     # Instance types that start with G, P are GPU powered: https://aws.amazon.com/ec2/instance-types/
```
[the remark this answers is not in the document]
nadiaya: Good point :)

tests/conftest.py
nadiaya: But it would run one of the versions again for every test.
We have a specific test for testing default value.

src/sagemaker/mxnet/estimator.py
nadiaya: https://github.com/aws/sagemaker-python-sdk/pull/82/files#diff-8e5bb247678c2ede89b070ed8c1926f2R192

We already do this indirectly.

I guess it would be good to have a unit test that tests just model creation.

src/sagemaker/fw_utils.py
nadiaya: it's doesn't, it should 👍

tests/unit/test_estimator.py
lukmis: Ah yes, sorry - I missed the 'myown' framework name.

tests/conftest.py
lukmis: That's true about the double run but I didn't see the special (integ) test for default value.

src/sagemaker/tensorflow/estimator.py
lukmis: What I had in mind is that the logic to deal with tag parsing should be externalized because we have same pattern for all the frameworks. It' same way as we do with 'framework_name_from_image'.
Version is specific to each but getting that version is common. If the pattern chanes in the future we will only update it in 1 place.

tests/conftest.py
nadiaya: https://github.com/aws/sagemaker-python-sdk/pull/82/files/bc748929fc3088d99d1e5c00787e3ddad1a3a51b#diff-8e5bb247678c2ede89b070ed8c1926f2R321 
and we have another similar one for mxnet

src/sagemaker/tensorflow/estimator.py
nadiaya: Got it, yeah probably makes sense even for 1 line function.

src/sagemaker/mxnet/estimator.py
nadiaya: Yep! 👍

src/sagemaker/mxnet/defaults.py
winstonaws: Good point, whoops.

PR #114, opened 2018-03-30
Use new repository names for Framework images

winstonaws:
We are switching to an ECR repo naming convention which more closely follows typical docker repo conventions. Instead of including device and python version info in the repo name, e.g. 'sagemaker-tensorflow-py2-cpu:1.5', we have one repo for all images of a framework, and put the rest of the info in the tag, e.g. 'sagemaker-tensorflow:1.5-cpu-py2'.

### DIFF
--- a/src/sagemaker/fw_utils.py
+++ b/src/sagemaker/fw_utils.py
@@ -28,28 +28,40 @@ instantiated with positional or keyword arguments.
 """
 
 
-def create_image_uri(region, framework, instance_type, framework_version, py_version, account='520713654638'):
+def create_image_uri(region, framework, instance_type, framework_version, py_version, account='520713654638',
+                     optimized_families=[]):
     """Return the ECR URI of an image.
 
     Args:
         region (str): AWS region where the image is uploaded.
         framework (str): framework used by the image.
-        instance_type (str): EC2 instance type. Used to determine whether to use the CPU image or GPU image.
+        instance_type (str): SageMaker instance type. Used to determine device type (cpu/gpu/family-specific optimized).
         framework_version (str): The version of the framework.
         py_version (str): Python version. One of 'py2' or 'py3'.
         account (str): AWS account that contains the image. (default: '520713654638')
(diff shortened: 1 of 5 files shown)


src/sagemaker/fw_utils.py
```diff
@@ -28,28 +28,40 @@
...
 
     Args:
         region (str): AWS region where the image is uploaded.
         framework (str): framework used by the image.
-        instance_type (str): EC2 instance type. Used to determine whether to use the CPU image or GPU image.
+        instance_type (str): SageMaker instance type. Used to determine device type (cpu/gpu/family-specific optimized).
         framework_version (str): The version of the framework.
         py_version (str): Python version. One of 'py2' or 'py3'.
         account (str): AWS account that contains the image. (default: '520713654638')
+        optimized_families (str): Instance families for which there exist specific optimized images.
```
nadiaya: do we need to pass it as an argument? is it going to be different per framework?

src/sagemaker/fw_utils.py
```diff
@@ -107,8 +119,11 @@ def framework_name_from_image(image_name):
     """Extract the framework and Python version from the image name.
 
     Args:
-        image_name (str): Image URI, which should take the form
-            '<account>.dkr.ecr.<region>.amazonaws.com/sagemaker-<framework>-<py_ver>-<device>:<tag>'
+        image_name (str): Image URI, which should either of the forms:
+            legacy:
+            '<account>.dkr.ecr.<region>.amazonaws.com/sagemaker-<framework>-<py_ver>-<device>:<fw_version>'
```
nadiaya: there is two types of legacy:
```
<account>.dkr.ecr.<region>.amazonaws.com/sagemaker-<framework>-<py_ver>-<device>:<container_version>
```
and
```
<account>.dkr.ecr.<region>.amazonaws.com/sagemaker-<framework>-<py_ver>-<device>:<fw_version>-<device>-<py_ver>
```

PR #115, opened 2018-03-30
Add Local Mode support

iquintero:
When passing "local" as the instance type for any estimator,
training and deployment happens locally.

Similarly, using "local_gpu" will use nvidia-docker-compose and
work for GPU training.

Local Mode also works for Hosting and inference calls can be done locally.

### DIFF
--- a/src/sagemaker/estimator.py
+++ b/src/sagemaker/estimator.py
@@ -21,7 +21,7 @@ from six import with_metaclass, string_types
 from sagemaker.fw_utils import tar_and_upload_dir
 from sagemaker.fw_utils import parse_s3_url
 from sagemaker.fw_utils import UploadedCode
-from sagemaker.local_session import LocalSession
+from sagemaker.local.local_session import LocalSession
 from sagemaker.model import Model
 from sagemaker.model import (SCRIPT_PARAM_NAME, DIR_PARAM_NAME, CLOUDWATCH_METRICS_PARAM_NAME,
                              CONTAINER_LOG_LEVEL_PARAM_NAME, JOB_NAME_PARAM_NAME, SAGEMAKER_REGION_PARAM_NAME)
--- a/src/sagemaker/estimator.py
+++ b/src/sagemaker/estimator.py
@@ -21,6 +21,7 @@ from six import with_metaclass, string_types
 from sagemaker.fw_utils import tar_and_upload_dir
 from sagemaker.fw_utils import parse_s3_url
 from sagemaker.fw_utils import UploadedCode
+from sagemaker.local_session import LocalSession
 from sagemaker.model import Model
 from sagemaker.model import (SCRIPT_PARAM_NAME, DIR_PARAM_NAME, CLOUDWATCH_METRICS_PARAM_NAME,
                              CONTAINER_LOG_LEVEL_PARAM_NAME, JOB_NAME_PARAM_NAME, SAGEMAKER_REGION_PARAM_NAME)
@@ -78,7 +79,17 @@ class EstimatorBase(with_metaclass(ABCMeta, object)):
         self.train_volume_size = train_volume_size
         self.train_max_run = train_max_run
         self.input_mode = input_mode
-        self.sagemaker_session = sagemaker_session or Session()
+
+        if self.train_instance_type in ('local', 'local_gpu'):
+            self.local_mode = True
+            if self.train_instance_type == 'local_gpu' and self.train_instance_count > 1:
+                raise RuntimeError("Distributed Training in Local GPU is not supported")
+
+            self.sagemaker_session = LocalSession()
+        else:
+            self.local_mode = False
+            self.sagemaker_session = sagemaker_session or Session()
+
         self.base_job_name = base_job_name
         self._current_job_name = None
         self.output_path = output_path
@@ -303,7 +314,7 @@ class _TrainingJob(object):
         """Create a new Amazon SageMaker training job from the estimator.
 
         Args:
-            estimator (sagemaker.estimator.Framework): Estimator object created by the user.
+            estimator (sagemaker.estimator.EstimatorBase): Estimator object created by the user.
             inputs (str): Parameters used when called  :meth:`~sagemaker.estimator.EstimatorBase.fit`.
 
         Returns:
--- a/src/sagemaker/image.py
+++ b/src/sagemaker/image.py
@@ -1,0 +1,446 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+import json
+import logging
+import os
+import platform
+import shutil
+import subprocess
+import sys
+import tempfile
+from subprocess import Popen, STDOUT, CalledProcessError
+from time import sleep
+
+import yaml
+
+CONTAINER_PREFIX = "algo"
+DOCKER_COMPOSE_FILENAME = 'docker-compose.yaml'
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+
+class SageMakerContainer(object):
+
+    def __init__(self, instance_type, instance_count, image, sagemaker_session=None):
+        from sagemaker.local_session import LocalSession
+        self.sagemaker_session = sagemaker_session or LocalSession()
+        self.instance_type = instance_type
+        self.instance_count = instance_count
+        self.image = image
+        self.hosts = ['{}-{}'.format(CONTAINER_PREFIX, i) for i in range(1, self.instance_count + 1)]
+        self.container_root = None
+        self.container = None
+
+    def train(self, input_data_config, hyperparameters):
+        """Run a training job locally using docker-compose.
+        Args:
+            input_data_config (dict): The Input Data Configuration, this contains data such as the
+                channels to be used for training.
+            hyperparameters (dict): The HyperParameters for the training job.
+
+        Returns (str): Location of the trained model.
+
+        """
+        self.container_root = _create_tmp_folder()
+
+        data_dir = tempfile.mkdtemp()
+        bucket_name = self.sagemaker_session.default_bucket()
+        bucket_name_with_prefix = 's3://{}/'.format(bucket_name)
+        volumes = []
+
+        # Set up the channels for the containers. For local data we will
+        # mount the local directory to the container. For S3 Data we will download the S3 data
+        # first.
+        for channel in input_data_config:
+            uri = channel['DataSource']['S3DataSource']['S3Uri']
+            key = uri[len(bucket_name_with_prefix):]
+
+            channel_name = channel['ChannelName']
+            channel_dir = os.path.join(data_dir, channel_name)
+            os.mkdir(channel_dir)
+
+            if uri.lower().startswith("s3://"):
+                self._download_folder(bucket_name, key, channel_dir)
+            else:
+                volumes.append(Volume(uri, channel=channel_name))
+
+        # Create the docker compose files for each container that we will create
+        # Each container will map the additional local volumes (if any).
+        for host in self.hosts:
+            _prepare_config_file_directory(self.container_root, host)
+            self.write_config_files(host, hyperparameters, input_data_config)
+            shutil.copytree(data_dir, os.path.join(self.container_root, host, 'input', 'data'))
+
+        compose_data = self._generate_compose_file('train', additional_volumes=volumes)
+        compose_command = self._compose()
+        _stream_output(compose_command)
+        _cleanup()
+
+        s3_model_artifacts = self.retrieve_model_artifacts(compose_data)
+        return s3_model_artifacts
+
+    def serve(self, primary_container):
+        """Host a local endpoint using docker-compose.
+        Args:
+            primary_container (dict): dictionary containing the container runtime settings
+                for serving. Expected keys:
+                - 'ModelDataUrl' pointing to either an s3 location or a local file
+                - 'Environment' a dictionary of environment variables to be passed to the hosting container.
+
+        """
+        logger.info("serving")
+
+        self.container_root = _create_tmp_folder()
+        logger.info('creating hosting dir in {}'.format(self.container_root))
+
+        model_dir = primary_container['ModelDataUrl']
+        if not model_dir.lower().startswith("s3://"):
+            for h in self.hosts:
+                host_dir = os.path.join(self.container_root, h)
+                os.makedirs(host_dir)
+                shutil.copytree(model_dir, os.path.join(self.container_root, h, 'model'))
+
+        env_vars = ['{}={}'.format(k, v) for k, v in primary_container['Environment'].items()]
+
+        self._generate_compose_file('serve', additional_env_vars=env_vars)
+        compose_command = self._compose()
+        self.container = _Container(self.container_root, compose_command)
+        self.container.up()
+
+    def stop_serving(self):
+        """Stop the serving container.
+
+        The serving container runs in async mode to allow the SDK to do other tasks.
+        """
+        if self.container:
+            self.container.down()
+
+    def retrieve_model_artifacts(self, compose_data):
+        """Get the model artifacts from all the container nodes.
+
+        Used after training completes to gather the data from all the individual containers. As the
+        official SageMaker Training Service, it will override duplicate files if multiple containers have
+        the same file names.
+
+        Args:
+            compose_data(dict): Docker-Compose configuration in dictionary format.
+
+        Returns: Local path to the collected model artifacts.
+
+        """
+        # Grab the model artifacts from all the Nodes.
+        s3_model_artifacts = os.path.join(self.container_root, 's3_model_artifacts')
+        os.mkdir(s3_model_artifacts)
+
+        for host in self.hosts:
+            volumes = compose_data['services'][str(host)]['volumes']
+
+            for volume in volumes:
+                container_dir, host_dir = volume.split(':')
+                if host_dir == '/opt/ml/model':
+                    self._recursive_copy(container_dir, s3_model_artifacts)
+
+        return s3_model_artifacts
+
+    def write_config_files(self, host, hyperparameters, input_data_config):
+        """Write the config files for the training containers.
+
+        This method writes the hyperparameters, resources and input data configuration files.
+
+        Args:
+            host (str): Host to write the configuration for
+            hyperparameters (dict): Hyperparameters for training.
+            input_data_config (dict): Training input channels to be used for training.
+
+        Returns:
+
+        """
+
+        config_path = os.path.join(self.container_root, host, 'input', 'config')
+
+        resource_config = {
+            'current_host': host,
+            'hosts': self.hosts
+        }
+
+        json_input_data_config = {
+            c['ChannelName']: {'ContentType': 'application/octet-stream'} for c in input_data_config
+        }
+
+        _write_json_file(os.path.join(config_path, 'hyperparameters.json'), hyperparameters)
+        _write_json_file(os.path.join(config_path, 'resourceconfig.json'), resource_config)
+        _write_json_file(os.path.join(config_path, 'inputdataconfig.json'), json_input_data_config)
+
+    def _recursive_copy(self, src, dst):
+        for root, dirs, files in os.walk(src):
+            root = os.path.relpath(root, src)
+            current_path = os.path.join(src, root)
+            target_path = os.path.join(dst, root)
+
+            for file in files:
+                shutil.copy(os.path.join(current_path, file), os.path.join(target_path, file))
+            for dir in dirs:
+                new_dir = os.path.join(target_path, dir)
+                if not os.path.exists(new_dir):
+                    os.mkdir(os.path.join(target_path, dir))
+
+    def _download_folder(self, bucket_name, prefix, target):
+        session = self.sagemaker_session.boto_session
+
+        s3 = session.resource('s3')
+        bucket = s3.Bucket(bucket_name)
+
+        for obj_sum in bucket.objects.filter(Prefix=prefix):
+
+            obj = s3.Object(obj_sum.bucket_name, obj_sum.key)
+            file_path = os.path.join(target, obj_sum.key[len(prefix) + 1:])
+
+            try:
+                os.makedirs(os.path.dirname(file_path))
+            except os.error:
+                pass
+
+            obj.download_file(file_path)
+
+    def _generate_compose_file(self, command, additional_volumes=None, additional_env_vars=None):
+        session = self.sagemaker_session.boto_session
+        additional_env_vars = additional_env_vars or []
+        additional_volumes = additional_volumes or {}
+        environment = []
+        optml_dirs = set()
+
+        aws_creds = _aws_credentials(session)
+        if aws_creds is not None:
+            environment.extend(aws_creds)
+
+        environment.extend(additional_env_vars)
+
+        if command == 'train':
+            optml_dirs = {'output', 'input'}
+
+        services = {
+            h: self._create_docker_host(h, environment, optml_dirs,
+                                        command, additional_volumes) for h in self.hosts
+        }
+
+        content = {
+            # Some legacy hosts only support the 2.1 format.
+            'version': '2.1',
+            'services': services
+        }
+
+        filename = os.path.join(self.container_root, DOCKER_COMPOSE_FILENAME)
+        yaml_content = yaml.dump(content, default_flow_style=False)
+        logger.info('docker compose file: \n{}'.format(yaml_content))
+        with open(filename, 'w') as f:
+            f.write(yaml_content)
+
+        return content
+
+    def _compose(self, detached=False):
+        compose_cmd = 'nvidia-docker-compose' if self.instance_type == "local_gpu" else 'docker-compose'
+
+        command = [
+            compose_cmd,
+            '-f',
+            os.path.join(self.container_root, DOCKER_COMPOSE_FILENAME),
+            'up',
+            '--build',
+            '--abort-on-container-exit'
+        ]
+
+        if detached:
+            command.append('-d')
+
+        logger.info('docker command: {}'.format(' '.join(command)))
+        return command
+
+    def _create_docker_host(self, host, environment, optml_subdirs, command, volumes):
+        optml_volumes = self._optml_volumes(host, optml_subdirs)
+        optml_volumes.extend(volumes)
+
+        host_config = {
+            'image': self.image,
+            'stdin_open': True,
+            'tty': True,
+            'volumes': [v.map for v in optml_volumes],
+            'environment': environment,
+            'command': command
+        }
+
+        if command == 'serve':
+            host_config.update({
+                'ports': [
+                    '8080:8080'
+                ]
+            })
+
+        return host_config
+
+    def _optml_volumes(self, host, subdirs, single_model_dir=False):
+        """
+        It takes a folder with the necessary files for training and creates a list of opt volumes that
+        the Container needs to start.
+        If args.single_model_dir is True, all the hosts will point the opt/ml/model subdir to the first container.
+        That is useful for distributed training, so all the containers can read and write the same checkpoints.
+
+        :param opt_root_folder: root folder with the contents to be mapped to the container
+        :param host: host name of the container
+        :param subdirs: list of subdirs that will be mapped. Example: ['input', 'output', 'model']
+        :return:
+        """
+        volumes = []
+
+        # If it is single mode dir we want to map the same model dir and share between hosts
+        if single_model_dir:
+            host_dir = os.path.join(self.container_root, 'algo-1/model')
+            volume = Volume(host_dir, '/opt/ml/model')
+            volumes.append(volume)
+        else:
+            # else we want to add model to the list of subdirs so it will be created for each container.
+            subdirs.add('model')
+
+        for subdir in subdirs:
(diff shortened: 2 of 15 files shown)


src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+                'ports': [
+                    '8080:8080'
+                ]
+            })
+
+        return host_config
+
+    def _optml_volumes(self, host, subdirs, single_model_dir=False):
+        """
+        It takes a folder with the necessary files for training and creates a list of opt volumes that
```
owen-t: Creates a list of opt volumes required for the container to start. 

This function name could be improved - it's doing work - so call it build_optml_volumes?

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+        the Container needs to start.
+        If args.single_model_dir is True, all the hosts will point the opt/ml/model subdir to the first container.
+        That is useful for distributed training, so all the containers can read and write the same checkpoints.
+
+        :param opt_root_folder: root folder with the contents to be mapped to the container
```
owen-t: We're using the google format for parameters. Also - opt_root_folder isn't a parameter?

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+                ]
+            })
+
+        return host_config
+
+    def _optml_volumes(self, host, subdirs, single_model_dir=False):
+        """
+        It takes a folder with the necessary files for training and creates a list of opt volumes that
+        the Container needs to start.
+        If args.single_model_dir is True, all the hosts will point the opt/ml/model subdir to the first container.
```
owen-t: Sure, but this isn't reflective of actual distributed training. This feels like we're trying to compensate for a problem in SageMaker distributed learning. Why are we doing this? Can we not do this?

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+        for subdir in subdirs:
+            host_dir = os.path.join(self.container_root, host, subdir)
+            container_dir = '/opt/ml/{}'.format(subdir)
+            volume = Volume(host_dir, container_dir)
+            volumes.append(volume)
+
+        return volumes
+
+
+class _Container(object):
```
owen-t: Suggest not making this a private class and adding some API docs

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+    def up(self):
+        self._process = Popen(self.command)
+        sleep(self.startup_delay)
+
+    def down(self):
+        self._process.terminate()
+        _cleanup()
+
+
+class Volume(object):
```
owen-t: Add some API docs

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+
+        self.map = '{}:{}'.format(self.host_dir, self.container_dir)
+
+
+def _cleanup():
+    _chain_docker_cmds('docker images -f dangling=true -q', 'docker rmi -f')
+    _check_output('docker network prune -f'.split(' '))
+
+
+def _chain_docker_cmds(cmd, cmd2):
```
owen-t: Can you add some developer docs to this.

Also - is there any reason why this doesn't take an arg list, why just two commands?

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+    if any(docker_tags):
+        try:
+            _check_output(cmd2.split() + docker_tags, stderr=STDOUT)
+        except CalledProcessError:
+            pass
+
+
+def _stream_output(cmd):
+    if isinstance(cmd, str):
+        cmd = cmd.split(" ")
```
owen-t: problem here is if there's a double space you'll get empty strings in the list. you probably want just cmd.split()

That being said, this can still blow up. See - https://docs.python.org/2/library/shlex.html#shlex.split

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    exit_code = None
+    while exit_code is None:
+        stdout = process.stdout.readline().decode("utf-8")
+        sys.stdout.write(stdout)
+
```
owen-t: remove this whitespace

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+        try:
+            _check_output(cmd2.split() + docker_tags, stderr=STDOUT)
+        except CalledProcessError:
+            pass
+
+
+def _stream_output(cmd):
+    if isinstance(cmd, str):
+        cmd = cmd.split(" ")
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
```
owen-t: Is there a reason you can't just replace this call and loop with ```check_call()```

https://docs.python.org/2/library/subprocess.html#subprocess.check_call

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+        exit_code = process.poll()
+
+    if exit_code != 0:
+        raise Exception("Failed to run %s, exit code: %s" % (",".join(cmd), exit_code))
+
+    return exit_code
+
+
+def _check_output(cmd, *popenargs, **kwargs):
+    if isinstance(cmd, str):
```
owen-t: again, don't do this for reasons stated above

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+
+        exit_code = process.poll()
+
+    if exit_code != 0:
+        raise Exception("Failed to run %s, exit code: %s" % (",".join(cmd), exit_code))
+
+    return exit_code
+
+
+def _check_output(cmd, *popenargs, **kwargs):
```
owen-t: Do you even need this function - why not just call subprocess.check_output directly?

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+    return output
+
+
+def _create_tmp_folder():
+    tmp = tempfile.mkdtemp()
+    os.mkdir(os.path.join(tmp, 'output'))
+
+    # Docker cannot mount Mac OS /var folder properly see
+    # https://forums.docker.com/t/var-folders-isnt-mounted-properly/9600
+    dir = '/private{}'.format(tmp) if platform.system() == 'Darwin' else tmp
```
owen-t: This won't work on Windows :-/

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+    return os.path.abspath(dir)
+
+
+def _prepare_config_file_directory(root, host):
```
owen-t: this creates the config directories. Can methods / functions in this file that do similar things be named in a similar way. E.g. if you are creating something - call it create_config_file_directory

src/sagemaker/local_session.py
```diff
@@ -0,0 +1,122 @@
...
+from botocore.exceptions import ClientError
+
+from sagemaker.image import SageMakerContainer
+from sagemaker.session import Session
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+
+class LocalSagemakerClient(object):
```
owen-t: Add api docs. All public members should have API docs, won't repeat this per member.

src/sagemaker/local_session.py
```diff
@@ -0,0 +1,122 @@
...
+        self.serve_container = SageMakerContainer(instance_type, instance_count,
+                                                  self.primary_container['Image'], self.sagemaker_session)
+        self.serve_container.serve(self.primary_container)
+        self.created_endpoint = True
+
+        i = 0
+        http = urllib3.PoolManager()
+        while True:
+            i += 1
+
```
owen-t: remove this whitespace

src/sagemaker/local_session.py
```diff
@@ -0,0 +1,122 @@
...
+        self.http = urllib3.PoolManager()
+
+    def invoke_endpoint(self, Body, EndpointName, ContentType, Accept):
+        r = self.http.request('POST', "http://localhost:8080/invocations", body=Body, preload_content=False,
+                              headers={'Content-type': ContentType, 'Accept': Accept})
+
+        return {'Body': r, 'ContentType': Accept}
+
+
+class LocalSession(Session):
```
owen-t: Add API docs

src/sagemaker/session.py
```diff
@@ -756,6 +756,32 @@ def __init__(self, s3_data, distribution='FullyReplicated', compression=None,
             self.config['RecordWrapperType'] = record_wrapping
 
 
+class Inputs(object):
```
owen-t: Add API docs

src/sagemaker/image.py
```diff
@@ -0,0 +1,446 @@
...
+        s3_model_artifacts = os.path.join(self.container_root, 's3_model_artifacts')
+        os.mkdir(s3_model_artifacts)
+
+        for host in self.hosts:
+            volumes = compose_data['services'][str(host)]['volumes']
+
+            for volume in volumes:
+                container_dir, host_dir = volume.split(':')
+                if host_dir == '/opt/ml/model':
+                    self._recursive_copy(container_dir, s3_model_artifacts)
```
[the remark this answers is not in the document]
iquintero: because shutil.copytree assumes that the target directory does not exist. When doing distributed training, the target dir will exist after getting the data out of the first container.

src/sagemaker/image.py
iquintero: Sorry! this was leftover from the experimental code and I didn't update these docs when I refactored. Will take care of it!.

src/sagemaker/image.py
iquintero: This method streams the output as it happens,  check_call() waits for the command to complete and then gets the output as once. This makes the training customer experience bad because they only get to see the logs after the containers exit, which could be a long time depending on the training job.

src/sagemaker/image.py
iquintero: Got rid of this function. again this was carry over that I didn't touch but I got rid of it now, it doesn't make much sense and it doesn't 'chain' docker commands.

iquintero:
I have just posted an update. I have added a configuration file


$HOME/.sagemaker/config.yaml

$ cat ~/.sagemaker/config.yaml
local:
    container_root: /Users/nacho/container_root
    serving_port: 8081

I will add docs regarding this to the README. It is completely optional and if it doesn't exist nothing will happen.

I improved the cleanup  of the containers and in general got rid of some methods that didn't make sense. Now we have support to specify a root for all the container data.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+            sagemaker_session (sagemaker.session.Session): a sagemaker session to use when interacting
+                with SageMaker.
+        """
+        from sagemaker.local_session import LocalSession
+        self.sagemaker_session = sagemaker_session or LocalSession()
+        self.instance_type = instance_type
+        self.instance_count = instance_count
+        self.image = image
+        self.hosts = ['{}-{}'.format(CONTAINER_PREFIX, i) for i in range(1, self.instance_count + 1)]
+        self.container_root = None
```
owen-t: container_root and container could do with some developer documentation explaining what they are

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        _cleanup()
+
+        s3_model_artifacts = self.retrieve_model_artifacts(compose_data)
+
+        # free up the training data directory as it may contain
+        # lots of data downloaded from S3. This doesn't delete any local
+        # data that was just mounted to the container.
+        shutil.rmtree(data_dir)
+        # Also free the container config files.
+        for host in self.hosts:
```
owen-t: What's left in self.container_root after this? could we / should we just delete self.container_root at this point?

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        self.hosts = ['{}-{}'.format(CONTAINER_PREFIX, i) for i in range(1, self.instance_count + 1)]
+        self.container_root = None
+        self.container = None
+        # set the local config. This is optional and will use reasonable defaults
+        # if not present.
+        self.local_config = None
+        if self.sagemaker_session.config and 'local'in self.sagemaker_session.config:
+            self.local_config = self.sagemaker_session.config['local']
+
+    def train(self, input_data_config, hyperparameters):
```
owen-t: When this is refactored into a TrainingSageMakerContainer - these should moved into the initializer method.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+            input_data_config (dict): The Input Data Configuration, this contains data such as the
+                channels to be used for training.
+            hyperparameters (dict): The HyperParameters for the training job.
+
+        Returns (str): Location of the trained model.
+
+        """
+        self.container_root = self._create_tmp_folder()
+
+        data_dir = tempfile.mkdtemp()
```
owen-t: I'm confused why this doesn't use self._create_tmp_folder(). We want to give a customer a way to control the directory that s3 data is downloaded to.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        # Each container will map the additional local volumes (if any).
+        for host in self.hosts:
+            _create_config_file_directories(self.container_root, host)
+            self.write_config_files(host, hyperparameters, input_data_config)
+            shutil.copytree(data_dir, os.path.join(self.container_root, host, 'input', 'data'))
+
+        compose_data = self._generate_compose_file('train', additional_volumes=volumes)
+        compose_command = self._compose()
+        _stream_output(compose_command)
+        _cleanup()
```
owen-t: This cleanup and the other cleanup commands should be merged into a single cleanup method. This should be called in a finally block of this method.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+
+        json_input_data_config = {
+            c['ChannelName']: {'ContentType': 'application/octet-stream'} for c in input_data_config
+        }
+
+        _write_json_file(os.path.join(config_path, 'hyperparameters.json'), hyperparameters)
+        _write_json_file(os.path.join(config_path, 'resourceconfig.json'), resource_config)
+        _write_json_file(os.path.join(config_path, 'inputdataconfig.json'), json_input_data_config)
+
+    def _recursive_copy(self, src, dst):
```
owen-t: This shouldn't be a method as self is not used. Make this a function.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+            target_path = os.path.join(dst, root)
+
+            for file in files:
+                shutil.copy(os.path.join(current_path, file), os.path.join(target_path, file))
+            for dir in dirs:
+                new_dir = os.path.join(target_path, dir)
+                if not os.path.exists(new_dir):
+                    os.mkdir(os.path.join(target_path, dir))
+
+    def _download_folder(self, bucket_name, prefix, target):
```
owen-t: This is only weakly needed to be a method - consider making it a function that takes in a boto_session

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        self.container.up()
+
+    def stop_serving(self):
+        """Stop the serving container.
+
+        The serving container runs in async mode to allow the SDK to do other tasks.
+        """
+        if self.container:
+            self.container.down()
+        shutil.rmtree(self.container_root)
```
owen-t: This should be in a cleanup method.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+                shutil.copytree(model_dir, os.path.join(self.container_root, h, 'model'))
+
+        env_vars = ['{}={}'.format(k, v) for k, v in primary_container['Environment'].items()]
+
+        self._generate_compose_file('serve', additional_env_vars=env_vars)
+        compose_command = self._compose()
+        self.container = Container(self.container_root, compose_command)
+        self.container.up()
+
+    def stop_serving(self):
```
owen-t: When this refactored into its own SageMakerHostingContainer, consider making this be a context manager.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        """Stop the serving container.
+
+        The serving container runs in async mode to allow the SDK to do other tasks.
+        """
+        if self.container:
+            self.container.down()
+        shutil.rmtree(self.container_root)
+
+    def retrieve_model_artifacts(self, compose_data):
```
owen-t: It's unlikely that someone is going to know how to call this method on an instance of the class, given the need for compose_data to be passed. 

This method will also cease to work if we do the cleanup in a finally. 

So - this is a usage question - do we want a client of this class to be able to call retrieve_model_artifacts after train has completed? I think not, because train returns the model artifacts. 

I'd consider getting rid of this method completely and moving the contents into train

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+            volumes = compose_data['services'][str(host)]['volumes']
+
+            for volume in volumes:
+                container_dir, host_dir = volume.split(':')
+                if host_dir == '/opt/ml/model':
+                    self._recursive_copy(container_dir, s3_model_artifacts)
+
+        return s3_model_artifacts
+
+    def write_config_files(self, host, hyperparameters, input_data_config):
```
owen-t: This should be a private method - it's unlikely a client is going to have a reason to call this. Also - the need to pass in a host argument is difficult. It can't be called at any time outside of a train or serve command - so make it private.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+
+        filename = os.path.join(self.container_root, DOCKER_COMPOSE_FILENAME)
+        yaml_content = yaml.dump(content, default_flow_style=False)
+        logger.info('docker compose file: \n{}'.format(yaml_content))
+        with open(filename, 'w') as f:
+            f.write(yaml_content)
+
+        return content
+
+    def _compose(self, detached=False):
```
owen-t: Can this be named ``_build_compose_command``

src/sagemaker/local_session.py
```diff
@@ -0,0 +1,160 @@
...
+        self.train_container = None
+        self.serve_container = None
+        self.sagemaker_session = sagemaker_session or LocalSession()
+        self.s3_model_artifacts = None
+        self.model_name = None
+        self.primary_container = None
+        self.role_arn = None
+        self.created_endpoint = False
+
+    def create_training_job(self, TrainingJobName, AlgorithmSpecification, RoleArn, InputDataConfig, OutputDataConfig,
```
owen-t: We need to decide if we want to support multiple training jobs or not.

If we expect customers to only do local training via estimators, then we don't. If we expect customers to use a LocalSession to do training, then we do. Create an issue for this, should frame the ambiguity.

The same applies to the hosting methods as well, for the different hosting entities - endpoint, model, endpoint config. Create an issue for this, again framing ambiguity.

Even if we only support one endpoint per session (as it is now), we need to be able to run multiple endpoints in different sessions concurrently. This means we need to flexible with ports in a way that we aren't now (I think). Create an issue for this.

src/sagemaker/estimator.py
```diff
@@ -78,7 +79,17 @@ def __init__(self, role, train_instance_count, train_instance_type,
         self.train_volume_size = train_volume_size
         self.train_max_run = train_max_run
         self.input_mode = input_mode
-        self.sagemaker_session = sagemaker_session or Session()
+
```
owen-t: So - we've made changes to support estimators, but what about models? If I have existing (local) model data, I should be able to deploy locally without training. Create an issue for this. This may already be supported - but I don't see tests for it.

I should be able to train two containers simultaneously on the same machine. I don't see a test for this. In future - can we add a test for that or explicitly forbid it. Create an issue for this.

src/sagemaker/estimator.py
```diff
@@ -78,7 +79,17 @@ def __init__(self, role, train_instance_count, train_instance_type,
...
+        if self.train_instance_type in ('local', 'local_gpu'):
+            self.local_mode = True
+            if self.train_instance_type == 'local_gpu' and self.train_instance_count > 1:
```
owen-t: Create an issue to support distributed local gpu in future, but that's going to be tough to do if the local machine only has a single gpu.

src/sagemaker/image.py
iquintero: The only thing left is the model. I didn't want to delete it because hosting will consume it locally, and it may be of interest for the customer afterwards. There is nothing else left after the whole thing runs.

src/sagemaker/image.py
iquintero: I agree with this, the only reason I made this public is because I wanted to have unit tests for it.

src/sagemaker/local_session.py
iquintero: You are right, the port as of now is configurable but its not flexible at runtime. This is one of the things I considered improving as it should be fairly transparent to the users

src/sagemaker/image.py
iquintero: Good point. I will change this.

src/sagemaker/estimator.py
iquintero: On the first point, it is not entirely possible yet. 

It is possible to train 2 containers simultaneously as they don't perform any action that is mutually exclusive. The exception to this would be training with GPU as there is a single hardware bottleneck.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+            volume = Volume(host_dir, container_dir)
+            volumes.append(volume)
+
+        return volumes
+
+
+class Container(object):
+    def __init__(self, tmpdir, command, startup_delay=5):
+        self.command = command
+        self.compose_file = os.path.join(tmpdir, DOCKER_COMPOSE_FILENAME)
```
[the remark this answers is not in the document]
iquintero: Disappeared !

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+    docker_tags = _check_output('docker images -f dangling=true -q').split('\n')
+    docker_clean = 'docker rmi -f' + ' '.join(docker_tags)
+    try:
+        _check_output(docker_clean)
+    except CalledProcessError:
+        pass
+    _check_output('docker network prune -f')
+
+
+def _stream_output(cmd):
```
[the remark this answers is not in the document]
iquintero: Done.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+                                        command, additional_volumes) for h in self.hosts
+        }
+
+        content = {
+            # Some legacy hosts only support the 2.1 format.
+            'version': '2.1',
+            'services': services
+        }
+
+        filename = os.path.join(self.container_root, DOCKER_COMPOSE_FILENAME)
```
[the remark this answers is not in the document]
iquintero: Done.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+        Args:
+            command (str): either 'train' or 'serve'
+            additional_volumes (list): a list of volumes that will be mapped to the containers
+            additional_env_vars (dict): a dictionary with additional environment variables to be
+                passed on to the containers.
+
+        Returns: (dict) A dictionary representation of the configuration that was written.
+
+        """
+        session = self.sagemaker_session.boto_session
```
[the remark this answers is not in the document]
iquintero: Done.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+                self._download_folder(bucket_name, key, channel_dir)
+            else:
+                volumes.append(Volume(uri, channel=channel_name))
+
+        # Create the configuration files for each container that we will create
+        # Each container will map the additional local volumes (if any).
+        for host in self.hosts:
+            _create_config_file_directories(self.container_root, host)
+            self.write_config_files(host, hyperparameters, input_data_config)
+            shutil.copytree(data_dir, os.path.join(self.container_root, host, 'input', 'data'))
```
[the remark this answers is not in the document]
iquintero: Done.

src/sagemaker/image.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def _cleanup():
+    docker_tags = _check_output('docker images -f dangling=true -q').split('\n')
+    docker_clean = 'docker rmi -f' + ' '.join(docker_tags)
+    try:
+        _check_output(docker_clean)
+    except CalledProcessError:
+        pass
+    _check_output('docker network prune -f')
```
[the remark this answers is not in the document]
iquintero: network prune will only get rid of networks that are unused.

src/sagemaker/image.py
owen-t: okay - in future - can we revisit this, because this is in a temporary directory, it may not be easy for a customer to access this. Let's create a SIM for this.

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+        for subdir in subdirs:
+            host_dir = os.path.join(self.container_root, host, subdir)
+            container_dir = '/opt/ml/{}'.format(subdir)
+            volume = _Volume(host_dir, container_dir)
+            volumes.append(volume)
+
+        return volumes
+
+    def _cleanup(self):
+        _check_output('docker network prune -f')
```
owen-t: In future - let's keep track of the networks created and just remove those.

owen-t:
Let's also open an issue for billable seconds - should be None

src/sagemaker/local/local_session.py
```diff
@@ -0,0 +1,168 @@
...
+                    logger.info("Container still not up, got: %s" % r.status)
+                else:
+                    return
+            except urllib3.exceptions.RequestError:
+                logger.info("Container still not up")
+
+            time.sleep(1)
+
+    def delete_endpoint(self, EndpointName):
+        self.serve_container.stop_serving()
```
mvsusp: serve_container is an instance of ```_SageMakerContainer``` and stop_serving will call invoke ```serve_container.container.down()```. Is it necessary to have these two container classes? Can we review the abstraction here?

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+            self._cleanup()
+        # for serving we can delete everything in the container root.
+        shutil.rmtree(self.container_root)
+
+    def retrieve_model_artifacts(self, compose_data):
+        """Get the model artifacts from all the container nodes.
+
+        Used after training completes to gather the data from all the individual containers. As the
+        official SageMaker Training Service, it will override duplicate files if multiple containers have
+        the same file names.
```
mvsusp: I believe that we should save the model in the S3 bucket in the same way that the model is saved in SageMaker Training Service. My reasoning is that other services in the future may rely with the model in S3.
We can provide methods to allow the customer to easily download the models but that is a nice to have.

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+            optml_dirs = {'output', 'input'}
+
+        services = {
+            h: self._create_docker_host(h, environment, optml_dirs,
+                                        command, additional_volumes) for h in self.hosts
+        }
+
+        content = {
+            # Some legacy hosts only support the 2.1 format.
+            'version': '2.1',
```
mvsusp: Are you sure that we still need to support the old 2.1 format here?

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+        host_config = {
+            'image': self.image,
+            'stdin_open': True,
+            'tty': True,
+            'volumes': [v.map for v in optml_volumes],
+            'environment': environment,
+            'command': command
+        }
+
+        serving_port = 8080 if self.local_config is None else self.local_config.get('serving_port', 8080)
```
mvsusp: You can revert this if:
```python
serving_port = self.local_config.get('serving_port', 8080) if self.local_config else 8080
```

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+            input_data_config (dict): The Input Data Configuration, this contains data such as the
+                channels to be used for training.
+            hyperparameters (dict): The HyperParameters for the training job.
+
+        Returns (str): Location of the trained model.
+        """
+        self.container_root = self._create_tmp_folder()
+        os.mkdir(os.path.join(self.container_root, 'output'))
+
+        data_dir = self._create_tmp_folder()
```
mvsusp: data_dir can be a normal tmp directory since is not going to be a docker volume.

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+        """Run a training job locally using docker-compose.
+        Args:
+            input_data_config (dict): The Input Data Configuration, this contains data such as the
+                channels to be used for training.
+            hyperparameters (dict): The HyperParameters for the training job.
+
+        Returns (str): Location of the trained model.
+        """
+        self.container_root = self._create_tmp_folder()
+        os.mkdir(os.path.join(self.container_root, 'output'))
```
mvsusp: Should we move the output folder creation inside the ```_create_tmp_folder```?

src/sagemaker/local/image.py
```diff
@@ -0,0 +1,508 @@
...
+        cmd(str or List): either a string or a List (in Popen Format) with the command to execute.
+
+    Returns (int): command exit code
+    """
+    if isinstance(cmd, str):
+        cmd = shlex.split(cmd)
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    exit_code = None
+    while exit_code is None:
+        stdout = process.stdout.readline().decode("utf-8")
```
mvsusp: Does decode works with 2.7 and 3.6?

src/sagemaker/local/local_session.py
```diff
@@ -0,0 +1,168 @@
...
+        Args:
+            TrainingJobName (str): Not used in this implmentation.
+
+        Returns: (dict) DescribeTrainingJob Response.
+
+        """
+        response = {'ResourceConfig': {'InstanceCount': self.train_container.instance_count},
+                    'TrainingJobStatus': 'Completed',
+                    'TrainingStartTime': datetime.datetime.now(),
+                    'TrainingEndTime': datetime.datetime.now(),
```
mvsusp: This another thing that we need to review soon. The timing and status returned should comply with the docker container execution.

mvsusp:
We need tests for the following scenarios:
Async training
Attachment 
Serving from a model in S3
Serving and training errors

PR #117, opened 2018-04-02
Local mode improvements

iquintero:
Switch local mode to use nvidia-docker2 + docker-compose
instead of nvidia-docker + nvidia-docker-compose.

Also get rid of the log output for billable seconds in local mode as it
doesn't make sense.

Finally, let the users know if we can't cleanup a container directory
but not fail the training job.

### DIFF
--- a/src/sagemaker/local/image.py
+++ b/src/sagemaker/local/image.py
@@ -488,7 +488,10 @@ def _delete_tree(path):
     try:
         shutil.rmtree(path)
     except OSError as exc:
-        # handle permission denied gracefully. Any other error we will raise the exception up.
+        # on Linux, when docker writes to any mounted volume, it uses the container's user. In most cases
+        # this is root. When the container exits and we try to delete them we can't because root owns those
+        # files. We expect this to happen, so we handle EACCESS. Any other error we will raise the
+        # exception up.
         if exc.errno == errno.EACCES:
             logger.warning("Failed to delete: %s Please remove it manually." % path)
         else:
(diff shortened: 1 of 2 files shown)


src/sagemaker/local/image.py
```diff
@@ -120,9 +120,16 @@ def train(self, input_data_config, hyperparameters):
         shutil.rmtree(data_dir)
         # Also free the container config files.
         for host in self.hosts:
-            shutil.rmtree(os.path.join(self.container_root, host))
+            container_config_path = os.path.join(self.container_root, host)
+            try:
```
winstonaws: Can you factor out this cleanup logic into a function? We don't know how long it'll stay around, and it might evolve over time, and we should keep it consistent between the various places.

src/sagemaker/local/image.py
```diff
@@ -120,9 +120,16 @@ def train(self, input_data_config, hyperparameters):
...
+                shutil.rmtree(container_config_path)
+            except OSError:
```
winstonaws: There are probably other ways an OSError can be thrown here, right? I don't love catching a fairly generic exception type and swallowing it without showing a stacktrace.

I think we could go two directions here - be more specific about what we catch, or just catch everything and print out this error message with a stacktrace / enough information to debug. I'm guessing there's not a great way to do the first, so maybe we should do the second? Are there any cases where we would want to rethrow an exception because we failed cleanup?

src/sagemaker/local/image.py
```diff
@@ -490,6 +484,18 @@ def _create_config_file_directories(root, host):
         os.makedirs(os.path.join(root, host, d))
 
 
+def _delete_tree(path):
+    try:
+        shutil.rmtree(path)
+    except OSError as exc:
+        # handle permission denied gracefully. Any other error we will raise the exception up.
```
winstonaws: Put a short explanation of the docker writing-files-as-root issue here in a comment.

PR #118, opened 2018-04-02
Update default versions to TensorFlow 1.6 and MXNet 1.1

winstonaws:
Also bumping our minor version up, since TensorFlow 1.6 / MXNet 1.1 may have some backwards incompatible API changes.

Also includes a minor change to integ tests - removed the framework version parameter from the TF/MXNet async fit tests. These tests cover the same cases as their synchronous fit tests.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,10 +5,10 @@ CHANGELOG
 1.2.0
 ========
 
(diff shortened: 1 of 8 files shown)


owen-t:
How do customers find out the versions of TF or mxnet we support?

PR #121, opened 2018-04-02
Login to ECR if needed for Local Mode

iquintero:
This also includes:

Use 'sagemaker-local' docker network for LocalMode

    Instead of having docker-compose generate a random network each time,
    use a constant network.

### DIFF
--- a/src/sagemaker/local/image.py
+++ b/src/sagemaker/local/image.py
@@ -335,7 +335,8 @@ class _SageMakerContainer(object):
             'tty': True,
             'volumes': [v.map for v in optml_volumes],
             'environment': environment,
-            'command': command
+            'command': command,
+            'network_mode': 'bridge'
         }
 
         serving_port = 8080 if self.local_config is None else self.local_config.get('serving_port', 8080)
--- a/src/sagemaker/local/image.py
+++ b/src/sagemaker/local/image.py
@@ -15,8 +15,10 @@ import json
 import logging
 import os
 import platform
+import random
 import shlex
 import shutil
+import string
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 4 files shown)


src/sagemaker/local/image.py
```diff
@@ -390,7 +409,8 @@ def _build_optml_volumes(self, host, subdirs):
         return volumes
 
     def _cleanup(self):
-        _check_output('docker network prune -f')
+        # we don't need to cleanup anything at the moment
+        pass
```
mvsusp: Add a TODO (rignacio) here to remove this function later.

src/sagemaker/local/image.py
```diff
@@ -525,3 +545,24 @@ def _aws_credentials(session):
...
+    ecr = boto_session.client('ecr')
+    auth = ecr.get_authorization_token(registryIds=[image.split('.')[0]])
+    authorization_data = auth['authorizationData'][0]
+
+    raw_token = base64.b64decode(authorization_data['authorizationToken'])
+    token = raw_token.decode('utf-8').strip('AWS:')
+    ecr_url = auth['authorizationData'][0]['proxyEndpoint']
+
+    cmd = "docker login -u AWS -p %s %s" % (token, ecr_url)
+    subprocess.check_output(cmd, shell=True)
```
mvsusp: Sometimes we are using **_check_output** and sometimes we are using **subprocess.checkoutput**. Can we just use one of them?

src/sagemaker/local/image.py
mvsusp: Them add a TODO (rignacio) here to write this function later :)

PR #167, opened 2018-05-04
Mark tests for continuous testing

laurenyu:
*Description of changes:*
Mark tests so that when we run tests continuously, we can use `pytest -v -m continuous_testing` to run only certain tests.

## Merge Checklist

### DIFF
--- a/tests/integ/test_mxnet_train.py
+++ b/tests/integ/test_mxnet_train.py
@@ -52,6 +52,7 @@ def test_attach_deploy(mxnet_training_job, sagemaker_session):
         predictor.predict(data)
 
 
+@pytest.mark.continuous_testing
 def test_deploy_model(mxnet_training_job, sagemaker_session):
     endpoint_name = 'test-mxnet-deploy-model-{}'.format(sagemaker_timestamp())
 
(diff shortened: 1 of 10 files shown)


tests/integ/test_mxnet_train.py
```diff
@@ -52,6 +52,7 @@ def test_attach_deploy(mxnet_training_job, sagemaker_session):
         predictor.predict(data)
 
 
+@pytest.mark.continuous_testing
 def test_deploy_model(mxnet_training_job, sagemaker_session):
```
nadiaya: why that one?

laurenyu: I think it tests a more straightforward code path than `test_attach_deploy`, though I could be persuaded to pick the other one since our TF test already tests the "straightforward" approach.

laurenyu: talked offline - will switch to `test_attach_deploy` since deploying through the estimator rather than creating a separate model object is probably more common

nadiaya: It doesn't really matter, i just find this test to be more unnatural, that's why was curious about the choice process :)
Approved.

PR #217, opened 2018-06-07
Allow Local Serving of Models in S3

iquintero:
if a model is located in S3, download and extract it and then start
serving as usual. By overriding the SageMaker Session on any Model
Object, local serving is possible on pre-trained models.

*Issue #, if available:*

*Description of changes:*

## Merge Checklist

### DIFF
--- a/src/sagemaker/local/image.py
+++ b/src/sagemaker/local/image.py
@@ -24,6 +24,7 @@ import shutil
 import string
 import subprocess
 import sys
+import tarfile
 import tempfile
 from fcntl import fcntl, F_GETFL, F_SETFL
 from six.moves.urllib.parse import urlparse
@@ -137,7 +138,7 @@ class _SageMakerContainer(object):
         Args:
             primary_container (dict): dictionary containing the container runtime settings
                 for serving. Expected keys:
-                - 'ModelDataUrl' pointing to a local file
+                - 'ModelDataUrl' pointing to a file or s3:// location.
                 - 'Environment' a dictionary of environment variables to be passed to the hosting container.
 
         """
@@ -147,22 +148,17 @@ class _SageMakerContainer(object):
         logger.info('creating hosting dir in {}'.format(self.container_root))
 
         model_dir = primary_container['ModelDataUrl']
-        if not model_dir.lower().startswith("s3://"):
-            for h in self.hosts:
-                host_dir = os.path.join(self.container_root, h)
-                os.makedirs(host_dir)
-                shutil.copytree(model_dir, os.path.join(self.container_root, h, 'model'))
-
+        volumes = self._prepare_serving_volumes(model_dir)
         env_vars = ['{}={}'.format(k, v) for k, v in primary_container['Environment'].items()]
 
-        _ecr_login_if_needed(self.sagemaker_session.boto_session, self.image)
-
         # If the user script was passed as a file:// mount it to the container.
-        script_dir = primary_container['Environment'][sagemaker.estimator.DIR_PARAM_NAME.upper()]
-        parsed_uri = urlparse(script_dir)
-        volumes = []
-        if parsed_uri.scheme == 'file':
-            volumes.append(_Volume(parsed_uri.path, '/opt/ml/code'))
+        if sagemaker.estimator.DIR_PARAM_NAME.upper() in primary_container['Environment']:
+            script_dir = primary_container['Environment'][sagemaker.estimator.DIR_PARAM_NAME.upper()]
+            parsed_uri = urlparse(script_dir)
+            if parsed_uri.scheme == 'file':
+                volumes.append(_Volume(parsed_uri.path, '/opt/ml/code'))
+
+        _ecr_login_if_needed(self.sagemaker_session.boto_session, self.image)
 
         self._generate_compose_file('serve',
                                     additional_env_vars=env_vars,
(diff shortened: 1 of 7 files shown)


src/sagemaker/local/image.py
```diff
@@ -453,8 +488,8 @@ def _build_optml_volumes(self, host, subdirs):
         volumes = []
 
         # Ensure that model is in the subdirs
-        if 'model' not in subdirs:
-            subdirs.add('model')
+        #if 'model' not in subdirs:
```
iquintero: will get rid of this

tests/integ/test_local_mode.py
```diff
@@ -230,6 +251,53 @@ def test_tf_local_data_local_script():
...
+    try:
+        # Since Local Mode uses the same port for serving, we need a lock in order
+        # to allow concurrent test execution. The serving test is really fast so it still
+        # makes sense to allow this behavior.
+        fcntl.lockf(local_mode_lock, fcntl.LOCK_EX)
+        predictor = s3_model.deploy(initial_instance_count=1, instance_type='local')
+        data = numpy.zeros(shape=(1, 1, 28, 28))
+        predictor.predict(data)
+    finally:
+        mxnet_model.sagemaker_session.delete_endpoint('')
```
winstonaws: Is the empty string passed here because that's the name we explicitly chose for the endpoint? 

Can we add delete_endpoint to the predictor class to make this more intuitive? That's where it should be anyway, IMO.

tests/integ/test_local_mode.py
```diff
@@ -54,6 +56,25 @@ def _initialize(self, boto_session, sagemaker_client, sagemaker_runtime_client):
...
+               train_instance_count=1, train_instance_type='local',
+               sagemaker_session=sagemaker_local_session)
+
+    train_input = mx.sagemaker_session.upload_data(path=os.path.join(data_path, 'train'),
+                                                   key_prefix='integ-test-data/mxnet_mnist/train')
+    test_input = mx.sagemaker_session.upload_data(path=os.path.join(data_path, 'test'),
+                                                  key_prefix='integ-test-data/mxnet_mnist/test')
+
+    mx.fit({'train': train_input, 'test': test_input})
+    model = mx.create_model(1)
```
winstonaws: Is the argument necessary here?

tests/integ/test_local_mode.py
```diff
@@ -230,6 +251,53 @@ def test_tf_local_data_local_script():
...
 
 
+def test_local_mode_serving_from_s3_model(sagemaker_local_session, mxnet_model):
+    local_mode_lock_fd = open(LOCK_PATH, 'w')
+    local_mode_lock = local_mode_lock_fd.fileno()
+
+    model_data = mxnet_model.model_data
+    boto_session = sagemaker_local_session.boto_session
+    default_bucket = sagemaker_local_session.default_bucket()
+    uploaded_data = tar_and_upload_dir(boto_session, default_bucket,
```
winstonaws: This is only necessary to create the model data, right? In the normal use cases, the user would either do both training and deploy locally, or create the MXNetModel pointing to local model data, or already have an S3 url with model data, right?

tests/integ/test_local_mode.py
```diff
@@ -230,6 +251,53 @@ def test_tf_local_data_local_script():
...
+                                       'test_mxnet_local_mode', '', model_data)
+
+    s3_model = MXNetModel(model_data=uploaded_data.s3_prefix, role='SageMakerRole',
+                          entry_point=mxnet_model.entry_point, image=mxnet_model.image,
+                          sagemaker_session=sagemaker_local_session)
```
winstonaws: Is it necessary to explicitly pass a LocalSession here, or is that only for our tests?

tests/integ/test_local_mode.py
```diff
@@ -230,6 +251,53 @@ def test_tf_local_data_local_script():
...
+    local_mode_lock_fd = open(LOCK_PATH, 'w')
+    local_mode_lock = local_mode_lock_fd.fileno()
+
+    model_data = mxnet_model.model_data
+    boto_session = sagemaker_local_session.boto_session
+    default_bucket = sagemaker_local_session.default_bucket()
+    uploaded_data = tar_and_upload_dir(boto_session, default_bucket,
+                                       'test_mxnet_local_mode', '', model_data)
+
+    s3_model = MXNetModel(model_data=uploaded_data.s3_prefix, role='SageMakerRole',
```
winstonaws: Is role still necessary here?

src/sagemaker/local/image.py
```diff
@@ -147,22 +148,17 @@ def serve(self, primary_container):
...
 
-        _ecr_login_if_needed(self.sagemaker_session.boto_session, self.image)
-
         # If the user script was passed as a file:// mount it to the container.
-        script_dir = primary_container['Environment'][sagemaker.estimator.DIR_PARAM_NAME.upper()]
-        parsed_uri = urlparse(script_dir)
-        volumes = []
-        if parsed_uri.scheme == 'file':
-            volumes.append(_Volume(parsed_uri.path, '/opt/ml/code'))
+        if sagemaker.estimator.DIR_PARAM_NAME.upper() in primary_container['Environment']:
```
winstonaws: What's the case where this is false?

tests/integ/test_local_mode.py
iquintero: yeah, this is because I wanted to make this faster by training using local mode instead of training on sagemaker like the chainer tests do for instance.

tests/integ/test_local_mode.py
iquintero: its not necessary, but the estimator expects one. For SageMaker this is a must have, on local mode its not even used but I dont know if we should make this optional.

src/sagemaker/local/image.py
iquintero: BYO :)  we only have sagemaker_submit_dir on our Frameworks.

tests/integ/test_local_mode.py
iquintero: It used to be required, now its only for our tests.

src/sagemaker/predictor.py
```diff
@@ -93,6 +93,12 @@ def predict(self, data):
         response_body.close()
         return data
 
+    def delete_endpoint(self):
+        def delete_endpoint(self):
```
winstonaws: Why is there an inner function here? 😅

src/sagemaker/predictor.py
```diff
@@ -93,6 +93,12 @@ def predict(self, data):
...
+            """Delete an Amazon SageMaker ``Endpoint``.
```
winstonaws: How about "Delete the Amazon SageMaker endpoint backing this predictor" or similar?

src/sagemaker/predictor.py
iquintero: its unclear XD. Fixed

src/sagemaker/predictor.py
iquintero: Will do

PR #223, opened 2018-06-11
Allow Framework Estimators to use custom image

iquintero:
Chainer, Tensorflow and MXNet estimators can now pass
an image_name argument to the constructor to use that image
instead of the default sagemaker ones.

*Issue #, if available:*

*Description of changes:*

## Merge Checklist

### DIFF
--- a/src/sagemaker/chainer/README.rst
+++ b/src/sagemaker/chainer/README.rst
@@ -175,6 +175,12 @@ The following are optional arguments. When you create a ``Chainer`` object, you 
 -  ``job_name`` Name to assign for the training job that the fit()
    method launches. If not specified, the estimator generates a default
    job name, based on the training image name and current timestamp
+-  ``image_name`` An alternative docker image to use for training and
+    serving.  If specified, the estimator will use this image for training and
+    hosting, instead of selecting the appropriate SageMaker official image based on
+    framework_version and py_version. Refer to: `SageMaker Chainer Docker Containers
+    <#sagemaker-chainer-docker-containers>`_ for details on what the Official images support
+    and where to find the source code to build your custom image.
 
 
 Distributed Chainer Training
@@ -655,6 +661,9 @@ You can select version of Chainer by passing a framework_version keyword arg to 
 Currently supported versions are listed in the above table. You can also set framework_version to only specify major and
 minor version, which will cause your training script to be run on the latest supported patch version of that minor
 version.
+
+Alternatively, you can build your own image by following the instructions in the SageMaker Chainer containers
+repository, and passing ``image_name`` to the Chainer Estimator constructor.
 
 You can visit the SageMaker Chainer containers repository here: https://github.com/aws/sagemaker-chainer-containers/
 
--- a/src/sagemaker/chainer/README.rst
+++ b/src/sagemaker/chainer/README.rst
@@ -176,11 +176,11 @@ The following are optional arguments. When you create a ``Chainer`` object, you 
    method launches. If not specified, the estimator generates a default
    job name, based on the training image name and current timestamp
 -  ``image_name`` An alternative docker image to use for training and
-    serving.  If specified, the estimator will use this image for training and
-    hosting, instead of selecting the appropriate SageMaker official image based on
-    framework_version and py_version. Refer to: `SageMaker Chainer Docker Containers
-    <#sagemaker-chainer-docker-containers>`_ for details on what the Official images support
-    and where to find the source code to build your custom image.
+   serving.  If specified, the estimator will use this image for training and
+   hosting, instead of selecting the appropriate SageMaker official image based on
+   framework_version and py_version. Refer to: `SageMaker Chainer Docker Containers
+   <#sagemaker-chainer-docker-containers>`_ for details on what the Official images support
+   and where to find the source code to build your custom image.
 
 
 Distributed Chainer Training
--- a/src/sagemaker/chainer/estimator.py
+++ b/src/sagemaker/chainer/estimator.py
@@ -69,8 +69,10 @@ class Chainer(Framework):
                 List of supported versions https://github.com/aws/sagemaker-python-sdk#chainer-sagemaker-estimators
             image_name (str): If specified, the estimator will use this image for training and hosting, instead of
                 selecting the appropriate SageMaker official image based on framework_version and py_version. It can
-                be an ECR url or dockerhub image and tag: 123.dkr.ecr.us-west-2.amazonaws.com/my-custom-image:1.0,
-                custom-image:latest.
+                be an ECR url or dockerhub image and tag.
+                Examples:
+                    123.dkr.ecr.us-west-2.amazonaws.com/my-custom-image:1.0
+                    custom-image:latest.
             **kwargs: Additional kwargs passed to the :class:`~sagemaker.estimator.Framework` constructor.
         """
         super(Chainer, self).__init__(entry_point, source_dir, hyperparameters,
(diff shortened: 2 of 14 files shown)


src/sagemaker/estimator.py
```diff
@@ -521,6 +526,9 @@ def __init__(self, entry_point, source_dir=None, hyperparameters=None, enable_cl
         self.container_log_level = container_log_level
         self._hyperparameters = hyperparameters or {}
         self.code_location = code_location
+        self.image_name = image_name
+        print(self.image_name)
+        print(kwargs)
```
iquintero: d'oh

src/sagemaker/estimator.py
```diff
@@ -226,6 +227,7 @@ def attach(cls, training_job_name, sagemaker_session=None):
         job_details = sagemaker_session.sagemaker_client.describe_training_job(TrainingJobName=training_job_name)
         init_params = cls._prepare_init_params_from_job_description(job_details)
 
+        print(init_params)
```
iquintero: :(

winstonaws: Remove debugging print?

src/sagemaker/estimator.py
```diff
@@ -521,6 +526,9 @@ def __init__(self, entry_point, source_dir=None, hyperparameters=None, enable_cl
         self.container_log_level = container_log_level
         self._hyperparameters = hyperparameters or {}
         self.code_location = code_location
+        self.image_name = image_name
+        print(self.image_name)
```
winstonaws: Remove debugging prints?

src/sagemaker/estimator.py
```diff
@@ -513,6 +515,9 @@ def __init__(self, entry_point, source_dir=None, hyperparameters=None, enable_cl
             code_location (str): Name of the S3 bucket where custom code is uploaded (default: None).
                 If not specified, default bucket created by ``sagemaker.session.Session`` is used.
             **kwargs: Additional kwargs passed to the ``EstimatorBase`` constructor.
+            image_name (str): An alternate image name to use instead of the official Sagemaker image
```
winstonaws: I'd go with "image" instead of "image_name" here since I think the "_name" part of it doesn't add any descriptiveness (it might even be a little confusing).

In the docstring you should explain the valid formats (e.g. ecr url, dockerhub name + tag) - include examples.

Also state that this is used for both training and deployment.

src/sagemaker/chainer/estimator.py
```diff
@@ -67,9 +67,12 @@ def __init__(self, entry_point, use_mpi=None, num_processes=None, process_slots_
                               One of 'py2' or 'py3'.
             framework_version (str): Chainer version you want to use for executing your model training code.
                 List of supported versions https://github.com/aws/sagemaker-python-sdk#chainer-sagemaker-estimators
+            image_name (str): The container image to use for training. This will override py_version and
```
winstonaws: Instead of "this will override", how about something like - "if specified, the estimator will use this image for training and hosting, instead of selecting the appropriate SageMaker official image based on framework_version and py_version.

Applies to everywhere this wording is used.

tests/unit/test_chainer.py
```diff
@@ -248,6 +248,32 @@ def test_create_model(sagemaker_session, chainer_version):
...
+    chainer = Chainer(entry_point=SCRIPT_PATH, role=ROLE, sagemaker_session=sagemaker_session,
+                      train_instance_count=INSTANCE_COUNT, train_instance_type=INSTANCE_TYPE,
+                      image_name=custom_image, container_log_level=container_log_level,
+                      py_version=PYTHON_VERSION, base_job_name='job', source_dir=source_dir,
+                      enable_cloudwatch_metrics=enable_cloudwatch_metrics)
+
+    job_name = 'new_name'
+    chainer.fit(inputs='s3://mybucket/train', job_name='new_name')
+    model = chainer.create_model()
+    chainer.container_log_level
```
winstonaws: What's this line for?

tests/unit/test_chainer.py
```diff
@@ -248,6 +248,32 @@ def test_create_model(sagemaker_session, chainer_version):
...
+
+    assert model.sagemaker_session == sagemaker_session
+    assert model.image == custom_image
```
winstonaws: I think I would keep the asserts minimal for this test, and just assert the image, assuming that other unit tests cover the other parameters already.

tests/unit/test_chainer.py
```diff
@@ -393,3 +419,46 @@ def test_attach_wrong_framework(sagemaker_session):
...
+    assert estimator.train_instance_count == 1
+    assert estimator.train_max_run == 24 * 60 * 60
+    assert estimator.input_mode == 'File'
+    assert estimator.base_job_name == 'neo'
+    assert estimator.output_path == 's3://place/output/neo'
+    assert estimator.output_kms_key == ''
+    assert estimator.hyperparameters()['training_steps'] == '100'
+    assert estimator.source_dir == 's3://some/sourcedir.tar.gz'
+    assert estimator.entry_point == 'iris-dnn-classifier.py'
+    assert estimator.train_image() == training_image
```
winstonaws: Same comment here about keeping minimal asserts (unless there's a specific reason that these interact with overriding the image.)

tests/unit/test_chainer.py
iquintero: this line was carried over from another test that I used as a template, I just realized this was present in other tests as well but its basically useless. I will get rid of it.

src/sagemaker/estimator.py
iquintero: I went with image_name because that is what the Estimator class uses, so I wanted to at least be consistent with that. I think just image is better but I dont know if we should have a different parameter name across different estimators.

winstonaws: Oh okay, let's keep image_name then.

src/sagemaker/chainer/README.rst
```diff
@@ -175,6 +175,12 @@ The following are optional arguments. When you create a ``Chainer`` object, you
 -  ``job_name`` Name to assign for the training job that the fit()
    method launches. If not specified, the estimator generates a default
    job name, based on the training image name and current timestamp
+-  ``image_name`` An alternative docker image to use for training and
```
winstonaws: Formatting looks broken here - take a look in the rich view?

src/sagemaker/chainer/estimator.py
```diff
@@ -67,8 +67,10 @@ def __init__(self, entry_point, use_mpi=None, num_processes=None, process_slots_
                               One of 'py2' or 'py3'.
             framework_version (str): Chainer version you want to use for executing your model training code.
                 List of supported versions https://github.com/aws/sagemaker-python-sdk#chainer-sagemaker-estimators
-            image_name (str): The container image to use for training. This will override py_version and
-                framework_version. The image is expected to be a modification of the SageMaker Chainer image.
+            image_name (str): If specified, the estimator will use this image for training and hosting, instead of
+                selecting the appropriate SageMaker official image based on framework_version and py_version. It can
+                be an ECR url or dockerhub image and tag: 123.dkr.ecr.us-west-2.amazonaws.com/my-custom-image:1.0,
```
winstonaws: Lets make it more explicit that those are two separate examples. You could do something like:

"It can be an ECR url or dockerhub image and tag. Examples: 123.dkr.ecr.us-west-2.amazonaws.com/my-custom-image:1.0
custom-image:latest"

Issue #291, opened 2018-07-13
[Documentation Request] Sagemaker update requires new IAM permissions


zmjjmz:
### System Information
- **Framework (e.g. TensorFlow) / Algorithm (e.g. KMeans)**: (n/a)
- **Framework Version**: (n/a)
- **Python Version**: 2.7.9
- **CPU or GPU**: (n/a)
- **Python SDK Version**: 1.5.3
- **Are you using a custom image**: (n/a)

### Describe the problem
Our ops team has fairly locked down IAM permissions, and when initially using the Sagemaker SDK I had to trial-and-error which ones I needed. In the most recent version of the Sagemaker SDK PR [268](https://github.com/aws/sagemaker-python-sdk/pull/268) introduced a new requirement that any role used by the Sagemaker SDK needed `iam:GetRole` granted, which I did not have previously granted for this role. While it wasn't hard to get this role granted, it was a breaking change for us that I would have appreciated a warning for.

Is there documentation on what IAM permissions are used by the SDK? If so, I'd just like to request that new permissions end up in the changelog / with a warning so that it's not a surprise.

### Minimal repro / logs
I don't have a good way to minimal repro this since of course you'd need our IAM setup, which is not doable :)

Here's the exact error I got though:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the GetRole operation: User: arn:aws:sts::<withheld>:assumed-role/sagemaker-user-role/
botocore-session-1531238236 is not authorized to perform: iam:GetRole on resource: role sagemaker-user-role
```

wendtek:
I am also looking for a list of permissions needed to write a minimal policy. This would be very useful.

@zmjjmz any chance you can supply what permissions your team has narrowed it down to so far? It would help give me a jump start past the initial trial and error process. Thanks in advance.

andremoeller:
Hi @zmjjmz ,

Thanks for the report! I sent a pull request to fix this behavior so that `get_execution_role` doesn't fail in this case: https://github.com/aws/sagemaker-python-sdk/pull/305

We don't have documentation about the IAM permissions needed for the SDK, but you're right that we should add clear documentation on this.

@wendtek I believe (but haven't confirmed) that the Python SDK doesn't need additional permissions beyond those your SageMaker IAM role would need (with this exception), so adding the AmazonSageMakerFullAccess policy to your IAM role or IAM user would be a good start, plus potentially some additional S3 permissions.

Thanks again!

[...]

PR #305, opened 2018-07-19
try to get_role, but fall back to role from regex

andremoeller:
*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/291

*Description of changes:*

## Merge Checklist

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -439,15 +439,24 @@ class Session(object):
         role = self.expand_role(role)
         primary_container = _expand_container_def(primary_container)
         LOGGER.info('Creating model with name: {}'.format(name))
-        LOGGER.debug("create_model request: {}".format({
+        LOGGER.debug('create_model request: {}'.format({
             'name': name,
             'role': role,
             'primary_container': primary_container
         }))
 
-        self.sagemaker_client.create_model(ModelName=name,
-                                           PrimaryContainer=primary_container,
-                                           ExecutionRoleArn=role)
+        try:
+            self.sagemaker_client.create_model(ModelName=name,
+                                               PrimaryContainer=primary_container,
+                                               ExecutionRoleArn=role)
+        except ClientError as e:
+            error_code = e.response['Error']['Code']
+            message = e.response['Error']['Message']
+
+            if error_code == 'ValidationException' and 'Cannot create already existing model' in message:
+                LOGGER.warning('Using already existing model: {}'.format(name))
+            else:
+                raise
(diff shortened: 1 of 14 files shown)


src/sagemaker/session.py
```diff
@@ -763,7 +763,10 @@ def get_caller_identity_arn(self):
 
         # Call IAM to get the role's path
         role_name = role[role.rfind('/') + 1:]
-        role = self.boto_session.client('iam').get_role(RoleName=role_name)['Role']['Arn']
+        try:
+            role = self.boto_session.client('iam').get_role(RoleName=role_name)['Role']['Arn']
+        except ClientError:
+            LOGGER.warning("Couldn't call 'get_role' to get Role ARN from role name {}.".format(role_name))
```
laurenyu: I'd include a note here that this will exclude a role's path if it has one

tests/unit/test_session.py
```diff
@@ -69,6 +69,16 @@ def test_get_caller_identity_arn_from_an_user(boto_session):
...
 
 
+def test_get_caller_identity_arn_from_an_user_without_permissions(boto_session):
+    sess = Session(boto_session)
+    arn = 'arn:aws:iam::369233609183:user/mia'
+    sess.boto_session.client('sts').get_caller_identity.return_value = {'Arn': arn}
+    sess.boto_session.client('iam').get_role.side_effect = ClientError('Bad permissions!', {})
+
+    actual = sess.get_caller_identity_arn()
+    assert actual == 'arn:aws:iam::369233609183:user/mia'
```
laurenyu: would it be possible to also test if the logger does warn?

src/sagemaker/session.py
harusametime: @andremoeller @laurenyu 
*AmazonSageMakerFullAccess* that is a managed policy for SageMaker \(see [doc](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-roles.html#sagemaker-roles-amazonsagemakerfullaccess-policy)\) does not allow `iam:GetRole`.  The line in `try`always fail and goes to `except` if users attach only *AmazonSageMakerFullAccess* to the role; `get_role()` does not work. Could you consider another implementation for this?

andremoeller: @harusametime 

You're right -- in that case. The purpose of this is to attempt to get the role using `iam:GetRole`, but never to fail in this case -- not all users will only be using `AmazonSageMakerFullAccess` policy.

PR #309, opened 2018-07-19
modify cifar test to avoid timeouts

andremoeller:
*Issue #, if available:*

*Description of changes:* This test frequently fails in FRA, seemingly due to difficulty getting p2.xlarge (takes more than 20 minutes to start in some cases), causing spurious canary failures.

## Merge Checklist

### DIFF
--- a/tests/integ/test_tf_cifar.py
+++ b/tests/integ/test_tf_cifar.py
@@ -35,7 +35,7 @@ class PickleSerializer(object):
 
 @pytest.mark.continuous_testing
 def test_cifar(sagemaker_session, tf_full_version):
-    with timeout(minutes=20):
+    with timeout(minutes=45):
         script_path = os.path.join(DATA_DIR, 'cifar_10', 'source')
 
         dataset_path = os.path.join(DATA_DIR, 'cifar_10', 'data')


tests/integ/test_tf_cifar.py
```diff
@@ -35,14 +35,14 @@ def __call__(self, data):
...
+    with timeout(minutes=25):
         script_path = os.path.join(DATA_DIR, 'cifar_10', 'source')
 
         dataset_path = os.path.join(DATA_DIR, 'cifar_10', 'data')
 
         estimator = TensorFlow(entry_point='resnet_cifar_10.py', source_dir=script_path, role='SageMakerRole',
-                               framework_version=tf_full_version, training_steps=500, evaluation_steps=5,
-                               train_instance_count=2, train_instance_type='ml.p2.xlarge',
+                               framework_version=tf_full_version, training_steps=100, evaluation_steps=5,
+                               train_instance_count=2, train_instance_type='ml.c4.xlarge',
```
laurenyu: I'm not sure there's a good solution for this, but it would be nice if we could keep using a GPU instance type in regions where it makes sense because this is our only integ test that currently uses GPU (which is why it made it continuous testing cut)

andremoeller: Yeah, I wasn't sure about this. Do you think we should use g2? The problem seems to be capacity, and I'd rather not keep increasing the timeouts. (We'd have to make it over an hour, since I think it's an hour before ICE).

laurenyu: ouch, yeah, an hour is too much. Are g2s more available?

Another possibility could be adding a parameter so that we can specify when running pytest if the particular test run should avoid trying to get p2s - haven't thought about it long enough to figure out if that would be really messy, though

yangaws: I guess let's increase the timeout temporarily and communicate with teams related to this for long term solution?

PR #353, opened 2018-08-15
Add VPC config to estimator for training job creation

icywang86rui:
CreateTraningJob api supports vpc. This change adds vpc config as an optional
argument to Estimator.

*Issue #, if available:*

*Description of changes:*

## Merge Checklist

### DIFF
--- a/tests/integ/test_tf.py
+++ b/tests/integ/test_tf.py
@@ -20,10 +20,10 @@ import pytest
 from sagemaker.tensorflow import TensorFlow
 from tests.integ import DATA_DIR, TRAINING_DEFAULT_TIMEOUT_MINUTES
 from tests.integ.timeout import timeout_and_delete_endpoint_by_name, timeout
+from tests.integ.vpc_utils import get_or_create_subnet_and_security_group
 
 DATA_PATH = os.path.join(DATA_DIR, 'iris', 'data')
-VPC_SUBNETS = ['subnet-06b8537735fac3757']
-VPC_SECURITY_GROUP_IDS = ['sg-0a1008de6e1f384c3']
+VPC_NAME = 'training-job-test'
 
 
 @pytest.mark.continuous_testing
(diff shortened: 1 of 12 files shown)


tests/integ/test_tf.py
```diff
@@ -22,6 +22,8 @@
 from tests.integ.timeout import timeout_and_delete_endpoint_by_name, timeout
 
 DATA_PATH = os.path.join(DATA_DIR, 'iris', 'data')
+VPC_SUBNETS = ['subnet-06b8537735fac3757']
```
ChoiByungWook: Is this dependent on our account?

icywang86rui: Yep. I created this VPC in PDX. The test I added the vpc_config is not used in canary so we are not running it in other regions.

ChoiByungWook: Can we create the VPC configurations and subnet during the test? Are we guaranteed to run this VPC config in the same account all the time?

I think the test will be better suited if we make these changes.

icywang86rui: We certainly can create the VPC in the test and delete it when the test finishes. I understand your concern. On the other hand I do not want our test to depend on the availability of vpc. Perhaps we can pre create vpc configs for all the regions? I am not too keen on this solution either since we might want to run the tests in different accounts in the future. Thoughts?

iquintero: Create if not exists?  This should be a good tradeoff

iquintero:
Create if not exists? I think this should be fine.
On Aug 17, 2018, 2:36 PM -0700, icywang86rui <notifications@github.com>, wrote:

@icywang86rui commented on this pull request.

________________________________

In tests/integ/test_tf.py<https://github.com/aws/sagemaker-python-sdk/pull/353#discussion_r211038405>:

> @@ -22,6 +22,8 @@
 from tests.integ.timeout import timeout_and_delete_endpoint_by_name, timeout

 DATA_PATH = os.path.join(DATA_DIR, 'iris', 'data')
+VPC_SUBNETS = ['subnet-06b8537735fac3757']


We certainly can create the VPC in the test and delete it when the test finishes. I understand your concern. On the other hand I do not want our test to depend on the availability of vpc. Perhaps we can pre create vpc configs for all the regions? I am not too keen on this solution either since we might want to run the tests in different accounts in the future. Thoughts?

PR #355, opened 2018-08-17
Increase read timeout for sagemaker runtime

ChoiByungWook:
Increase read timeout for SageMaker runtime, as the hosting endpoint has a time limit on how long an inference can take.

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -29,7 +29,6 @@ from sagemaker.user_agent import prepend_user_agent
 from sagemaker.utils import name_from_image, secondary_training_status_message, secondary_training_status_changed
 import sagemaker.logs
 
-
 logging.basicConfig()
 LOGGER = logging.getLogger('sagemaker')
 LOGGER.setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


src/sagemaker/session.py
```diff
@@ -93,7 +92,9 @@ def _initialize(self, boto_session, sagemaker_client, sagemaker_runtime_client):
         self.sagemaker_client = sagemaker_client or self.boto_session.client('sagemaker')
         prepend_user_agent(self.sagemaker_client)
 
-        self.sagemaker_runtime_client = sagemaker_runtime_client or self.boto_session.client('runtime.sagemaker')
+        config = botocore.config.Config(read_timeout=80)
+        self.sagemaker_runtime_client = sagemaker_runtime_client or self.boto_session.client('runtime.sagemaker',
+                                                                                             config=config)
```
laurenyu: it'd be better to create `config` only when needed (i.e. only when `sagemaker_runtime_client` is `None`), and since there's an awkward line break with this change now, I think I'm more inclined toward

```python
if sagemaker_runtime_client is not None:
    self.sagemaker_runtime_client = sagemaker_runtime_client
else:
    config = botocore.config.Config(read_timeout=80)
    self.sagemaker_runtime_client = self.boto_session.client('runtime.sagemaker', config=config)
```

personal preference, though, and not a strong one at that.

ChoiByungWook: Good point. Updating.

PR #367, opened 2018-08-28
Use synthetic data in keras integ test

yangaws:
*Issue #, if available:*

*Description of changes:*
Keras integ tests failed for not downloading cifar10 data. Will use synthetic data instead to fix this problem. The synthetic data is generated in same way as tf_cifar integ test.

## Merge Checklist


nadiaya:
But customers are expected to download data from s3.
Changing to use synthetic data makes tests less relevant to the real user workflow.

laurenyu:
@nadiaya we have other tests that do download data from S3. This change matches what the TF CIFAR test does, which was a change introduced to make our tests more reliable and run faster.

tests/data/cifar_10/source/keras_cnn_cifar_10.py
```diff
@@ -77,122 +78,34 @@ def keras_model_fn(hyperparameters):
...
-    record_bytes = HEIGHT * WIDTH * DEPTH + 1
-    return tf.data.FixedLengthRecordDataset(filenames, record_bytes)
+def eval_input_fn(training_dir, hyperparameters):
+    return _generate_synthetic_data(tf.estimator.ModeKeys.EVAL, batch_size=BATCH_SIZE)
 
 
-def _filenames(mode, data_dir):
-    """Returns a list of filenames based on 'mode'."""
-    data_dir = os.path.join(data_dir, 'cifar-10-batches-bin')
+def _generate_synthetic_data(mode, batch_size):
```
laurenyu: can you put this in a different file that's used by both the TF and Keras tests?

yangaws: They are actually different. For the keras model created, the label type should match feature which is float32. Since I will talk with Marcio about whether to use real cifar10 data in tests, I think this PR is just a quick fix for current failure. We probably don't need to have a separate file with this function (and with additional parameters for type).

I will update this part after talking with Marcio in the future.

PR #412, opened 2018-09-29
Enable setting VPC config when creating/deploying models

mmeidl:
*Issue #, if available:* N/A

*Description of changes:* Previous commits (#353 , #370 ) enabled users to set VpcConfig on Estimators/Training Jobs. The same VpcConfig can be set on Models in order to protect network traffic in inference use cases (see [docs](https://docs.aws.amazon.com/sagemaker/latest/dg/host-vpc.html)). 

In this PR I've added the optional parameter `vpc_config` to `Model` constructors and `Session` methods, and modified most `create_model` and `deploy` implementations to pass forward the same VPC config used in training, or take an override from the optional parameter `vpc_config`. 

To ensure the functionality works for both training and inference, I've improved the integ tests to cover VPC features for a variety of models. **Note that name and resource configurations belonging to the test VPC have changed. Before running the integ tests in your account, you should delete the existing test VPC.**

## Merge Checklist

### DIFF
--- a/src/sagemaker/amazon/factorization_machines.py
+++ b/src/sagemaker/amazon/factorization_machines.py
@@ -19,6 +19,7 @@ from sagemaker.amazon.validation import gt, isin, ge
 from sagemaker.predictor import RealTimePredictor
 from sagemaker.model import Model
 from sagemaker.session import Session
+from sagemaker.utils import vpc_config_dict
 
 
 class FactorizationMachines(AmazonAlgorithmEstimatorBase):
@@ -163,11 +164,19 @@ class FactorizationMachines(AmazonAlgorithmEstimatorBase):
         self.factors_init_sigma = factors_init_sigma
         self.factors_init_value = factors_init_value
 
-    def create_model(self):
+    def create_model(self, vpc_config=None):
         """Return a :class:`~sagemaker.amazon.FactorizationMachinesModel` referencing the latest
-        s3 model data produced by this Estimator."""
+        s3 model data produced by this Estimator.
 
-        return FactorizationMachinesModel(self.model_data, self.role, sagemaker_session=self.sagemaker_session)
+        Args:
+            vpc_config (dict[str, list[str]]): Overrides VpcConfig set on the model.
+                If None, defaults to VpcConfig used for training (default: None)
+                * 'Subnets' (list[str]): List of subnet ids.
+                * 'SecurityGroupIds' (list[str]): List of security group ids.
+        """
+        vpc_config = vpc_config or vpc_config_dict(self.subnets, self.security_group_ids)
+        return FactorizationMachinesModel(self.model_data, self.role, sagemaker_session=self.sagemaker_session,
+                                          vpc_config=vpc_config)
 
 
 class FactorizationMachinesPredictor(RealTimePredictor):
@@ -195,7 +204,7 @@ class FactorizationMachinesModel(Model):
     """Reference S3 model data created by FactorizationMachines estimator. Calling :meth:`~sagemaker.model.Model.deploy`
     creates an Endpoint and returns :class:`FactorizationMachinesPredictor`."""
 
-    def __init__(self, model_data, role, sagemaker_session=None):
+    def __init__(self, model_data, role, sagemaker_session=None, **kwargs):
         sagemaker_session = sagemaker_session or Session()
         repo = '{}:{}'.format(FactorizationMachines.repo_name, FactorizationMachines.repo_version)
         image = '{}/{}'.format(registry(sagemaker_session.boto_session.region_name), repo)
@@ -203,4 +212,5 @@ class FactorizationMachinesModel(Model):
                                                          image,
                                                          role,
                                                          predictor_cls=FactorizationMachinesPredictor,
-                                                         sagemaker_session=sagemaker_session)
+                                                         sagemaker_session=sagemaker_session,
+                                                         **kwargs)
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -446,30 +446,30 @@ class Session(object):
             primary_container (str or dict[str, str]): Docker image which defines the inference code.
                 You can also specify the return value of ``sagemaker.container_def()``, which is used to create
                 more advanced container configurations, including model containers which need artifacts from S3.
-            vpc_config (dict[str, list[str]]): Contains values for VpcConfig:
-                * subnets (list[str]): List of subnet ids.
-                    The key in vpc_config is 'Subnets'.
-                * security_group_ids (list[str]): List of security group ids.
-                    The key in vpc_config is 'SecurityGroupIds'.
+            vpc_config (dict[str, list[str]]): The VpcConfig set on the model (default: None)
+                * 'Subnets' (list[str]): List of subnet ids.
+                * 'SecurityGroupIds' (list[str]): List of security group ids.
 
         Returns:
             str: Name of the Amazon SageMaker ``Model`` created.
         """
         role = self.expand_role(role)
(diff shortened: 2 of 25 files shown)


src/sagemaker/session.py
```diff
@@ -489,6 +496,11 @@ def create_model_from_job(self, training_job_name, name=None, role=None, primary
             model_data_url (str): S3 location of the model data (default: None). If None, defaults to
                 the ``ModelS3Artifacts`` of ``training_job_name``.
             env (dict[string,string]): Model environment variables (default: {}).
+            vpc_config (dict[str, list[str]]):
```
icywang86rui: Delete

src/sagemaker/session.py
```diff
@@ -671,6 +684,10 @@ def endpoint_from_job(self, job_name, initial_instance_count, instance_type,
             wait (bool): Whether to wait for the endpoint deployment to complete before returning (default: True).
             model_environment_vars (dict[str, str]): Environment variables to set on the model container
                 (default: None).
+            model_vpc_config (dict[str, list[str]]): Overrides VpcConfig set on the model.
```
icywang86rui: Tom mentioned he wants us to use a marker like USE_MODEL_VPC. Are we still doing that?

src/sagemaker/session.py
```diff
@@ -721,7 +743,8 @@ def endpoint_from_model_data(self, model_s3_location, deployment_image, initial_
                               role=role,
                               primary_container=container_def(image=deployment_image,
                                                               model_data_url=model_s3_location,
-                                                              env=model_environment_vars))
+                                                              env=model_environment_vars),
+                              vpc_config=model_vpc_config)
```
icywang86rui: indentation

tests/integ/vpc_utils.py
```diff
@@ -47,25 +49,43 @@ def _get_route_table_id(ec2_client, vpc_id):
...
 
 
-def create_vpc_with_name(ec2_client, name):
+def create_vpc_with_name(ec2_client, region, name):
     vpc_id = ec2_client.create_vpc(CidrBlock='10.0.0.0/16')['Vpc']['VpcId']
+    print('created vpc: {}'.format(vpc_id))
 
-    subnet_id = ec2_client.create_subnet(CidrBlock='10.0.0.0/24', VpcId=vpc_id)['Subnet']['SubnetId']
+    # sagemaker endpoints require subnets in at least 2 different AZs for vpc mode
+    subnet_id_a = ec2_client.create_subnet(CidrBlock='10.0.0.0/24', VpcId=vpc_id,
```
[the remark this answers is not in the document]
mmeidl: I think I'll keep this under one `create_vpc_with_name` method. Most of the `create_subnet` parameters are coupled to the vpc's ID and CIDR range anyway.

src/sagemaker/utils.py
```diff
@@ -88,6 +88,13 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def vpc_config_dict(subnets, security_group_ids):
```
mvsusp: This function should not be in utils. Maybe in session.py where container_def is defined? https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/session.py#L917

tests/integ/test_vpc.py
```diff
@@ -0,0 +1,364 @@
...
+from sagemaker.pytorch import PyTorch
+from sagemaker.tensorflow import TensorFlow
+from tests.integ import DATA_DIR, TRAINING_DEFAULT_TIMEOUT_MINUTES, PYTHON_VERSION
+from tests.integ.timeout import timeout_and_delete_endpoint_by_name, timeout
+from tests.integ.vpc_utils import get_or_create_subnets_and_security_group
+
+DATA_PATH = os.path.join(DATA_DIR, 'iris', 'data')
+
+
+@pytest.mark.skipif(PYTHON_VERSION != 'py2', reason="TensorFlow image supports only python 2.")
```
mvsusp: nit: use not instead of !=

src/sagemaker/utils.py
mvsusp: Please, add docstrings.

tests/integ/test_vpc.py
```diff
@@ -0,0 +1,364 @@
...
+        assert len(result['predictions']) == 10
+        for prediction in result['predictions']:
+            assert prediction['score'] is not None
+
+
+def fm_serializer(data):
+    js = {'instances': []}
+    for row in data:
+        js['instances'].append({'features': row.tolist()})
+    return json.dumps(js)
```
mvsusp: I appreciate the effort to write vpc tests for each framework. One concern is that each of these integration tests can take even 45 minutes, depending of the instance type. Given that all the frameworks share the same Estimator, is it possible to add vpc in some of the integration tests previously created for the frameworks to avoid repetitive tests?

Arguably vpc should be part of other integrations tests and not integration tests itself.

src/sagemaker/utils.py
mmeidl: The function is imported and used in all Estimators too. It's how they translate the `subnets` and `security_group_ids` attributes into a `vpc_config` passed to the Model or Session method. For this reason I put it in utils so it can be imported from a central module.

Would you rather see 
```
from sagemaker.session import Session, vpc_config_dict
```
or
```
from sagemaker.session import Session
from sagemaker.utils import vpc_config_dict
```

either way, yes I will add docstring

src/sagemaker/session.py
mmeidl: Tom clarified that he was asking for a global constant like this which would be used as the default value. It makes it more clear for a user to reuse training's `vpc_config`, or else override to some different value, or even `None`. I will add this.

src/sagemaker/utils.py
mmeidl: I also considered moving to a totally new module `vpc_utils` which would include the following:

```
VPC_CONFIG_DEFAULT = {}


def vpc_config_dict(subnets, security_group_ids):
    ...


def vpc_config_unpack(vpc_config):
    ...
```

Any guiding principles for when to split off a new utility module?

src/sagemaker/amazon/factorization_machines.py
```diff
@@ -163,11 +164,19 @@ def __init__(self, role, train_instance_count, train_instance_type,
         self.factors_init_sigma = factors_init_sigma
         self.factors_init_value = factors_init_value
 
-    def create_model(self):
+    def create_model(self, vpc_config=None):
```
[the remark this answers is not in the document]
mvsusp: That might be fine. This project follows the Python Google Style standards https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods. Please take them in consideration.

src/sagemaker/utils.py
mvsusp: Based on https://github.com/google/styleguide/blob/gh-pages/pyguide.md#224-decision. I would prefer 

```
from sagemaker import session

session.Session ...
session.vpc_config_dict
```

I am ok if you decide to move the vpc to a separate file thought.

tests/integ/test_vpc.py
mvsusp: I suggest to check how longer will be the test run after your changes.
If you have the same functionality replicated in multiple classes that might be a sign that some refactoring could be done.

mmeidl: After more thinking about this (and witnessing first-hand that the VPC tests took a long time to run), I've decided to consolidate. VPC features will be covered by a single cases in the TensorFlow integ tests. Other functionality like changing/dropping VPC config will be added in unit tests.

Issue #434, opened 2018-10-18
Support PEP-396: sagemaker.__version__ doesn't work


leopd:
Please fill out the form below.

### System Information
- **Framework (e.g. TensorFlow) / Algorithm (e.g. KMeans)**: all
- **Framework Version**: n/a
- **Python Version**: 3.6.3
- **CPU or GPU**: CPU
- **Python SDK Version**:  well that's the funny thing...  (pip freeze says 1.11.3)
- **Are you using a custom image**:  n/a

### Describe the problem
I'd like to be able to know which version of the sagemaker python SDK I'm using, so I naturally try what's recommended by PEP-396, and it doesn't work.

### Minimal repro / logs
```
>>> sagemaker.__version__
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-6450e5659cf0> in <module>()
----> 1 sagemaker.__version__

AttributeError: module 'sagemaker' has no attribute '__version__'
```

- **Exact command to reproduce**:

`sagemaker.__version__`

jesterhazy:
fixed by https://github.com/aws/sagemaker-python-sdk/pull/435

jesterhazy:
this will ship in our next release

PR #435, opened 2018-10-19
add PEP-396 version attribute to sagemaker module

jesterhazy:
*Issue #, if available:*

https://github.com/aws/sagemaker-python-sdk/issues/434

*Description of changes:*

add PEP-396 version attribute to sagemaker module

## Merge Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,10 +12,19 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 
-from glob import glob
 import os
+import re
+from glob import glob
 
 from setuptools import setup, find_packages
+
+
+def get_version():
+    root = os.path.dirname(__file__)
+    version_re = re.compile(r'''__version__ = ['"]([0-9.]+)['"]''')
+
+    init = read(os.path.join(root, 'src/sagemaker', '__init__.py'))
+    return version_re.search(init).group(1)
 
 
 def read(fname):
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -12,18 +12,27 @@
...
 
-from glob import glob
 import os
+import re
+from glob import glob
 
 from setuptools import setup, find_packages
 
 
+def get_version():
```
nadiaya: I believe we should follow pep-0396 guidance https://www.python.org/dev/peps/pep-0396/#classic-distutils :
In classic distutils, the simplest way to add the version string to the setup() function in setup.py is to do something like this:
```
from elle import __version__
setup(name='elle', version=__version__)
```
In the PEP author's experience however, this can fail in some cases, such as when the module uses automatic Python 3 conversion via the 2to3 program (because setup.py is executed by Python 3 before the elle module has been converted).

In that case, it's not much more difficult to write a little code to parse the __version__ from the file rather than importing it. Without providing too much detail, it's likely that modules such as distutils2 will provide a way to parse version strings from files. E.g.:
```
from distutils2 import get_version
setup(name='elle', version=get_version('elle.py'))
```

nadiaya: Nevermind, from offline conversation that seems to be the new way (especially since Distutils2 development is stopped):
https://packaging.python.org/guides/single-sourcing-package-version/

PR #528, opened 2018-12-06
Add AugmentedManifestFile & ShuffleConfig support

ishaaq:
*Description of changes:*
Add changes to support for [AugmentedManifestFile](https://docs.aws.amazon.com/sagemaker/latest/dg/API_S3DataSource.html#SageMaker-Type-S3DataSource-S3DataType) and [ShuffleConfig](https://docs.aws.amazon.com/sagemaker/latest/dg/API_Channel.html#SageMaker-Type-Channel-ShuffleConfig).

## Merge Checklist


owen-t:
I'm going to add Lauren Yu to this as well

src/sagemaker/session.py
```diff
@@ -1264,6 +1270,24 @@ def __init__(self, s3_data, distribution='FullyReplicated', compression=None,
             self.config['RecordWrapperType'] = record_wrapping
         if input_mode is not None:
             self.config['InputMode'] = input_mode
+        if attribute_names is not None:
+            self.config['DataSource']['S3DataSource']['AttributeNames'] = attribute_names
+        if shuffle_config is not None:
+            self.config['ShuffleConfig'] = {'Seed': shuffle_config.seed}
+
+
+class ShuffleConfig(object):
```
owen-t: The s3 input class uses snake case. Should we do the same here? I'll defer to Lauren on style issues as I'm not sure if there is a style guide they are using.

owen-t:
Using a class for shuffle config is a great idea and keeping with the pattern used in s3 input. I don't think this needs an integration test as it's pretty straightforward API parameter passing.

src/sagemaker/session.py
ishaaq: Well, I had a choice of either taking a dict or a class. A dict would work though it forces the user to go look at SageMaker docs to understand that the dict requires a single key called "Seed" with a long value. We also cannot take just the seed as input since SageMaker Training's ShuffleConfig could eventually become a richer object.

owen-t:
+1

PR #553, opened 2018-12-13
Fix __all__ in sagemaker.rl

tyrion:
`__all__` should be a list of names (str), not objects.
Moreover it

> should be placed after the module docstring but before any import statements except from `__future__` imports (PEP8)

I confirm that my contribution is made under the terms of the Apache 2.0 license.


laurenyu:
@tyrion sorry for the delay in response. If you're up for fixing this in the rest of the estimators, feel free to do so in this PR. Otherwise, if you open an issue about it, one of us can probably take care of it

PR #590, opened 2019-01-08
Specify task of training/tuning job in Airflow transform/deploy operator

yangaws:
*Issue #, if available:*

*Description of changes:*

Specify task id and type of the training/tuning job in Airflow transform/endpoint operator. Then the API knows that which training job to transform/deploy if there are multiple jobs created in the DAG. To make this happen, additional changes are:
1. Make default name of jobs (training/tuning/transform/model/endpoint) random instead of global(all same). This also enables retry feature.
2. Remove the Airflow naming generation part in utils. Use the timestamp name postfix as before.

## Merge Checklist


CHANGELOG.rst
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 1.17.0.dev
 ==========
 
+* enhancement: Specify tasks from which training/tuning operator to transform/deploy in related operators
```
laurenyu: nit: and something to denote this is for Airflow/workflows

src/sagemaker/workflow/airflow.py
```diff
@@ -29,7 +29,7 @@ def prepare_framework(estimator, s3_operations):
     """
     if estimator.code_location is not None:
         bucket, key = fw_utils.parse_s3_url(estimator.code_location)
-        key = os.path.join(key, 'source', 'sourcedir.tar.gz')
+        key = os.path.join(key, estimator._current_job_name, 'source', 'sourcedir.tar.gz')
```
laurenyu: iirc we have a potential bug here where `_current_job_name` isn't set if an estimator is created using `attach`

src/sagemaker/workflow/airflow.py
```diff
@@ -279,6 +279,57 @@ def tuning_config(tuner, inputs, job_name=None):
...
+        estimator (sagemaker.estimator.Framework): The Framework estimator to update.
+        job_name (str): The new job name included in the submit S3 URI
+
+    Returns:
+        str: The updated S3 URI of framework source directory
+    """
+    if estimator.uploaded_code is None:
+        return
+
+    postfix = os.path.join('/', 'source', 'sourcedir.tar.gz')
```
laurenyu: if we're actually concerned about using the correct separator for the OS, then you should use `os.path.sep` instead of `'/'`. however, if this is something where we actually just need slashes, then you can just use a format string here.

src/sagemaker/workflow/airflow.py
```diff
@@ -279,6 +279,57 @@ def tuning_config(tuner, inputs, job_name=None):
...
+
+    # update the S3 URI with the latest training job.
+    # s3://path/old_job/source/sourcedir.tar.gz will become s3://path/new_job/source/sourcedir.tar.gz
+    submit_uri = estimator.uploaded_code.s3_prefix
+    submit_uri = submit_uri[:len(submit_uri) - len(postfix)]
+    submit_uri = submit_uri[:submit_uri.rfind('/') + 1] + job_name + postfix
```
laurenyu: would this be potentially cleaner with a regex?

src/sagemaker/workflow/airflow.py
```diff
@@ -279,6 +279,57 @@ def tuning_config(tuner, inputs, job_name=None):
...
+    if task_id is None:
+        return
+    if task_type.lower() == 'training':
+        training_job = "{{ ti.xcom_pull(task_ids='%s')['Training']['TrainingJobName'] }}" % task_id
+        job_name = training_job
+    elif task_type.lower() == 'tuning':
+        training_job = "{{ ti.xcom_pull(task_ids='%s')['Tuning']['BestTrainingJob']['TrainingJobName'] }}" % task_id
+        # need to strip the double quotes in json to get the string
+        job_name = "{{ ti.xcom_pull(task_ids='%s')['Tuning']['TrainingJobDefinition']['StaticHyperParameters']" \
+                   "['sagemaker_job_name'].replace('%s', '') }}" % (task_id, '"')
```
laurenyu: would it make more sense to use `strip` instead of `replace`?

src/sagemaker/workflow/airflow.py
```diff
@@ -369,16 +421,22 @@ def model_config(instance_type, model, role=None, image=None):
     return config
 
 
-def model_config_from_estimator(instance_type, estimator, role=None, image=None, model_server_workers=None,
-                                vpc_config_override=vpc_utils.VPC_CONFIG_DEFAULT):
+def model_config_from_estimator(instance_type, estimator, task_id, task_type, role=None, image=None, name=None,
```
laurenyu: is this a breaking change?

src/sagemaker/workflow/airflow.py
yangaws: After more thoughts, I think it's more than this PR. We can have a new design and PR for attach in this case.

PR #600, opened 2019-01-15
Prepend SageMaker Notebook Instance version ID to boto3 User Agent

mckev-amazon:
SageMaker Notebook Instances provide the version ID as a file on a
Notebook Instance.  In order to enhance version adoption from within
SageMaker, provide an extra segment to the user_agent supplied to the
boto3 ClientConfig.

*Issue #, if available:*

*Description of changes:*

## Merge Checklist

### DIFF
--- a/src/sagemaker/user_agent.py
+++ b/src/sagemaker/user_agent.py
@@ -25,9 +25,23 @@ OS_VERSION = platform.release() or 'UnresolvedOSVersion'
 PYTHON_VERSION = '{}.{}.{}'.format(sys.version_info.major, sys.version_info.minor, sys.version_info.micro)
 
 
-def prepend_user_agent(client):
+def determine_prefix():
     prefix = 'AWS-SageMaker-Python-SDK/{} Python/{} {}/{} Boto3/{} Botocore/{}'\
         .format(SDK_VERSION, PYTHON_VERSION, OS_NAME, OS_VERSION, boto3.__version__, botocore.__version__)
+
+    try:
+        with open('/etc/opt/ml/sagemaker-notebook-instance-version.txt') as sagemaker_nbi_file:
+            prefix = 'AWS-SageMaker-Notebook-Instance/{} {}'.format(sagemaker_nbi_file.read().strip(), prefix)
+    except IOError:
+        # This file isn't expected to always exist, and we DO want to silently ignore failures.
+        pass
+
+    return prefix
+
+
+def prepend_user_agent(client):
+    prefix = determine_prefix()
+
     if client._client_config.user_agent is None:
         client._client_config.user_agent = prefix
     else:
--- a/tests/unit/test_session.py
+++ b/tests/unit/test_session.py
@@ -19,7 +19,7 @@ import logging
 import pytest
 import six
 from botocore.exceptions import ClientError
-from mock import Mock, patch, call
+from mock import MagicMock, Mock, patch, call, mock_open
 
 import sagemaker
 from sagemaker import s3_input, Session, get_execution_role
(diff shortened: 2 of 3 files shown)


src/sagemaker/user_agent.py
```diff
@@ -25,9 +25,23 @@
 PYTHON_VERSION = '{}.{}.{}'.format(sys.version_info.major, sys.version_info.minor, sys.version_info.micro)
 
 
-def prepend_user_agent(client):
+def determine_prefix():
     prefix = 'AWS-SageMaker-Python-SDK/{} Python/{} {}/{} Boto3/{} Botocore/{}'\
         .format(SDK_VERSION, PYTHON_VERSION, OS_NAME, OS_VERSION, boto3.__version__, botocore.__version__)
+
+    try:
+        with open('/etc/opt/ml/sagemaker-notebook-instance-version.txt') as sagemaker_nbi_file:
```
nadiaya: Add tests.

src/sagemaker/user_agent.py
```diff
@@ -25,9 +25,23 @@
 PYTHON_VERSION = '{}.{}.{}'.format(sys.version_info.major, sys.version_info.minor, sys.version_info.micro)
 
 
-def prepend_user_agent(client):
+def determine_prefix():
     prefix = 'AWS-SageMaker-Python-SDK/{} Python/{} {}/{} Boto3/{} Botocore/{}'\
         .format(SDK_VERSION, PYTHON_VERSION, OS_NAME, OS_VERSION, boto3.__version__, botocore.__version__)
+
+    try:
```
laurenyu: would it be better to use `os.path.exists` rather than a try block?

src/sagemaker/user_agent.py
```diff
@@ -25,9 +25,23 @@
...
+        with open('/etc/opt/ml/sagemaker-notebook-instance-version.txt') as sagemaker_nbi_file:
+            prefix = 'AWS-SageMaker-Notebook-Instance/{} {}'.format(sagemaker_nbi_file.read().strip(), prefix)
+    except IOError:
```
nadiaya: Should we make the error even broader to minimize any possibility of it affecting customers?

mckev-amazon: I don't think so, if something besides an IOError is thrown it should fail pretty hard, IMO.

src/sagemaker/user_agent.py
mckev-amazon: No, to do so wouldn't be pythonic and also exposes a race condition (the file could exist in line 1 but not exist or isn't accessible by the time line 2 is run).  Doing so via a try...except block is idomatic in Python.  This is called "EAFP" - see a good example here:

https://docs.python.org/3.4/library/os.html?highlight=eafp#os.access

src/sagemaker/user_agent.py
mckev-amazon: Yep, I can do that.  Do note that prior to this there weren't any tests on the user_agent.py class, I'll do what I can.

tests/unit/test_session.py
```diff
@@ -139,6 +145,57 @@ def test_delete_endpoint(boto_session):
...
+    assert 'AWS-SageMaker-Python-SDK' in sess.sagemaker_runtime_client._client_config.user_agent
+    assert 'AWS-SageMaker-Notebook-Instance' not in sess.sagemaker_client._client_config.user_agent
+    assert 'AWS-SageMaker-Notebook-Instance' not in sess.sagemaker_runtime_client._client_config.user_agent
+
+
+def patch_open(*args, **kwargs):
+    """Depending on the current Python version, patch the appropriate built-in open() function."""
+
+    try:
+        # Python 3.x
```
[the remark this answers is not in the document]
mckev-amazon: TIL, thanks for pointing this out.  Will make the change.

Issue #447, opened 2018-10-27
Estimator.delete_endpoint does not delete the Model or Endpoint Configuration


mvsusp:
Please fill out the form below.

### System Information
- **Framework (e.g. TensorFlow) / Algorithm (e.g. KMeans)**:
Any Framework Container
- **Python Version**:
2 and 3
- **Python SDK Version**:
1.12

### Describe the problem
Issue found by @andrewcking here https://github.com/aws/sagemaker-python-sdk/issues/402

In SageMaker Hosting, the process to create an endpoint requires to [create an model, an endpoint configuration, and an endpoint](https://docs.aws.amazon.com/sagemaker/latest/dg/how-it-works-hosting.html).

SageMaker Python SDK abstract these 3 layers for you. These are the [lines where this process happens](https://github.com/aws/sagemaker-python-sdk/blob/dd44d25247927abe409f7bf9e5c4561f7e55585e/src/sagemaker/model.py#L105)

```python
        container_def = self.prepare_container_def(instance_type)
        self.name = self.name or name_from_image(container_def['Image'])
        self.sagemaker_session.create_model(self.name, self.role, container_def, vpc_config=self.vpc_config)
        production_variant = sagemaker.production_variant(self.name, instance_type, initial_instance_count)
        self.endpoint_name = endpoint_name or self.name
        self.sagemaker_session.endpoint_from_production_variants(self.endpoint_name, [production_variant], tags)
```

To be able to use Python SDK to create an endpoint with the same name, you will have to delete the model, the endpoint_config, and the endpoint. SageMaker Python SDK [delete_endpoint()](https://github.com/aws/sagemaker-python-sdk/blob/dd44d25247927abe409f7bf9e5c4561f7e55585e/src/sagemaker/estimator.py#L330) only deletes the endpoint, which is why the issue is happening.

### Minimal repro / logs
Please provide any logs and a bare minimum reproducible test case, as this will be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached.

- **Exact command to reproduce**:
```python
from sagemaker.pytorch import PyTorch

# create estimator
estimator = PyTorch(entry_point='train.py', 
                    role='SageMakerRole', 
                    framework_version='0.4.0', 
                    train_instance_count=1, 
                    train_instance_type='ml.c5.xlarge', 
                    source_dir='source', 
                    hyperparameters={'epochs': 6,})

# fit estimator
[...]

PR #630, opened 2019-02-07
Add new APIs to clean up resources from predictor and transformer.

ChuyangDeng:
… transformer to delete model.

*Issue #, if available:*
#447 

*Description of changes:*
- Add support to delete endpoint configuration by default within delete_endpoint() to predictor class.
- Add delete_model() to transformer class. Instead of linking a model class to it, the transformer class directly calls session's delete_model() method because it already has access to the model_name. We don't want to pass in a model object on top of that. We don't want to construct a model object here either because that requires extra parameters, such as the s3 path to the model artifacts, while we can just utilize the model_name variable.

## Merge Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
+* feature: Support for predictor to delete endpoint and endpoint configuration
+* feature: Support for transformer to delete model
 
 1.18.2
 ======
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
-* feature: Support for predictor to delete endpoint, and delete endpoint configuration with ``delete_endpoint()`` by default
+* feature: Make predictor to delete endpoint configuration by default when calling ``delete_endpoint()``
 * feature: Support for model class to delete SageMaker model
 * feature: Support for transformer to delete Sagemaker model
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,9 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
-* feature: Support for predictor to delete endpoint and endpoint configuration
-* feature: Support for transformer to delete model
+* feature: Support for predictor to delete endpoint, and delete endpoint configuration with ``delete_endpoint()`` by default
+* feature: Support for model class to delete SageMaker model
+* feature: Support for transformer to delete Sagemaker model
 
 1.18.2
 ======
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
-* feature: Make predictor to delete endpoint configuration by default when calling ``delete_endpoint()``
+* feature: Support for predictor class to delete endpoint configuration by default when calling ``delete_endpoint()``
 * feature: Support for model class to delete SageMaker model
 * feature: Support for transformer to delete Sagemaker model
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,9 +8,9 @@ CHANGELOG
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
 * doc-fix: add ``transform_fn`` information and fix ``input_fn`` signature in the MXNet README
-* feature: Support for predictor class to delete endpoint configuration by default when calling ``delete_endpoint()``
-* feature: Support for model class to delete SageMaker model
-* feature: Support for transformer to delete Sagemaker model
+* feature: Support for ``Predictor`` to delete endpoint configuration by default when calling ``delete_endpoint()``
+* feature: Support for ``model`` to delete SageMaker model
+* feature: Support for ``Transformer`` to delete SageMaker model
 
 1.18.2
 ======
--- a/README.rst
+++ b/README.rst
@@ -189,13 +189,14 @@ Here is an end to end example of how to use a SageMaker Estimator:
     # Serializes data and makes a prediction request to the SageMaker endpoint
     response = mxnet_predictor.predict(data)
 
-    # Tears down the SageMaker endpoint
+    # Tears down the SageMaker endpoint and endpoint configuration
     mxnet_predictor.delete_endpoint()
 
 
-The above example will delete endpoint and endpoint configuration at the same time. If you want to keep endpoint configuration, you can do the following:
+The example above will eventually delete both the SageMaker endpoint and endpoint configuration through `delete_endpoint()`. If you want to keep your SageMaker endpoint configuration, use the value False for the `delete_endpoint_config` parameter, as shown below.
 
 .. code:: python
+    # Only delete the endpoint and keep the endpoint endpoint configuration
     mxnet_predictor.delete_endpoint(delete_endpoint_config=False)
 
 Additionally, it is possible to deploy a different endpoint configuration, which links to your model, to an already existing SageMaker endpoint.
@@ -225,7 +226,7 @@ For more `information <https://boto3.amazonaws.com/v1/documentation/api/latest/r
     # Serializes data and makes a prediction request to the SageMaker endpoint
     response = mxnet_predictor.predict(data)
 
-    # Tears down the SageMaker endpoint
+    # Tears down the SageMaker endpoint and endpoint configuration
     mxnet_predictor.delete_endpoint()
 
 Training Metrics
@@ -279,7 +280,7 @@ We can take the example in  `Using Estimators <#using-estimators>`__ , and use e
     # Serializes data and makes a prediction request to the local endpoint
     response = mxnet_predictor.predict(data)
 
-    # Tears down the endpoint container
+    # Tears down the endpoint and endpoint configuration
     mxnet_predictor.delete_endpoint()
 
 
@@ -302,7 +303,7 @@ Here is an end-to-end example:
     data = numpy.zeros(shape=(1, 1, 28, 28))
     predictor.predict(data)
 
-    # Tear down the endpoint container
+    # Tear down the endpoint and endpoint configuration
     predictor.delete_endpoint()
 
 
@@ -326,6 +327,8 @@ Here is an end-to-end example:
     transformer = mxnet_estimator.transformer(1, 'local', assemble_with='Line', max_payload=1)
     transformer.transform('s3://my/transform/data, content_type='text/csv', split_type='Line')
     transformer.wait()
+
+    # Deletes the SageMaker model
     transformer.delete_model()
 
 
--- a/README.rst
+++ b/README.rst
@@ -190,8 +190,13 @@ Here is an end to end example of how to use a SageMaker Estimator:
     response = mxnet_predictor.predict(data)
 
     # Tears down the SageMaker endpoint
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
 
+
+The above example will delete endpoint and endpoint configuration at the same time. If you want to keep endpoint configuration, you can do the following:
+
+.. code:: python
+    mxnet_predictor.delete_endpoint(delete_endpoint_config=False)
 
 Additionally, it is possible to deploy a different endpoint configuration, which links to your model, to an already existing SageMaker endpoint.
 This can be done by specifying the existing endpoint name for the ``endpoint_name`` parameter along with the ``update_endpoint`` parameter as ``True`` within your ``deploy()`` call.
@@ -221,7 +226,7 @@ For more `information <https://boto3.amazonaws.com/v1/documentation/api/latest/r
     response = mxnet_predictor.predict(data)
 
     # Tears down the SageMaker endpoint
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
 
 Training Metrics
 ~~~~~~~~~~~~~~~~
@@ -275,7 +280,7 @@ We can take the example in  `Using Estimators <#using-estimators>`__ , and use e
     response = mxnet_predictor.predict(data)
 
     # Tears down the endpoint container
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
 
 
 If you have an existing model and want to deploy it locally, don't specify a sagemaker_session argument to the ``MXNetModel`` constructor.
@@ -321,6 +326,7 @@ Here is an end-to-end example:
     transformer = mxnet_estimator.transformer(1, 'local', assemble_with='Line', max_payload=1)
     transformer.transform('s3://my/transform/data, content_type='text/csv', split_type='Line')
     transformer.wait()
+    transformer.delete_model()
 
 
 For detailed examples of running Docker in local mode, see:
--- a/README.rst
+++ b/README.rst
@@ -196,7 +196,7 @@ Here is an end to end example of how to use a SageMaker Estimator:
 The example above will eventually delete both the SageMaker endpoint and endpoint configuration through `delete_endpoint()`. If you want to keep your SageMaker endpoint configuration, use the value False for the `delete_endpoint_config` parameter, as shown below.
 
 .. code:: python
-    # Only delete the endpoint and keep the endpoint endpoint configuration
+    # Only delete the SageMaker endpoint, while keeping the corresponding endpoint configuration.
     mxnet_predictor.delete_endpoint(delete_endpoint_config=False)
 
 Additionally, it is possible to deploy a different endpoint configuration, which links to your model, to an already existing SageMaker endpoint.
--- a/README.rst
+++ b/README.rst
@@ -280,7 +280,7 @@ We can take the example in  `Using Estimators <#using-estimators>`__ , and use e
     # Serializes data and makes a prediction request to the local endpoint
     response = mxnet_predictor.predict(data)
 
-    # Tears down the endpoint and endpoint configuration
+    # Tears down the endpoint container and deletes the corresponding endpoint configuration
     mxnet_predictor.delete_endpoint()
 
 
@@ -303,7 +303,7 @@ Here is an end-to-end example:
     data = numpy.zeros(shape=(1, 1, 28, 28))
     predictor.predict(data)
 
-    # Tear down the endpoint and endpoint configuration
+    # Tear down the endpoint container and delete the corresponding endpoint configuration
     predictor.delete_endpoint()
 
 
--- a/src/sagemaker/model.py
+++ b/src/sagemaker/model.py
@@ -302,6 +302,16 @@ class Model(object):
                            env=env, tags=tags, base_transform_job_name=self.name,
                            volume_kms_key=volume_kms_key, sagemaker_session=self.sagemaker_session)
 
+    def delete_model(self):
+        """Delete an Amazon SageMaker ``Model``.
+
+        Raises: ValueError if model is not deployed yet.
+
+        """
+        if self.name is None:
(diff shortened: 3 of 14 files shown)


ChuyangDeng:
I only included changes regarding deleting endpoint/endpoint config in this PR. model class linkage changes will be sent in a separate one because many files will be changed to link model and I don't want this PR to be too long.

src/sagemaker/model.py
```diff
@@ -302,6 +302,16 @@ def transformer(self, instance_count, instance_type, strategy=None, assemble_wit
...
                            volume_kms_key=volume_kms_key, sagemaker_session=self.sagemaker_session)
 
+    def delete_model(self):
+        """Delete an Amazon SageMaker ``Model``.
+
+        Raises: ValueError if model is not deployed yet.
+
+        """
+        if self.name is None:
+            raise ValueError('The SageMaker model is not deployed yet.')
```
ChoiByungWook: For the exception, I would specify something along the lines of...

"The SageMaker model must be deployed first before attempting to delete."

Although, the small problem here is that technically my SageMaker model can exist before I actually create an endpoint.

ChoiByungWook: I think this None check can also go in the session class. That way you don't have to check in the Transformer class either.

The unit test for this is missing I believe.

src/sagemaker/predictor.py
```diff
@@ -105,11 +105,27 @@ def _create_request_args(self, data, initial_args=None):
...
+
+        Args:
+            delete_endpoint_config (bool): Flag to indicate whether to delete endpoint configuration together with
+                endpoint. If False, only endpoint will be deleted. Default: True.
+
         """
         self.sagemaker_session.delete_endpoint(self.endpoint)
 
+        if delete_endpoint_config:
+            self._delete_endpoint_config()
```
ChoiByungWook: I think I'm a little worried about the endpoint configuration happening after the delete_endpoint, as we call ``describe_endpoint`` within the newly introduced ``_delete_endpoint_config`` to get metadata on the endpoint to determine the corresponding endpoint configuration.

I think it might be possible for the endpoint to be deleted and lose the trace to the AWS account. This might end with not being able to get the metadata needed to properly delete the endpoint config.

README.rst
```diff
@@ -190,9 +190,14 @@ Here is an end to end example of how to use a SageMaker Estimator:
     response = mxnet_predictor.predict(data)
 
     # Tears down the SageMaker endpoint
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
```
ChoiByungWook: awesome!

tests/integ/test_transformer.py
```diff
@@ -104,6 +104,46 @@ def test_attach_transform_kmeans(sagemaker_session):
         attached_transformer.wait()
 
 
+def test_transformer_delete_model(sagemaker_session):
```
ChoiByungWook: I think we can maybe make use of the [timeout](https://github.com/aws/sagemaker-python-sdk/blob/7f33ac79cd4b10f58eafd7a1b2f9af0dc5c48edd/tests/integ/timeout.py) class for handling delete model.

Otherwise, I wouldn't advise going through the training and transform in a whole new test in order to test the delete_model method, since you are only testing deleting the corresponding model and don't necessarily care about the training or deploy portion, since this is the transformer and we can assume they have to come in with a SageMaker model. I think it is okay to use an existing test and modify it to include deleting of the model.

Either that, or create a test that reuses the model of another test.

tests/unit/test_local_session.py
```diff
@@ -156,6 +156,19 @@ def test_create_model(LocalSession):
     assert 'my-model' in sagemaker.local.local_session.LocalSagemakerClient._models
 
 
+@patch('sagemaker.local.local_session.LocalSession')
+def test_delete_model(LocalSession):
+    local_sagemaker_client = sagemaker.local.local_session.LocalSagemakerClient()
+    model_name = 'my-model'
```
ChoiByungWook: You don't have to make this change, however I believe some tests are sharing common variables, such as this one. I think it would great if we could declare these near the top, so that they can be shared. :D

tests/unit/test_local_session.py
```diff
@@ -316,7 +342,7 @@ def test_update_endpoint(LocalSession):
     endpoint_name = 'my-endpoint'
     endpoint_config = 'my-endpoint-config'
     expected_error_message = 'Update endpoint name is not supported in local session.'
-    with pytest.raises(NotImplementedError, message=expected_error_message):
+    with pytest.raises(NotImplementedError, match=expected_error_message):
```
ChoiByungWook: Nice catch!

README.rst
```diff
@@ -321,6 +326,7 @@ Here is an end-to-end example:
     transformer = mxnet_estimator.transformer(1, 'local', assemble_with='Line', max_payload=1)
     transformer.transform('s3://my/transform/data, content_type='text/csv', split_type='Line')
     transformer.wait()
+    transformer.delete_model()
```
ChoiByungWook: Can we put a small comment above similar to the one above in regards to delete_endpoint and add a newline.

'# Deletes the SageMaker model'

CHANGELOG.rst
```diff
@@ -7,6 +7,8 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
+* feature: Support for predictor to delete endpoint and endpoint configuration
```
ChoiByungWook: You also added support to delete model through the Model class.

I would also call out that we by default delete the endpoint configuration in ``delete_endpoint``

README.rst
```diff
@@ -190,9 +190,14 @@ Here is an end to end example of how to use a SageMaker Estimator:
     response = mxnet_predictor.predict(data)
 
     # Tears down the SageMaker endpoint
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
 
 
+The above example will delete endpoint and endpoint configuration at the same time. If you want to keep endpoint configuration, you can do the following:
```
ChoiByungWook: Drop in.

"The example above will eventually delete both the SageMaker endpoint and endpoint configuration through ``delete_endpoint()``. If you want to keep your SageMaker endpoint configuration, use the value False for the ``delete_endpoint_config`` parameter, as shown below."

README.rst
```diff
@@ -190,9 +190,14 @@ Here is an end to end example of how to use a SageMaker Estimator:
     response = mxnet_predictor.predict(data)
 
     # Tears down the SageMaker endpoint
```
ChoiByungWook: Let's update this comment to reflect that both the endpoint and endpoint configuration are removed.

README.rst
```diff
@@ -190,9 +190,14 @@ Here is an end to end example of how to use a SageMaker Estimator:
...
-    mxnet_estimator.delete_endpoint()
+    mxnet_predictor.delete_endpoint()
 
 
+The above example will delete endpoint and endpoint configuration at the same time. If you want to keep endpoint configuration, you can do the following:
+
+.. code:: python
+    mxnet_predictor.delete_endpoint(delete_endpoint_config=False)
```
ChoiByungWook: Add a small comment similar to the one above stating that only the endpoint is deleted.

src/sagemaker/predictor.py
ChoiByungWook: You don't have to do a try-except within the _delete_endpoint_config. My main concern was that the code was doing a [delete_endpoint -> delete_endpoint_configuration](https://github.com/aws/sagemaker-python-sdk/blob/7f33ac79cd4b10f58eafd7a1b2f9af0dc5c48edd/src/sagemaker/predictor.py#L124), when the delete_endpoint_configuration calls describe_endpoint, which depends on the endpoint.

As long as we delete endpoint_configuration first and then endpoint, I think we should be fine and don't need to have extra try catch logic within the _delete_endpoint_config.

CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
+* feature: Support for predictor to delete endpoint, and delete endpoint configuration with ``delete_endpoint()`` by default
```
ChoiByungWook: I think the ability to delete endpoint has always been there. This should reflect that the ability to delete endpoint configuration was added.

README.rst
ChoiByungWook: "endpoint endpoint", there are double "endpoint" in here.

Drop-in

Only delete the SageMaker endpoint, while keeping the corresponding endpoint configuration.

CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
+* feature: Make predictor to delete endpoint configuration by default when calling ``delete_endpoint()``
```
ChoiByungWook: Support to for predictor class to delete endpoint configuration by default when calling ``delete_endpoint()``

src/sagemaker/predictor.py
```diff
@@ -105,9 +105,25 @@ def _create_request_args(self, data, initial_args=None):
...
-        """Delete the Amazon SageMaker endpoint backing this predictor.
+    def _delete_endpoint_config(self):
+        """Delete the Amazon SageMaker endpoint configuration
         """
+        endpoint_description = self.sagemaker_session.sagemaker_client.describe_endpoint(EndpointName=self.endpoint)
+        endpoint_config_name = endpoint_description['EndpointConfigName']
+        self.sagemaker_session.delete_endpoint_config(endpoint_config_name)
+
+    def delete_endpoint(self, delete_endpoint_config=True):
+        """Delete the Amazon SageMaker endpoint backing this predictor. Also delete the endpoint configuration attached
```
ChoiByungWook: I think it is okay to say:

"Delete the Amazon SageMaker endpoint and endpoint configuration backing this predictor."

Since it is deleted by default.

src/sagemaker/predictor.py
```diff
@@ -105,9 +105,25 @@ def _create_request_args(self, data, initial_args=None):
...
+           to it if delete_endpoint_config is True.
+
+        Args:
+            delete_endpoint_config (bool): Flag to indicate whether to delete endpoint configuration together with
```
ChoiByungWook: Drop-in

"Flag to indicate whether or not to delete the corresponding SageMaker endpoint configuration tied to the endpoint. If False, only the endpoint will be deleted. (default: True)"

Nitpick: Please wrap the default with parentheses. Similar to https://github.com/aws/sagemaker-python-sdk/blob/757c8ab8a48000344e37bc0683348e9060ffde17/src/sagemaker/predictor.py#L51

src/sagemaker/session.py
```diff
@@ -780,6 +780,31 @@ def delete_endpoint(self, endpoint_name):
...
+        """Delete an Amazon SageMaker ``Model``.
+
+        Args:
+            model_name (str): Name of the Amazon SageMaker model to delete.
+
+        Raises: ValueError if model is not deployed yet.
+
+        """
+        try:
+            self.sagemaker_client.describe_model(ModelName=model_name)
```
ChoiByungWook: My mistake, I think keeping the none check within the model class makes sense because that class handles the deploy, while in the session class it should just call the API on it's own.

This is a little dangerous as the corresponding exception message is catering towards the model not being deployed, when in fact it can exist, however not have a corresponding endpoint.

I think having this method just do sagemaker_client.deleteModel() is fine. You shouldn't do the extra check here.

src/sagemaker/transformer.py
```diff
@@ -52,6 +52,8 @@ def __init__(self, model_name, instance_count, instance_type, strategy=None, ass
                 using the default AWS configuration chain.
             volume_kms_key (str): Optional. KMS key ID for encrypting the volume attached to the ML
                 compute instance (default: None).
+            model (sagemaker.model.Model): A SageMaker Model object, used for SageMaker Model interactions
```
ChoiByungWook: this docstring shouldn't be here.

src/sagemaker/transformer.py
```diff
@@ -112,6 +114,12 @@ def transform(self, data, data_type='S3Prefix', content_type=None, compression_t
         self.latest_transform_job = _TransformJob.start_new(self, data, data_type, content_type, compression_type,
                                                             split_type)
 
+    def delete_model(self):
+        """Delete a SageMaker Model.
```
ChoiByungWook: drop-in

"Delete the corresponding SageMaker model for this Transformer."

tests/integ/timeout.py
```diff
@@ -87,6 +87,54 @@ def timeout_and_delete_endpoint_by_name(endpoint_name, sagemaker_session, second
...
+                    if no_errors:
+                        _cleanup_model_logs(transformer.model_name, sagemaker_session)
+                        return
+                except ClientError as ce:
+                    if ce.response['Error']['Code'] == 'ValidationException':
+                        pass
+                sleep(10)
+
+
+def _show_model_logs(model_name, sagemaker_session):
```
ChoiByungWook: is this being used or going to be used anywhere?

tests/integ/timeout.py
```diff
@@ -87,6 +87,54 @@ def timeout_and_delete_endpoint_by_name(endpoint_name, sagemaker_session, second
...
+        LOGGER.info('cloudwatch logs for log group {}'.format(log_group))
+        logs = AWSLogs(log_group_name=log_group, log_stream_name='ALL', start='1d',
+                       aws_region=sagemaker_session.boto_session.region_name)
+        logs.list_logs()
+    except Exception:
+        LOGGER.exception('Failure occurred while listing cloudwatch log group %s. Swallowing exception but printing '
+                         'stacktrace for debugging.', log_group)
+
+
+def _cleanup_model_logs(model_name, sagemaker_session):
```
ChoiByungWook: this method looks really similar to _cleanup_endpoint_logs. The only difference is the log_group, can we maybe modify that method to allow for this use case as well?

tests/unit/test_transformer.py
```diff
@@ -59,6 +59,13 @@ def transformer(sagemaker_session):
                        volume_kms_key=KMS_KEY_ID)
 
 
+def test_delete_model(sagemaker_session):
+    transformer = Transformer(MODEL_NAME, INSTANCE_COUNT, INSTANCE_TYPE, output_path=OUTPUT_PATH,
```
ChoiByungWook: Feel free to remove the optional parameters.

https://github.com/aws/sagemaker-python-sdk/blob/757c8ab8a48000344e37bc0683348e9060ffde17/src/sagemaker/transformer.py#L24

README.rst
```diff
@@ -274,8 +280,8 @@ We can take the example in  `Using Estimators <#using-estimators>`__ , and use e
     # Serializes data and makes a prediction request to the local endpoint
     response = mxnet_predictor.predict(data)
 
-    # Tears down the endpoint container
-    mxnet_estimator.delete_endpoint()
+    # Tears down the endpoint and endpoint configuration
```
ChoiByungWook: this is still in regards to local mode. So it should state that it "Tear downs the endpoint container and deletes the corresponding endpoint configuration"

README.rst
```diff
@@ -297,7 +303,7 @@ Here is an end-to-end example:
     data = numpy.zeros(shape=(1, 1, 28, 28))
     predictor.predict(data)
 
-    # Tear down the endpoint container
+    # Tear down the endpoint and endpoint configuration
```
ChoiByungWook: same thing here.

tests/integ/timeout.py
ChuyangDeng: This is being used in the integ/test_transformer only.

tests/integ/timeout.py
ChuyangDeng: I modified the _show_logs and _cleanup_logs to take a resource_type parameter to differentiate Models and Endpoints log_group.

tests/integ/test_transformer.py
ChoiByungWook: You can replace the timeout with your time function. Place the assertion within the timeout.

tests/unit/test_model.py
```diff
@@ -335,3 +335,21 @@ def test_model_package_create_transformer_with_product_id(sagemaker_session):
...
+    model = DummyFrameworkModel(sagemaker_session, source_dir=str(tmpdir))
+    model.deploy(instance_type=INSTANCE_TYPE, initial_instance_count=1)
+    model.delete_model()
+
+    sagemaker_session.delete_model.assert_called_with(model.name)
+
+
+@patch('sagemaker.fw_utils.tar_and_upload_dir', MagicMock())
+@patch('time.strftime', MagicMock(return_value=TIMESTAMP))
+def test_delete_non_deployed_model(sagemaker_session, tmpdir):
```
ChoiByungWook: Do you need tmpdir or to mock tar_and_upload_dir?

The code path doesn't go through deploy() in this test.

tests/unit/test_transformer.py
ChoiByungWook: Do you need the output_path or volume_kms_key?

tests/integ/test_transformer.py
ChuyangDeng: This test is deleted because I replaced the timeout with the updated timeout function.

tests/unit/test_transformer.py
ChuyangDeng: This was updated in previous commit.

src/sagemaker/model.py
```diff
@@ -302,6 +302,16 @@ def transformer(self, instance_count, instance_type, strategy=None, assemble_wit
                            env=env, tags=tags, base_transform_job_name=self.name,
                            volume_kms_key=volume_kms_key, sagemaker_session=self.sagemaker_session)
 
+    def delete_model(self):
+        """Delete an Amazon SageMaker ``Model``.
+
+        Raises: ValueError if model is not deployed yet.
```
laurenyu: this should be formatted:

```
Raises:
    ValueError: if the model has not been created
```

also I think you mean "created" not "deployed" - only once a SageMaker model is created can it then be deployed

src/sagemaker/model.py
```diff
@@ -302,6 +302,16 @@ def transformer(self, instance_count, instance_type, strategy=None, assemble_wit
                            env=env, tags=tags, base_transform_job_name=self.name,
                            volume_kms_key=volume_kms_key, sagemaker_session=self.sagemaker_session)
 
+    def delete_model(self):
+        """Delete an Amazon SageMaker ``Model``.
```
laurenyu: don't put "Model" in backticks here - it's not referring to the Python class

CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ CHANGELOG
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
+* feature: Support for predictor class to delete endpoint configuration by default when calling ``delete_endpoint()``
+* feature: Support for model class to delete SageMaker model
+* feature: Support for transformer to delete Sagemaker model
```
laurenyu: for class names, capitalize them and put them in backticks

src/sagemaker/session.py
```diff
@@ -780,6 +780,25 @@ def delete_endpoint(self, endpoint_name):
...
+        """Delete an Amazon SageMaker endpoint configuration.
+
+        Args:
+            endpoint_config_name (str): Name of the Amazon SageMaker endpoint configuration to delete.
+        """
+        LOGGER.info('Deleting endpoint configuration with name: {}'.format(endpoint_config_name))
+        self.sagemaker_client.delete_endpoint_config(EndpointConfigName=endpoint_config_name)
+
+    def delete_model(self, model_name):
+        """Delete an Amazon SageMaker ``Model``.
```
laurenyu: same here

src/sagemaker/model.py
ChuyangDeng: I updated the doc strings here and below as well as test's expected error message.

CHANGELOG.rst
```diff
@@ -8,6 +8,9 @@ CHANGELOG
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
 * doc-fix: add ``transform_fn`` information and fix ``input_fn`` signature in the MXNet README
+* feature: Support for ``Predictor`` to delete endpoint configuration by default when calling ``delete_endpoint()``
+* feature: Support for ``model`` to delete SageMaker model
```
ChoiByungWook: the model should be uppercase, but we can fix in the other PR.

PR #638, opened 2019-02-11
Model linkage

ChuyangDeng:
*Issue #, if available:*
#447 

*Description of changes:*
Continuation of https://github.com/aws/sagemaker-python-sdk/pull/630
Let `Predictor` delete SageMaker model directly in align with `Transformer`.

## Merge Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ CHANGELOG
 ==========
 
 * doc-fix: Remove incorrect parameter for EI TFS Python README
+* feature: ``Predictor``: delete SageMaker model
+* feature: ``Pipeline``: delete SageMaker model
 
 1.18.3.post1
 ============
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,8 @@
 CHANGELOG
 =========
 
-1.18.3.dev
-==========
+1.18.3
+======
 
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,13 @@
 CHANGELOG
 =========
 
+<<<<<<< HEAD
 1.18.4.dev
 ==========
+=======
+1.18.3
+======
+>>>>>>> Bump version to v1.18.3 (#641)
 
 * doc-fix: Remove incorrect parameter for EI TFS Python README
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,8 +8,8 @@ CHANGELOG
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
 * doc-fix: add ``transform_fn`` information and fix ``input_fn`` signature in the MXNet README
-* feature: Support for ``Predictor`` to delete endpoint configuration by default when calling ``delete_endpoint()``
-* feature: Support for ``model`` to delete SageMaker model
+* feature: Support for ``Predictor`` to delete model, and delete endpoint configuration by default when calling ``delete_endpoint()``
+* feature: Support for ``Model`` to delete SageMaker model
 * feature: Support for ``Transformer`` to delete SageMaker model
 * bug-fix: Default account for SKLearnModel fixed
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,13 +2,8 @@
 CHANGELOG
 =========
 
-<<<<<<< HEAD
 1.18.4.dev
 ==========
-=======
-1.18.3
-======
->>>>>>> Bump version to v1.18.3 (#641)
 
 * doc-fix: Remove incorrect parameter for EI TFS Python README
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,8 +5,7 @@ CHANGELOG
 1.18.4.dev
 ==========
 
-* feature: ``Predictor``: delete SageMaker Model
-* feature: ``Pipeline``: delete SageMaker Model
+* doc-fix: Remove incorrect parameter for EI TFS Python README
 
 1.18.3.post1
 ============
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 CHANGELOG
 =========
 
+1.18.3.post1
+============
+
+* doc-fix: fix README for PyPI
+
 1.18.3
 ======
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 CHANGELOG
 =========
 
+1.18.4.dev
+==========
+
+* feature: ``Predictor``: delete SageMaker Model
+* feature: ``Pipeline``: delete SageMaker Model
+
 1.18.3.post1
 ============
 
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ if sys.version_info < (3, 4):
     required_packages.append('enum34>=1.1.6')
 
 setup(name="sagemaker",
-      version='1.18.3',
+      version='1.18.3.post1',
       description="Open source library for training and deploying models on Amazon SageMaker.",
       packages=find_packages('src'),
       package_dir={'': 'src'},
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ if sys.version_info < (3, 4):
     required_packages.append('enum34>=1.1.6')
 
 setup(name="sagemaker",
-      version='1.18.2',
+      version='1.18.3',
       description="Open source library for training and deploying models on Amazon SageMaker.",
       packages=find_packages('src'),
       package_dir={'': 'src'},
--- a/src/sagemaker/predictor.py
+++ b/src/sagemaker/predictor.py
@@ -134,11 +134,15 @@ class RealTimePredictor(object):
         for model_name in self._model_names:
             self.sagemaker_session.delete_model(model_name)
 
-    def _get_model_names(self):
+    def _get_endpoint_config_desc(self):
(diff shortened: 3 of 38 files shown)


src/sagemaker/predictor.py
```diff
@@ -49,13 +49,15 @@ def __init__(self, endpoint, sagemaker_session=None, serializer=None, deserializ
             content_type (str): The invocation's "ContentType", overriding any ``content_type`` from
                 the serializer (default: None).
             accept (str): The invocation's "Accept", overriding any accept from the deserializer (default: None).
+            model (sagemaker.model.Model): A SageMaker Python SDK Model object.
```
ChoiByungWook: Default? What happens when it is None?

src/sagemaker/predictor.py
```diff
@@ -105,11 +107,35 @@ def _create_request_args(self, data, initial_args=None):
...
+            delete_endpoint_config (bool): Flag to indicate whether to delete the corresponding SageMaker endpoint
+                configuration tied to the endpoint. If False, only the endpoint will be deleted. (default: True)
+
+        """
+        if delete_endpoint_config:
+            self._delete_endpoint_config()
+
         self.sagemaker_session.delete_endpoint(self.endpoint)
 
+    def get_model(self):
```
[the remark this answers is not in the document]
ChoiByungWook: What will happen here if a user created the predictor outside of estimator or model.deploy().

It will return None I believe, and there isn't a nice way of getting the model.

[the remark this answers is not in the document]
ChuyangDeng: As we discussed offline, we will modify ``Predictor`` to delete the model directly, in align with ``Transformer``.

CHANGELOG.rst
```diff
@@ -8,8 +8,8 @@ CHANGELOG
 * doc-fix: update information about saving models in the MXNet README
 * doc-fix: change ReadTheDocs links from latest to stable
 * doc-fix: add ``transform_fn`` information and fix ``input_fn`` signature in the MXNet README
-* feature: Support for ``Predictor`` to delete endpoint configuration by default when calling ``delete_endpoint()``
-* feature: Support for ``model`` to delete SageMaker model
+* feature: Support for ``Predictor`` to delete model, and delete endpoint configuration by default when calling ``delete_endpoint()``
```
ChoiByungWook: Your feature addition needs to be separate and in a new version, as your last PR go merged and released to PyPI already.

https://github.com/aws/sagemaker-python-sdk/releases/tag/v1.18.3

src/sagemaker/predictor.py
```diff
@@ -114,18 +114,35 @@ def _delete_endpoint_config(self):
...
     def delete_endpoint(self, delete_endpoint_config=True):
-        """Delete the Amazon SageMaker endpoint and endpoint configuration backing this predictor.
+        """Delete the Amazon SageMaker endpoint backing this predictor. Also delete the endpoint configuration attached
+           to it if delete_endpoint_config is True.
 
         Args:
-            delete_endpoint_config (bool): Flag to indicate whether to delete the corresponding SageMaker endpoint
-                configuration tied to the endpoint. If False, only the endpoint will be deleted. (default: True)
+            delete_endpoint_config (bool): Flag to indicate whether to delete endpoint configuration together with
+                endpoint. If False, only endpoint will be deleted. Default: True.
```
ChoiByungWook: Let's follow this for our docstrings.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

src/sagemaker/predictor.py
```diff
@@ -114,18 +114,35 @@ def _delete_endpoint_config(self):
...
+        for model_name in model_names:
+            self.sagemaker_session.delete_model(model_name)
+
+    def _get_model_names(self):
+        endpoint_desc = self.sagemaker_session.sagemaker_client.describe_endpoint(EndpointName=self.endpoint)
+        endpoint_config_name = endpoint_desc['EndpointConfigName']
+        endpoint_config = self.sagemaker_session.sagemaker_client.describe_endpoint_config(
+            EndpointConfigName=endpoint_config_name)
+        production_variants = endpoint_config['ProductionVariants']
+        return map(lambda d: d['ModelName'], production_variants)
```
ChoiByungWook: It looks like it is possible for multiple models to exist behind an endpoint configuration and endpoint.

Let's make sure this class also has the delete_model functionality as well.
https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/pipeline.py

Here is an example on how it is done:
https://github.com/awslabs/amazon-sagemaker-examples/blob/master/sagemaker-python-sdk/scikit_learn_inference_pipeline/Inference%20Pipeline%20with%20Scikit-learn%20and%20Linear%20Learner.ipynb

We should make sure the deletion of multiple models in the same endpoint configuration is tested in an integ test.

tests/integ/test_kmeans.py
```diff
@@ -75,6 +75,11 @@ def test_kmeans(sagemaker_session):
             assert record.label["closest_cluster"] is not None
             assert record.label["distance_to_cluster"] is not None
 
+    predictor.delete_model()
```
ChoiByungWook: is there an integ test that successfully deletes the model?

src/sagemaker/predictor.py
ChoiByungWook: Something to consider is that this solution assumes the endpoint configuration exists. It is possible for a user to call delete_endpoint() and that might error out when they call delete_model().

Technically, the model object does still exist within SageMaker. It arguably makes sense to actually delete the endpoint first, then endpoint configuration and then model since they all have dependency on each other. I believe SageMaker allows you to delete them regardless of order, since they snapshot everything, however this problem can still persist if the user deletes the endpoint first.

tests/unit/test_predictor.py
```diff
@@ -466,3 +466,21 @@ def test_delete_endpoint_only():
 
     sagemaker_session.delete_endpoint.assert_called_with(ENDPOINT)
     sagemaker_session.delete_endpoint_config.assert_not_called()
+
+
+def test_delete_model():
+    endpoint_desc = {
+        'EndpointConfigName': 'my-endpoint-config'
+    }
+    endpoint_config_desc = {
```
ChoiByungWook: can we unit test an endpoint configuration with multiple models?

CHANGELOG.rst
ChuyangDeng: ohh. So the next version should be 1.18.4.dev?

src/sagemaker/predictor.py
ChuyangDeng: Just to confirm, for `Predictor`, `delete_model()` should delete all the SageMaker models from the list[sagemaker.Model] and delete the model with `name` (if `name` is not None)?

And for the order of deletions, if endpoint/endpoint config is deleted before attempting to delete model, should we say something like "model cannot be deleted because the attached endpoint/endpoint config is already gone"?

tests/integ/test_kmeans.py
ChuyangDeng: The existing integ tests only delete endpoint, except for test_transformer which deletes model.

tests/unit/test_predictor.py
ChuyangDeng: Yes, will do this in next revision.

tests/integ/test_kmeans.py
ChoiByungWook: Disregard this comment. This test deletes the model. You're doing a describe to confirm it.

CHANGELOG.rst
icywang86rui: I would say 'add support for..." to be consistent

CHANGELOG.rst
```diff
@@ -8,8 +8,8 @@ CHANGELOG
...
+* feature: Support for ``Model`` to delete SageMaker model
 * feature: Support for ``Transformer`` to delete SageMaker model
```
icywang86rui: I guess we are already not consistent -_- . So it's up to you

src/sagemaker/predictor.py
icywang86rui: Let's initialize the model name in the constructor per our in person discussion.

ChuyangDeng: We're going to call `describe_endpoint()` and `describe_endpoint_config()` once the `Predictor` gets `endpoint_name` in the constructor, so that we don't restrict the order of deletions to model -> endpoint/endpoint config.

CHANGELOG.rst
laurenyu: even better would be something like "feature: Predictor: delete model and endpoint configuration by default when calling `delete_endpoint()`", though we've been extremely inconsistent in keeping that format

src/sagemaker/predictor.py
ChoiByungWook: Change the existing delete_endpoint() functionality to depend on the endpoint_configuration_name instead of describing it.

https://github.com/aws/sagemaker-python-sdk/blob/d7a2bf6be25e8da5ef0ec0e8273c3f9a8d9053d5/src/sagemaker/predictor.py#L112

src/sagemaker/predictor.py
```diff
@@ -109,23 +110,38 @@ def _delete_endpoint_config(self):
...
         self.sagemaker_session.delete_endpoint(self.endpoint)
 
+    def delete_model(self):
+        """Deletes the Amazon SageMaker models backing this predictor.
+
+        """
+        for model_name in self._model_names:
+            self.sagemaker_session.delete_model(model_name)
+
+    def _get_model_names(self):
```
ChoiByungWook: this method does more than get the model names, it also get's the endpoint config name and sets it.

setup.py
```diff
@@ -33,7 +33,7 @@ def read(fname):
     required_packages.append('enum34>=1.1.6')
 
 setup(name="sagemaker",
-      version='1.18.2',
+      version='1.18.3.post1',
```
ChoiByungWook: your branch might be out of date?

ChoiByungWook:
Closing in favor of https://github.com/aws/sagemaker-python-sdk/pull/647

PR #1032, opened 2019-09-09
change: clean up git support integ tests

laurenyu:
*Description of changes:*
follow-up to #1030 and #1029. In addition to removing (more) credentials, I also tried to clean up the code a bit:

* renamed the tests so that it's clearer what they're supposed to be testing (rather than which arbitrary framework was chosen for testing purposes)
* cleaned up the logic around deploying to an endpoint with MXNet, which for some reason was constructing the Model class instance separately
* removed the skip based on Python version - we're testing the Python SDK logic, not the SKLearn image functionality, so it should be fine to use the Python 3 image all test runs.

## Merge Checklist


tests/integ/test_git.py
```diff
@@ -126,46 +123,33 @@ def test_git_support_with_mxnet(sagemaker_local_session):
...
+            predictor = mx.deploy(1, "local", entry_point=serving_script_path)
+
             data = numpy.zeros(shape=(1, 1, 28, 28))
             result = predictor.predict(data)
             assert result is not None
         finally:
             predictor.delete_endpoint()
 
 
-@pytest.mark.skipif(PYTHON_VERSION != "py3", reason="Scikit-learn image supports only python 3.")
```
knakad: Can we keep this here but commented out?  
Otherwise, if the tribal knowledge is lost, the dev that looks into fixing this might be confused, and may not think to look at git logs.

laurenyu: I'll leave a comment in the code. I don't want to leave a commented-out annotation, though.

knakad: Works for me =)

PR #1041, opened 2019-09-13
change: use Estimator.create_model in Estimator.transformer

laurenyu:
*Description of changes:*
This change in the `Estimator.transformer` implementation should help make sure various parameters don't get lost when going from an estimator to a model. This was specifically seen with network isolation, but there might be similar things in the future.

## Merge Checklist

### DIFF
--- a/tests/integ/test_transformer.py
+++ b/tests/integ/test_transformer.py
@@ -279,13 +279,14 @@ def test_transform_byo_estimator(sagemaker_session, cpu_instance_type):
     with timeout(minutes=TRAINING_DEFAULT_TIMEOUT_MINUTES):
         kmeans.fit(records, job_name=job_name)
 
+    estimator = Estimator.attach(training_job_name=job_name, sagemaker_session=sagemaker_session)
+    estimator._enable_network_isolation = True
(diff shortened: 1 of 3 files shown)


tests/integ/test_transformer.py
```diff
@@ -297,6 +298,8 @@ def test_transform_byo_estimator(sagemaker_session, cpu_instance_type):
         model_desc = sagemaker_session.sagemaker_client.describe_model(
             ModelName=transformer.model_name
         )
+        assert model_desc["EnableNetworkIsolation"]
```
knakad: [Food for thought] Is there a way to verify a wider range of variables are getting passed in properly using some sort of a dynamic check?

laurenyu: I think we'd have to look for a struct that's identical to the output of `DescribeModel` for us to be able to do that dynamically. It wouldn't be technically hard to implement some kind of method to do that for Estimator/Model/etc. classes, but I'm not entirely sure how that would fit in. It does also begin to sound similar (but not the same) to what we have for workflows/Airflow.

PR #1061, opened 2019-09-23
fix: enable kms support for repack_model

icywang86rui:
Currently repack_model doesn't accept a kms key. This change
added a kms_key argument to the fucntion. In addition repack_model
will always use the output_kms_key inside the Estimator if it's set.

*Issue #, if available:*

*Description of changes:*

## Merge Checklist


src/sagemaker/chainer/estimator.py
```diff
@@ -162,6 +162,7 @@ def create_model(
         entry_point=None,
         source_dir=None,
         dependencies=None,
+        **kwargs
```
laurenyu: is there something here that should actually get used?

src/sagemaker/estimator.py
```diff
@@ -547,6 +547,7 @@ def deploy(
                 )
             model = self._compiled_models[family]
         else:
+            kwargs["output_kms_key"] = self.output_kms_key
```
laurenyu: I realize that I'm probably the source of this problem (with one of my recent code changes), but should we distinguish more between `output_kms_key` and `kms_key`? also, is there a case where someone might want to specify `output_kms_key` in `deploy`?

src/sagemaker/model.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         vpc_config=None,
         sagemaker_session=None,
         enable_network_isolation=False,
+        output_kms_key=None,
```
laurenyu: docstring? (same for the rest)

src/sagemaker/chainer/estimator.py
icywang86rui: I am only fixing TensorFlow and SKLearn for now. I think all other frameworks needs more fixes than this one(similar to the fixes I added to sklearn). That work should be properly scheduled and we should add integ tests as well.

src/sagemaker/model.py
icywang86rui: will add

src/sagemaker/estimator.py
icywang86rui: When we repack the model the new model tar ball is uploaded to the same bucket. we need to use the same output_kms_key in the estimator, otherwise uploading the repacked tar ball will fail. 

I can't really think of any use case where you would want a different kms key since we don't allow user to specify a new bucket when deploy is called.

laurenyu: maybe let's rename the `Model`'s `output_kms_key` to something like `model_data_kms_key`? (`model_kms_key`?)

src/sagemaker/model.py
```diff
@@ -114,6 +115,8 @@ def __init__(
                 network isolation in the endpoint, isolating the model
                 container. No inbound or outbound network calls can be made to
                 or from the model container.
+            output_kms_key (str): KMS key ARN, it's only used to encrypt the
+                repacked model archive file.
```
laurenyu: "KMS key ARN used to encrypt the repacked model archive file if the model is repacked"

src/sagemaker/model.py
laurenyu: `**kwargs` also need docstring entries - if possible, include a link to where the `**kwargs` are eventually sent to

src/sagemaker/estimator.py
icywang86rui: i will do model_kms_key

src/sagemaker/model.py
icywang86rui: k

src/sagemaker/model.py
icywang86rui: k

src/sagemaker/tensorflow/estimator.py
```diff
@@ -537,6 +538,8 @@ def create_model(
                 If not specified and ``endpoint_type`` is 'tensorflow-serving', ``dependencies`` is
                 set to ``None``.
                 If ``endpoint_type`` is also ``None``, then the dependencies from training are used.
+            **kwargs: Additional kwargs passed to ``_create_tfs_model`` and
+                ``_create_default_model``.
```
laurenyu: I think you'd be better off mentioning the model class names rather than these private methods

PR #1087, opened 2019-10-12
fix: Validation args now use default framework_version for TensorFlow

hyche:
Which should be passed modified variable.

*Description of changes:*
framework_version is passed from passed argument but not modified self.framework_version, which leads to false validation, although it shouldn't

## Merge Checklist


src/sagemaker/tensorflow/estimator.py
```diff
@@ -302,7 +302,7 @@ def __init__(
         self._validate_args(
             py_version=py_version,
             script_mode=script_mode,
-            framework_version=framework_version,
+            framework_version=self.framework_version,
```
ChoiByungWook: Nice catch!

We should probably pass in the self referenced version of all the passed parameters in order to keep it consistent. Is that okay with you?

ChoiByungWook: Actually... I'm wondering if it makes more sense to validate before defaulting. Since there is a chance that a user wants to use X, but due to this change it will default and pass when in theory it should have failed. Your thoughts?

hyche: I don't think I understand what you meant. Because the documents says
` framework_version (str): ..If not specified, this will default to 1.11.`
That's why I made this change, so we should validate after defaulting right, otherwise if will throw exception on `None` value whereas it shouldn't ? However, I think there is better way to fix this, ie. Following this approach (https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/tensorflow/model.py#L56), which using not-None default arguments.

ChoiByungWook: @hyche, I agree with you.

As for the better way of fixing it, I think we can't default at the parameter level due to this method expecting a None: https://github.com/aws/sagemaker-python-sdk/blob/d505210e5352aaffaa6546127ca3d746493c88ff/src/sagemaker/tensorflow/estimator.py#L282

The original code change you provided works!

PR #1144, opened 2019-12-04
feature: support Multi-Model endpoints

dlragha:
*Issue #, if available:*

*Description of changes:*
Add support to create Multi-Model endpoints via Python SDK with unit and integration tests. 
Feature details: https://docs.aws.amazon.com/sagemaker/latest/dg/multi-model-endpoints.html

## Merge Checklist


src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+                `model.image` is used instead. (default: None)
+            predictor_cls (callable[string, sagemaker.session.Session]): A
+                function to call to create a predictor (default: None). If not
+                None, ``deploy`` will return the result of invoking this
+                function on the created endpoint name.
+            **kwargs: Keyword arguments passed to the ``Model`` initializer.
+        """
+        self.model_name = model_name
+        self._model_data_prefix = model_data_prefix
+        self.model_data_prefix = model_data_prefix
```
ChuyangDeng: What is the difference between the private and public `model_data_prefix` here?

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+        self.model = model
+        self.container_mode = MULTI_MODEL_CONTAINER_MODE
+        self.image = image
+        self.role = role
+        self.predictor_cls = predictor_cls
+
+        # Copy the values from the model if these attributes are not provided in the constructor
+        if self.model:
+            self.image = self.image or self.model.image
+            self.role = self.role or self.model.role
```
ChuyangDeng: Is it possible that `MultiDataModel` and `Model` have conflicting images or roles? What would be the consequence in that case?

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+            self.predictor_cls = self.predictor_cls or self.model.predictor_cls
+
+        super(MultiDataModel, self).__init__(
+            self.model_data_prefix,
+            self.image,
+            self.role,
+            name=self.model_name,
+            predictor_cls=self.predictor_cls,
+            **kwargs
+        )
```
fm1ch4: if `self.model` has a vpc_config, doesn't it get ignored when .deploy() is called? same with enable_network_isolation.

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if not self.sagemaker_session:
+            self.sagemaker_session = sagemaker.session.Session()
```
ChuyangDeng: We usually have a `sagemaker_session` arg in the constructor, and do `self.sagemaker_session = sagemaker_session or Session()`

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+        Args:
+        s3_url: S3 path of the trained model artifact
+        model_data_path: S3 path where the trained model artifact
+                should be uploaded relative to `self.model_data_prefix`.
+                (default: None). If None, then the entire s3_url path
+                (after the source bucketname) will be copied to
+                `model_data_prefix` location
+        """
+        # Validate s3_url
+        if not s3_url.startswith("s3://"):
```
ChuyangDeng: Is it possible that we also support local files? Python SDK can take care of uploading local model files to S3 paths using S3Uploader: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/s3.py#L37

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+                # Get the metadata about each object returned
+                s3_objects_metadata = page["Contents"]
+                for s3_object in s3_objects_metadata:
+                    # Return the model paths relative to the model_data_prefix
+                    # Ex: "a/b/c.tar.gz" -> "b/c.tar.gz" if url_prefix = "a/"
+                    yield s3_object["Key"].replace(url_prefix, "")
+            except KeyError:
+                return
+
+    def _parse_s3_uri(self, s3_url):
```
ChuyangDeng: This method can go into: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/s3.py with a different name

tests/data/multimodel/container/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+RUN mkdir -p /home/model-server/
+
+# Copy the default custom service file to handle incoming data and inference requests
+COPY model_handler.py /home/model-server/model_handler.py
+
+# Define an entrypoint script for the docker image
+ENTRYPOINT ["python", "/usr/local/bin/dockerd-entrypoint.py"]
+
+# Define command to be passed to the entrypoint
+CMD ["serve"]
```
ChuyangDeng: new line at EOF.

src/sagemaker/multidatamodel.py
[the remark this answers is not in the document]
ChuyangDeng: That method does different things, but we could move this `_parse_s3_uri` to sagemaker/s3 and name it differently.

dlragha:
> looks like dummy_model.tar.gz is 40MB - it could be smaller since its contents don't really matter. or you could use one of the existing .tar.gz in the repo.

Good point! Removed this model and reusing an existing .tar.gz from the repo.

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Placeholder docstring"""
```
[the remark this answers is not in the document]
dlragha: Fixed

src/sagemaker/multidatamodel.py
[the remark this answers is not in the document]
dlragha: We needed the private variable as a helper for the validation that happens in the model_data_prefix.setter and self.model_data_prefix to actually set the value (which also triggers the setter and in the process validates the input). But in the new commit, I rewrote the logic to directly do the validation in init to avoid confusion

tests/data/multimodel/container/model_handler.py
```diff
@@ -0,0 +1,90 @@
...
+        :param data: list of objects, raw input from request
+        :param context: model server context
+        :return: list of target model to send back to client
+        """
+        self.error = None  # reset earlier errors
+
+        try:
+            preprocessed_data = self.preprocess(data)
+            inference_result = self.inference(preprocessed_data)
+            postprocessed_data = self.postprocess(inference_result)
```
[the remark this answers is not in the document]
dlragha: Fixed

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+
+MULTI_MODEL_CONTAINER_MODE = "MultiModel"
+
+
+class MultiDataModel(Model):
+    """SageMaker ``MultiDataModel`` used to deploy multiple models to the same ``Endpoint``.
+
+    This class defines methods to create a Model with Multi-Model container, deploy it to an
+    endpoint, add more models to make them available to a deployed endpoint, list available
+    models and run predictions on a Multi-Model endpoint.
```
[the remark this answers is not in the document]
dlragha: Thanks, that makes sense. Updated the doctring

tests/unit/test_multidatamodel.py
```diff
@@ -0,0 +1,289 @@
...
+
+    assert model.sagemaker_session == sagemaker_session
+    assert model.model_name == MODEL_NAME
+    assert model.model_data_prefix == VALID_MULTI_MODEL_DATA_PREFIX
+    assert model.role == ROLE
+    assert model.image == IMAGE
+    assert model.vpc_config is None
+
+
+def test_multi_data_model_create_with_model_arg(sagemaker_session, mxnet_model):
```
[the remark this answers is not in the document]
dlragha: Sure, added a test to cover this use case

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+            **kwargs: Keyword arguments passed to the ``Model`` initializer.
+        """
+        self.model_name = model_name
+        self._model_data_prefix = model_data_prefix
+        self.model_data_prefix = model_data_prefix
+        self.model = model
+        self.container_mode = MULTI_MODEL_CONTAINER_MODE
+        self.image = image
+        self.role = role
+        self.predictor_cls = predictor_cls
```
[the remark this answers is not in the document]
dlragha: Removed these from init but we need the `model_name` ( now renamed to `name` for consistency) since it is referenced in `deploy()` method

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+            predictor_cls=self.predictor_cls,
+            **kwargs
+        )
+
+        if not self.sagemaker_session:
+            self.sagemaker_session = sagemaker.session.Session()
+        self.s3_client = self.sagemaker_session.boto_session.client("s3")
+
+        # Create the S3 prefix path to ensure model deployment succeeds
+        self._create_s3_model_data_path()
```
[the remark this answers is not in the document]
dlragha: Removed this as part of new commit. I had this in here to ensure that the create model does not fail in case `model_data_prefix` path does not exist yet (i.e. there are no models added to `MultiDataModel` object before `deploy()`. Makes sense to let the error propagate to the user if this occurs. Thanks for the suggestion.

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+            model_data_path = source_model_data_path
+
+        # Construct the destination path
+        dst_url = os.path.join(self.model_data_prefix, model_data_path)
+        destination_bucket, destination_model_data_path = self._parse_s3_uri(dst_url)
+
+        # Copy the model artifact
+        self.s3_client.copy(copy_source, destination_bucket, destination_model_data_path)
+
+    def list_models(self):
```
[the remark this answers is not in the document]
dlragha: Updated code to use `list_s3_files()`

src/sagemaker/multidatamodel.py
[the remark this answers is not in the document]
dlragha: Was able to reuse `parse_s3_url`. Removed `_parse_s3_uri`.

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Returns:
+            dict[str, str]: A complete container definition object usable with the CreateModel API
+        """
+        # Copy the trained model's image and environment variables if they exist.
+        # Models trained with FrameworkEstimator set framework specific environment variables
+        # which need to be copied over
+        if self.model:
+            container_definition = self.model.prepare_container_def(instance_type, accelerator_type)
+            self.image = self.image or container_definition["Image"]
```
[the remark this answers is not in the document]
dlragha: This was handling the case when self.image is None, in which case it would be updated with the value of `container_definition["Image"]`. But I have updated the logic now to use `self.model` params when available and ignore local params per our discussion

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+        return self._model_data_prefix
+
+    @model_data_prefix.setter
+    def model_data_prefix(self, model_data_prefix):
+        """
+        Args:
+            model_data_prefix:
+        """
+        # Validate path
+        if not (model_data_prefix.startswith("s3://") and model_data_prefix.endswith("/")):
```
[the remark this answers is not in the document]
dlragha: Fixed

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,211 @@
...
+        # Create the S3 prefix path to ensure model deployment succeeds
+        self._create_s3_model_data_path()
+
+    @property
+    def model_data_prefix(self):
+        """Placeholder docstring"""
+        return self._model_data_prefix
+
+    @model_data_prefix.setter
+    def model_data_prefix(self, model_data_prefix):
```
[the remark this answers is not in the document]
dlragha: Created the property and setter to validate the input `model_data_prefix` path. This would be triggered when we attempt to assign any value to `self.model_data_prefix`. But I have updated the code to do the validation in `init` to simplify this and avoid confusion.

src/sagemaker/multidatamodel.py
dlragha: Correct, the logic was to ignore and model's values and use the ones in `init` instead. But updated the code to use `self.model` parameters when available

src/sagemaker/multidatamodel.py
dlragha: We needed the private variable as a helper for the validation that happens in the model_data_prefix.setter and self.model_data_prefix to actually set the value (which also triggers the setter and in the process validates the input). But in the new commit, I rewrote the logic to directly do the validation in init to avoid confusion

src/sagemaker/multidatamodel.py
dlragha: We will use `Model` parameters if `model` parameter is not None

src/sagemaker/multidatamodel.py
dlragha: Sure, fixed

src/sagemaker/multidatamodel.py
dlragha: Updated logic to support local files as well. Thanks for the suggestion

src/sagemaker/multidatamodel.py
dlragha: Replaced this with `s3.parse_s3_url`

tests/data/multimodel/container/Dockerfile
dlragha: Fixed

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,308 @@
...
+                ``self.predictor_cls`` on the created endpoint name,
+                if ``self.predictor_cls``
+                is not None. Otherwise, return None.
+        """
+        # Set model specific parameters
+        if self.model:
+            enable_network_isolation = self.model.enable_network_isolation()
+            role = self.model.role
+            vpc_config = self.model.vpc_config
+            predictor = self.model.predictor_cls
```
fm1ch4: could this not simply be:

self.model.deploy()

else do super.deploy() ?

src/sagemaker/multidatamodel.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+        self.name = name
+        self.model_data_prefix = model_data_prefix
+        self.model = model
+        self.container_mode = MULTI_MODEL_CONTAINER_MODE
+        self.sagemaker_session = sagemaker_session or Session()
+        self.s3_client = self.sagemaker_session.boto_session.client("s3")
+
+        # Set the ``Model`` parameters if the model parameter is not specified
+        if not self.model:
```
fm1ch4: I think it would be better to just raise a ValueError if model and other parameters like image, role, etc. are provided, instead of silently ignoring them.

src/sagemaker/multidatamodel.py
dlragha: I tried `self.model.deploy` but realized that since we set the values of `model_data_url` and `container_mode` parameters in `MultiDataModel`'s `prepare_container_def()` method, we'd need to replicate the logic from `Model`'s `deploy()` here, else it ends up creating a container with `single_model` mode and incorrect `model_data_url`

src/sagemaker/multidatamodel.py
dlragha: Sure will add the validation

PR #1168, opened 2019-12-13
feature: allow setting the default bucket in Session

knakad:
*Description of changes:* Session now allows default bucket name to be passed in. Particularly useful in controlled environment where new buckets cannot be created.

*Testing done:* New tests added to validate expected custom bucket 

## Merge Checklist

### DIFF
--- a/tests/integ/test_processing.py
+++ b/tests/integ/test_processing.py
@@ -14,7 +14,10 @@ from __future__ import absolute_import
 
 import os
 
+import boto3
 import pytest
+from botocore.config import Config
+from sagemaker import Session
 from sagemaker.fw_registry import default_framework_uri
 
 from sagemaker.processing import ProcessingInput, ProcessingOutput, ScriptProcessor, Processor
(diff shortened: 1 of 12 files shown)


tests/integ/test_processing.py
```diff
@@ -170,6 +202,90 @@ def test_sklearn_with_customizations(
...
+    sagemaker_session_with_custom_bucket,
+    image_uri,
+    sklearn_full_version,
+    cpu_instance_type,
+    output_kms_key,
+):
+
+    input_file_path = os.path.join(DATA_DIR, "dummy_input.txt")
+
+    sklearn_processor = SKLearnProcessor(
```
andremoeller: IMO: since this functionality lives in the parent `Processor` and propagates down to child classes, the test should use the `Processor` class rather than `SKLearnProcessor`. (If we validate using `Processor`, we validate for all child classes, but if we validate using `SKLearnProcessor`, we don't validate for parent classes.)

knakad: I see it the other way around.
I prefer to use the implementation in this case (SKLearnProcessor), because it allows us to validate the superclass inherited methods (normalization of inputs) AS WELL as the custom implementation logic (automatic code upload).

andremoeller: ^ ah, I realize I should've said `ScriptProcessor`, not `Processor` -- if another child class of `ScriptProcessor` is added, and if the integ tests use `ScriptProcessor`, automatic code upload would be tested for that new class by dint of it being tested with `ScriptProcessor`. (For example: what if `SKLearnProcessor` has some code added to it that lets this test pass just for SKLearnProcessor?)

Issue #1174, opened 2019-12-16
Bug in v1.47.0 after allowing default bucket to be specified by the user


elvinx:
@andremoeller thank you for the quick fix!
I am afraid this feature (https://github.com/aws/sagemaker-python-sdk/pull/1168) introduced a bug :/ And now by default the SDK tries to create the bucket all the time
 - I upgraded to SageMaker 1.47 
 - Run below lines of code
```python
import sagemaker
role = sagemaker.get_execution_role()
```
 - I get the below error 
`ClientError: An error occurred (AccessDenied) when calling the CreateBucket operation: Access Denied`

<img width="865" alt="Screen Shot 2019-12-16 at 2 22 24 PM" src="https://user-images.githubusercontent.com/8008982/70910195-7fdcda80-200f-11ea-83a5-b0045858a377.png">

https://github.com/aws/sagemaker-python-sdk/issues/1165

andremoeller:
@elvinx 

Thanks for reporting this! We're looking into this now.

knakad:
While we push a revert out, I believe a workaround is to specify the session with the custom bucket name as part of the session you pass in to .get_execution_role().

i.e.:
```
my_session = Session(..., default_bucket="my_custom_bucket")
role = sagemaker.get_execution_role(sagemaker_session=my_session)
```

Alternatively, you can also downgrade to a previous version to unblock yourself.

In the meantime, we're working on a version of this fix that doesn't create the bucket on init.

ajaykarpur:
Hi @elvinx, we've merged in the fix for this bug. We'll have a release out soon.

ajaykarpur:
The bugfix has been released. Please feel free to reach out if you have any further questions.

PR #1176, opened 2019-12-16
feature: allow setting the default bucket in Session

knakad:
Default bucket not created on init.

*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/1174

*Description of changes:* Identical to previous change except default bucket is no longer created on init.

*Testing done:* Shored up integ tests to automatically catch this in the future.

## Merge Checklist

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -97,11 +97,16 @@ class Session(object):  # pylint: disable=too-many-public-methods
                 ``InvokeEndpoint`` calls to Amazon SageMaker (default: None). Predictors created
                 using this ``Session`` use this client. If not provided, one will be created using
                 this instance's ``boto_session``.
-            default_bucket (str): The default s3 bucket to be used by this session.
-                Ex: "sagemaker-us-west-2"
+            default_bucket (str): The default Amazon S3 bucket to be used by this session.
+                This will be created the next time an Amazon S3 bucket is needed (by calling
+                :func:`default_bucket`).
+                If not provided, a default bucket will be created based on the following format:
+                "sagemaker-{region}-{aws-account-id}".
+                Example: "sagemaker-my-custom-bucket".
 
         """
         self._default_bucket = None
+        self._default_bucket_name_override = default_bucket
 
         # currently is used for local_code in local mode
         self.config = None
@@ -139,9 +144,6 @@ class Session(object):  # pylint: disable=too-many-public-methods
             )
 
         prepend_user_agent(self.sagemaker_runtime_client)
-
-        self._default_bucket = None
-        self._desired_default_bucket_name = default_bucket
 
         self.local_mode = False
 
@@ -331,9 +333,9 @@ class Session(object):  # pylint: disable=too-many-public-methods
         if self._default_bucket:
             return self._default_bucket
 
-        default_bucket = self._desired_default_bucket_name
         region = self.boto_session.region_name
 
+        default_bucket = self._default_bucket_name_override
         if not default_bucket:
             account = self.boto_session.client(
                 "sts", region_name=region, endpoint_url=sts_regional_endpoint(region)
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -76,7 +76,13 @@ class Session(object):  # pylint: disable=too-many-public-methods
     bucket based on a naming convention which includes the current AWS account ID.
     """
 
-    def __init__(self, boto_session=None, sagemaker_client=None, sagemaker_runtime_client=None):
+    def __init__(
+        self,
+        boto_session=None,
+        sagemaker_client=None,
+        sagemaker_runtime_client=None,
+        default_bucket=None,
+    ):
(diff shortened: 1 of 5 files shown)


src/sagemaker/session.py
```diff
@@ -126,6 +140,9 @@ def _initialize(self, boto_session, sagemaker_client, sagemaker_runtime_client):
 
         prepend_user_agent(self.sagemaker_runtime_client)
 
+        self._default_bucket = None
+        self._desired_default_bucket_name = default_bucket
```
laurenyu: l. 143 is already being done in l. 105. I would just move l. 144 to l. 106. You also then wouldn't have to duplicate the logic in the local mode test object.

and maybe `_default_bucket_name_override`?

src/sagemaker/session.py
```diff
@@ -314,11 +331,14 @@ def default_bucket(self):
...
-        account = self.boto_session.client(
-            "sts", region_name=region, endpoint_url=sts_regional_endpoint(region)
-        ).get_caller_identity()["Account"]
-        default_bucket = "sagemaker-{}-{}".format(region, account)
+
+        if not default_bucket:
+            account = self.boto_session.client(
+                "sts", region_name=region, endpoint_url=sts_regional_endpoint(region)
+            ).get_caller_identity()["Account"]
+            default_bucket = "sagemaker-{}-{}".format(region, account)
```
laurenyu: nit: I'd move l. 334 to be right before the `if` statement because it's clearer that's it's basically just an `if`/`else` there. Or just make it an `if`/`else`:

```python
if self._desired_default_bucket_name
    default_bucket = self._desired_default_bucket_name
else
    ...
```

src/sagemaker/session.py
```diff
@@ -91,15 +97,23 @@ def __init__(self, boto_session=None, sagemaker_client=None, sagemaker_runtime_c
                 ``InvokeEndpoint`` calls to Amazon SageMaker (default: None). Predictors created
                 using this ``Session`` use this client. If not provided, one will be created using
                 this instance's ``boto_session``.
+            default_bucket (str): The default s3 bucket to be used by this session.
+                Ex: "sagemaker-us-west-2"
```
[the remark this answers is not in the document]
knakad: +1. Great feedback!

src/sagemaker/session.py
knakad: +1. Fixed!

src/sagemaker/session.py
knakad: I moved the line closer, but I like to avoid ifelses when possible.
I think it's more readable to have a default case get overridden under specific circumstances. Also guarantees that the variable always gets set.

src/sagemaker/session.py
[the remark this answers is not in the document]
knakad: +1. Great feedback!

src/sagemaker/session.py
```diff
@@ -91,13 +97,25 @@ def __init__(self, boto_session=None, sagemaker_client=None, sagemaker_runtime_c
                 ``InvokeEndpoint`` calls to Amazon SageMaker (default: None). Predictors created
                 using this ``Session`` use this client. If not provided, one will be created using
                 this instance's ``boto_session``.
+            default_bucket (str): The default Amazon S3 bucket to be used by this session.
+                This will be created the next time an Amazon S3 bucket is needed (by calling
+                :func:`default_bucket`).
+                If not provided, a default bucket will be created based on the following format:
+                "sagemaker-{region}-{aws-account-id}".
+                Example: "sagemaker-my-custom-bucket".
```
laurenyu: nit: if you want the "example" block to be formatted, you need a newline before it. (if you were going for having it in the same paragraph, then ignore this comment.)

src/sagemaker/session.py
```diff
@@ -91,13 +97,25 @@ def __init__(self, boto_session=None, sagemaker_client=None, sagemaker_runtime_c
...
 
         # currently is used for local_code in local mode
         self.config = None
 
-        self._initialize(boto_session, sagemaker_client, sagemaker_runtime_client)
+        self._initialize(
+            boto_session=boto_session,
+            sagemaker_client=sagemaker_client,
+            sagemaker_runtime_client=sagemaker_runtime_client,
+        )
```
laurenyu: guessing `black-format` will put this back in one line?

knakad: Don't think so? Based on my git hooks, I wouldn't have been able to commit it if black-check didn't return 1.
Unit tests also passed.
Am I misunderstanding?

src/sagemaker/session.py
knakad: Gotcha. I actually want this example to match default_bucket's param precisely.

src/sagemaker/session.py
laurenyu: never mind. it looked strange to me that there should be any diff remaining in this part of the code, and my understanding was that black had a strict threshold for when it moved to multiple lines (for these kinds of cases)

PR #1218, opened 2020-01-08
fix: add default framework version warning message in Model classes

ChuyangDeng:
*Issue #, if available:*
Framework Model classes do not log warning when they are defaulted to default framework versions (not the latest versions) as Estimator classes do.
 
*Description of changes:*
Default ``framework_version`` to ``None`` in Model classes, emit warning message and then assign default values.

*Testing done:*
Unit tests in each modified framework unit tests:
Chainer, MXNet, PyTorch, TensorFlow

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/chainer/model.py
+++ b/src/sagemaker/chainer/model.py
@@ -16,7 +16,8 @@ from __future__ import absolute_import
 import logging
 
 import sagemaker
-from sagemaker.fw_utils import create_image_uri, model_code_key_prefix, python_deprecation_warning
+from sagemaker.fw_utils import create_image_uri, model_code_key_prefix, python_deprecation_warning, \
+    empty_framework_version_warning
 from sagemaker.model import FrameworkModel, MODEL_SERVER_WORKERS_PARAM_NAME
 from sagemaker.chainer.defaults import CHAINER_VERSION
 from sagemaker.predictor import RealTimePredictor, npy_serializer, numpy_deserializer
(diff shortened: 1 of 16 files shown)


src/sagemaker/chainer/model.py
```diff
@@ -54,14 +59,17 @@ class ChainerModel(FrameworkModel):
 
     __framework_name__ = "chainer"
 
+    LATEST_VERSION = "5.0.0"
```
laurenyu: should we instead read this from the estimator so that we don't have to update multiple constants?

ChuyangDeng: Yes. Duplicating the constant increases future efforts of maintaining them. 

However, Model class does not import corresponding estimator and LATEST_VERSION is an instance variable of an estimator object.  We need to do:

`import sagemaker.chainer.estimator as estimator`
and then
`estimator.LATEST_VERSION`

Model classes currently do not import corresponding Estimator classes, I was wondering if such imports worth it for "LATEST_VERSION"

laurenyu: hmm...maybe it is worth moving it to `defaults.py` then even though it's not a default.

ChuyangDeng: Can we add a LATEST_VERSION in ``defaults.py`` but still keep the one in ``Estimator`` class? so that customer code which uses estimator.LATEST_VERSION won't break

laurenyu: so you'll still have to set it in estimator, but you can do something like

```python
from sagemaker.<framework> import defaults

class <Framework>:
    LATEST_VERSION = defaults.LATEST_VERSION
```

PR #1273, opened 2020-01-30
fix: account for EI and version-based ECR repo naming in serving_image_uri()

laurenyu:
*Issue #, if available:*

*Description of changes:*
* #1247 didn't account for the fact that for each framework version with a "framework-serving" ECR repo, earlier versions of the framework didn't have the suffix.
* In addition, the `serving_image_uri()` method that was originally added in #1245 didn't originally have an option for EI because we don't have parity for Airflow yet, but there was the opportunity to remove duplicated code by reusing the new method in each model class, so I added the accelerator option for that.
* Lastly, I changed the Airflow unit test that was changed in #1247 to use the latest MXNet image URI format rather than change it back to the old one.

*Testing done:*
* `tox -e pylint,flake8,black-format`
* unit tests

## Merge Checklist

#### General

#### Tests


src/sagemaker/pytorch/model.py
```diff
@@ -137,34 +136,21 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
                 For example, 'ml.p2.xlarge'.
             accelerator_type (str): The Elastic Inference accelerator type to
                 deploy to the instance for loading and making inferences to the
-                model. For example, 'ml.eia1.medium'.
+                model. Currently unsupported with PyTorch.
```
knakad: Should we instead remove this param entirely? Alternatively, add a warning to spew this out if it's passed in?

src/sagemaker/pytorch/model.py
```diff
@@ -137,34 +136,21 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
-
-            deploy_image = create_image_uri(
-                region_name,
-                framework_name,
-                instance_type,
-                self.framework_version,
-                self.py_version,
-                accelerator_type=accelerator_type,
+            deploy_image = self.serving_image_uri(
+                region_name, instance_type, accelerator_type=accelerator_type
```
knakad: [Minor] Keyword/Named arguments make kittens happy.

src/sagemaker/pytorch/model.py
```diff
@@ -174,22 +160,41 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
+    def serving_image_uri(self, region_name, instance_type, accelerator_type=None):
         """Create a URI for the serving image.
 
         Args:
             region_name (str): AWS region where the image is uploaded.
             instance_type (str): SageMaker instance type. Used to determine device type
                 (cpu/gpu/family-specific optimized).
+            accelerator_type (str): The Elastic Inference accelerator type to
+                deploy to the instance for loading and making inferences to the
+                model. Currently unsupported with PyTorch.
```
knakad: Same comment as above. What happens if this is passed in? Is the img not found?

src/sagemaker/tensorflow/model.py
```diff
@@ -146,13 +144,8 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
             region_name = self.sagemaker_session.boto_region_name
-            deploy_image = create_image_uri(
-                region_name,
-                self.__framework_name__,
-                instance_type,
-                self.framework_version,
-                self.py_version,
-                accelerator_type=accelerator_type,
+            deploy_image = self.serving_image_uri(
+                region_name, instance_type, accelerator_type=accelerator_type
```
knakad: [Minor] Keyword/Named arguments make kittens happy.

src/sagemaker/tensorflow/model.py
```diff
@@ -165,22 +158,26 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
 
         Returns:
             str: The appropriate image URI based on the given parameters.
 
         """
-        return fw_utils.create_image_uri(
+        return create_image_uri(
             region_name,
-            "-".join([self.__framework_name__, "serving"]),
+            self.__framework_name__,
```
knakad: Can you clarify why tensorflow-serving isn't used here?

src/sagemaker/mxnet/model.py
```diff
@@ -175,22 +157,41 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
+                model (default: None). For example, 'ml.eia1.medium'.
 
         Returns:
             str: The appropriate image URI based on the given parameters.
 
         """
-        return fw_utils.create_image_uri(
+        framework_name = self.__framework_name__
+        if self._is_mms_version():
+            framework_name += "-serving"
```
ajaykarpur: String join is more efficient than string concatenation. For more details, see:
- https://stackoverflow.com/questions/3055477/how-slow-is-pythons-string-concatenation-vs-str-join
- https://waymoot.org/home/python_string/

```suggestion
            framework_name = "-".join([framework_name, "serving"])
```

src/sagemaker/pytorch/model.py
```diff
@@ -174,22 +160,41 @@ def prepare_container_def(self, instance_type, accelerator_type=None):
...
 
         Returns:
             str: The appropriate image URI based on the given parameters.
 
         """
-        return fw_utils.create_image_uri(
+        framework_name = self.__framework_name__
+        if self._is_mms_version():
+            framework_name += "-serving"
```
ajaykarpur: String join is more efficient than string concatenation. For more details, see:
- https://stackoverflow.com/questions/3055477/how-slow-is-pythons-string-concatenation-vs-str-join
- https://waymoot.org/home/python_string/

```suggestion
            framework_name = "-".join([framework_name, "serving"])
```

src/sagemaker/pytorch/model.py
laurenyu: a warning does get printed out, but the logic is in a different spot

code: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/fw_utils.py#L56, https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/fw_utils.py#L273-L277
unit test: https://github.com/aws/sagemaker-python-sdk/blob/master/tests/unit/test_pytorch.py#L346-L352

src/sagemaker/mxnet/model.py
laurenyu: talked offline - we'll go with format strings as a compromise between performance and readability, since the performance gain is negligible in this case

an aside - fwiw, a tuple would be more efficient than a list because it's immutable

src/sagemaker/pytorch/model.py
laurenyu: talked offline - I don't see a significant benefit of doing `region_name=region_name` over `region_name`

src/sagemaker/tensorflow/model.py
laurenyu: this class is used only for the legacy TF images - those images supported both training and serving, and were simply named "sagemaker-tensorflow." The sagemaker-tensorflow-serving images are handled by the `sagemaker.tensorflow.serving.Model` class.

Issue #1283, opened 2020-02-06
ListBucket Permission


leezu:
**Describe the bug**
With `model = MXNetModel('s3://lausen-public/bert_sst.tar.gz',`, your package attempts to list the lausen-public bucket and fails due to lack of permission. Note that while I don't grant you permission to list my bucket, `'s3://lausen-public/bert_sst.tar.gz` is public.

**Expected behavior**
No need to list bucket. Just take the public file.

**Screenshots or logs**
https://cbx2jz1ir3.execute-api.us-west-2.amazonaws.com/Prod/buildlogs?key=fa40d2ee-eb3d-48c4-b361-075697d5d814%2Fbuild.log

**System information**
CI run of https://github.com/aws/sagemaker-mxnet-serving-container/pull/94

laurenyu:
thanks for bringing this to our attention!

the reason we use a `ListObjects` call is to [determine whether or not the S3 path is a directory](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L347-L353). Since both folders and single objects are allowed, the code has to be able to handle both. Unfortunately, S3 doesn't provide a clean way to [check if an object is a folder](https://stackoverflow.com/questions/22312833/how-can-you-tell-if-an-object-is-a-folder-on-aws-s3), so that's where the `ListObjects` call comes into play.

I do think we can flip the logic around to make this work, and I'll try to post an update here when I can

laurenyu:
https://github.com/aws/sagemaker-python-sdk/pull/1295

laurenyu:
fix was released in [v1.50.13](https://github.com/aws/sagemaker-python-sdk/releases/tag/v1.50.13)

PR #1295, opened 2020-02-18
fix: allow download_folder to download file even if bucket is more restricted

laurenyu:
*Issue #, if available:*
#1283

*Description of changes:*
S3 doesn't provide a way of definitively determining if a prefix is a folder, so originally we used `ListObjects` to try and gauge that. However, `ListObjects` requires extra permissions on the bucket, and so download a public file from a private bucket fails. In this PR, I've changed the logic to try and download the file first, and then if that incurs a 404 because it's a folder (no idea why that's how S3 chooses to implement this...), then the code will proceed with as if the prefix points to a folder.

Side note - I noticed that pylint was requiring me to put `botocore` in the wrong order for the imports. I'll fix that in a separate PR.

*Testing done:*
tried the example S3 path from #1283:

```python
>>> import sagemaker
>>> from sagemaker import utils
>>> utils.download_folder('lausen-public', 'bert_sst.tar.gz', '/tmp', sagemaker.session.Session())
```

```bash
$ ls /tmp/bert_sst.tar.gz
/tmp/bert_sst.tar.gz
```

(also ran `tox tests/unit`)

## Merge Checklist

#### General

- [x] I have used the regional endpoint when creating S3 and/or STS clients (if appropriate)

#### Tests

### DIFF
--- a/src/sagemaker/utils.py
+++ b/src/sagemaker/utils.py
@@ -23,13 +23,12 @@ import sys
 import tarfile
 import tempfile
 import time
-
-
 from datetime import datetime
 from functools import wraps
 
 import six
 from six.moves.urllib import parse
+import botocore
 
 
 ECR_URI_PATTERN = r"^(\d+)(\.)dkr(\.)ecr(\.)(.+)(\.)(amazonaws.com|c2s.ic.gov)(/)(.*:.*)$"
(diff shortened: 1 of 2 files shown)


src/sagemaker/utils.py
```diff
@@ -338,21 +337,25 @@ def download_folder(bucket_name, prefix, target, sagemaker_session):
...
+    except botocore.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] == "404" and e.response["Error"]["Message"] == "Not Found":
+            # S3 also throws this error if the object is a folder,
+            # so assume that is the case here, and then raise for an actual 404 later.
+            pass
+        else:
+            raise
 
-    # the prefix points to an s3 'directory' download the whole thing
+    # Assume the prefix points to an S3 'directory' and download the whole thing
```
nadiaya: Suggestion: Could moving it into a function and calling it from inside the except when object is not a file make it better?

PR #1345, opened 2020-03-10
infra: generate test job name at test start instead of module start

knakad:
*Description of changes:* This eliminates a race condition that results in two different tests
attempting to use the same job name.

*Testing done:* IT

## Merge Checklist

#### General

#### Tests

- [XXXXXXX] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)


tests/integ/test_auto_ml.py
```diff
@@ -116,14 +115,15 @@ def test_auto_ml_fit_optional_args(sagemaker_session):
         output_path=output_path,
         problem_type=problem_type,
         job_objective=job_objective,
+        base_job_name=BASE_JOB_NAME,
     )
     inputs = TRAINING_DATA
     with timeout(minutes=AUTO_ML_DEFAULT_TIMEMOUT_MINUTES):
-        auto_ml.fit(inputs, job_name=JOB_NAME)
+        auto_ml.fit(inputs)
```
laurenyu: let's use `unique_name_from_base` here instead.

knakad: Good callout. Fixed!

tests/integ/test_auto_ml.py
```diff
@@ -119,11 +118,11 @@ def test_auto_ml_fit_optional_args(sagemaker_session):
...
     inputs = TRAINING_DATA
     with timeout(minutes=AUTO_ML_DEFAULT_TIMEMOUT_MINUTES):
-        auto_ml.fit(inputs, job_name=JOB_NAME)
+        auto_ml.fit(inputs, job_name=unique_name_from_base(BASE_JOB_NAME))
 
-    auto_ml_desc = auto_ml.describe_auto_ml_job(job_name=JOB_NAME)
+    auto_ml_desc = auto_ml.describe_auto_ml_job(job_name=auto_ml.latest_auto_ml_job.job_name)
     assert auto_ml_desc["AutoMLJobStatus"] == "Completed"
-    assert auto_ml_desc["AutoMLJobName"] == JOB_NAME
+    assert auto_ml_desc["AutoMLJobName"] == auto_ml.latest_auto_ml_job.job_name
```
laurenyu: nit: I feel like it would make more sense to save `job_name` separately before calling `fit`, and then asserting that the job name we get later equals that. don't feel strongly about it, though, because I can't clearly articulate why that's my instinct here.

knakad: I thought of that and chose this version purposefully.
If .fit() ever mutates the job name for whatever reason, the source of truth would be auto_ml.latest_job.name . Having this reference come from the object ensures it's closer to the source.
I think this is more reliable for the future, even though realistically, there's no difference.

laurenyu: I think my argument would be that if we were to change the logic around how `fit()` handles a specified job name, we would want our tests that are expecting otherwise to break. Otherwise we're just verifying the job name without a strong expectation of what the job name should be.

To be fair, though, I also don't think the integ tests are the right place for this kind of verification now that I think about it - it's much more of a "business logic to be tested by unit tests" kind of thing 😂

knakad: I can get behind both of those arguments =)

Issue #1358, opened 2020-03-14
Support using local S3 instance (e.g. minio)


larsll:
**Is your feature request related to a problem? Please describe.**
We are working on setting up local/offline DeepRacer training; using Sagemaker and Robomaker. This means having a local S3 instance (using minio) to host the files.

**Describe the solution you'd like**
It should be possible to configure the S3 Endpoint URL to point to a specific local URL/port. Current proposal is to allow to pass a pre-configured s3Client into an Estimator on creation.

**Describe alternatives you've considered**
An alternative would be to store the required configuration in the Sagemaker Local Session. It would nevertheless require changes to all areas of code where a new s3Client is created.

laurenyu:
Thanks again for submitting #1359! Did it address your issue (or is there more you're looking for with this)?

larsll:
Yes - this can be closed. #1359 solved this issue -- we no longer need a fork to support local S3 instances (e.g. minio).

PR #1359, opened 2020-03-14
feature: Allow setting S3 endpoint URL for Local Session

larsll:
*Issue #, if available:* #1358 

*Description of changes:* Makes it possible to pass in a different S3 endpoint URL when instantiating a Local Session. This enables local jobs to connect to a locally hosted endpoint, e.g. Minio, in offline settings, or when it is not desirable to transfer large amounts of data into the cloud. 

*Testing done:*

## Merge Checklist

#### General

#### Tests


ChuyangDeng:
Thanks for the contribution! Could you add unit tests for your changes and preferably one inte test as well?

The unit test can be added in to an existing framework training integ test file and using the local S3 url

src/sagemaker/estimator.py
```diff
@@ -1416,6 +1416,7 @@ def __init__(
         checkpoint_s3_uri=None,
         checkpoint_local_path=None,
         enable_sagemaker_metrics=None,
+        s3_client=None,
```
laurenyu: given that one can already pass in a sagemaker session to an estimator, it seems like it would be more appropriate to configure a sagemaker session with the desired s3 client rather than introducing a slightly overlapping parameter here.

larsll: @laurenyu - thank you for the feedback. I will check if this is feasible!

larsll: @laurenyu - I have refactored now to allow to pass in an s3_endpoint_url; this then creates the s3_resource and s3_client objects. Wherever these are used it will check if one is available, or create one. All parts of the SDK should now benefit from it. The feature is only available in local mode.

larsll:
> Thanks for the contribution! Could you add unit tests for your changes and preferably one inte test as well?
> 
> The unit test can be added in to an existing framework training integ test file and using the local S3 url

@ChuyangDeng - I added one Unit test to ensure that the passing of the variable is passed through to an object. Regarding integration tests; what is the correct approach there? The change allows for a local S3 instance (e.g. powered by Minio) -- would this require changes to the integration environment setup?

src/sagemaker/fw_utils.py
```diff
@@ -354,7 +363,12 @@ def tar_and_upload_dir(
         else:
             extra_args = None
 
-        session.resource("s3").Object(bucket, key).upload_file(tar_file, ExtraArgs=extra_args)
+        if s3_resource is None:
+            s3_resource = session.resource("s3")
```
laurenyu: I know this wasn't your code originally, but can you supply the `region_name` arg?

src/sagemaker/estimator.py
```diff
@@ -1713,6 +1713,7 @@ def _stage_user_code_in_s3(self):
             directory=self.source_dir,
             dependencies=self.dependencies,
             kms_key=kms_key,
+            s3_resource=self.sagemaker_session.s3_resource,
```
laurenyu: I'm still unconvinced of this. what if someone passes in both a `sagemaker_session` and an `s3_resource`? I think it would be better to let the user just pass in a `sagemaker_session` with whatever configuration (i.e. `s3_resource`) they want already defined.

src/sagemaker/multidatamodel.py
```diff
@@ -92,7 +92,11 @@ def __init__(
         self.model = model
         self.container_mode = MULTI_MODEL_CONTAINER_MODE
         self.sagemaker_session = sagemaker_session or Session()
-        self.s3_client = self.sagemaker_session.boto_session.client("s3")
+
+        if self.sagemaker_session.s3_client is None:
+            self.s3_client = self.sagemaker_session.boto_session.client("s3")
```
laurenyu: same thing here and below - can you add `region_name`?

src/sagemaker/estimator.py
larsll: This function call is wierd; one passes in a boto session (and then we added the s3_resource); we could pass in a sagemaker session instead, and then get out both the boto session and the s3_resource. It would break the API call though.

laurenyu: ah, I see. my bad, I was thinking this was in a different part of the code. this is fine, then.

src/sagemaker/fw_utils.py
```diff
@@ -331,6 +338,8 @@ def tar_and_upload_dir(
             copied into /opt/ml/lib
         kms_key (str): Optional. KMS key ID used to upload objects to the bucket
             (default: None).
+        s3_resource (boto3.resource('s3')): Optional. Pre-instantiated Boto3 Resource
+            for S3 connections, can be used to set e.g. the endpoint URL (default: None).
```
laurenyu:
```suggestion
        s3_resource (boto3.resource("s3")): Optional. Pre-instantiated Boto3 Resource
            for S3 connections, can be used to customize the configuration, e.g. set the endpoint URL (default: None).
```

src/sagemaker/local/image.py
```diff
@@ -139,11 +140,18 @@ def train(self, input_data_config, output_data_config, hyperparameters, job_name
...
+            training_env_vars[
+                S3_ENDPOINT_URL_ENV_NAME
+            ] = self.sagemaker_session.s3_resource.meta.client._endpoint.host
+
         compose_data = self._generate_compose_file(
             "train", additional_volumes=volumes, additional_env_vars=training_env_vars
         )
         compose_command = self._compose()
 
+        logger.info("Trying to launch image: %s", str(self.image))
```
laurenyu: leftover from debugging?

(same below)

src/sagemaker/local/local_session.py
```diff
@@ -377,7 +378,9 @@ def invoke_endpoint(
 class LocalSession(Session):
     """Placeholder docstring"""
 
-    def __init__(self, boto_session=None):
+    def __init__(self, boto_session=None, s3_endpoint_url=None):
```
laurenyu: it seems a little strange to me to modify this constructor but not `Session`'s

src/sagemaker/session.py
```diff
@@ -199,7 +199,10 @@ def upload_data(self, path, bucket=None, key_prefix="data", extra_args=None):
             key_suffix = name
 
         bucket = bucket or self.default_bucket()
-        s3 = self.boto_session.resource("s3")
+        if self.s3_resource is None:
+            s3 = self.boto_session.resource("s3")
```
laurenyu: could you include the region here as well? (same below)

tests/unit/test_local_session.py
```diff
@@ -475,3 +476,58 @@ def test_local_session_is_set_to_local_mode():
     boto_session = Mock(region_name="us-west-2")
     local_session = sagemaker.local.local_session.LocalSession(boto_session=boto_session)
     assert local_session.local_mode
+
+
+ENDPOINT_URL = "http://127.0.0.1:9000"
+BUCKET_NAME = "mybucket"
+LS_FILES = {"Contents": [{"Key": "/data/test.csv"}]}
```
laurenyu: let's put these constants up top with the other ones

src/sagemaker/local/local_session.py
larsll: The idea is that you would only do this in LocalSession; if you are running 'real' Sagemaker then you would use 'real' S3.

laurenyu: I see. I guess that's fine then. Please update the PR title to make this clear.

src/sagemaker/local/local_session.py
```diff
@@ -98,6 +98,7 @@ def create_training_job(
         )
         training_job = _LocalTrainingJob(container)
         hyperparameters = kwargs["HyperParameters"] if "HyperParameters" in kwargs else {}
+        logger.info("Starting training job")
```
laurenyu: another leftover debugging statement?

larsll: This was more to actually output that things are starting. I don't know what the guidance on logging itself is. Happy to remove it.

laurenyu: I think I'm okay with this one then. iirc training on SageMaker prints a similar message

PR #1373, opened 2020-03-24
feature: inferentia support

ChuyangDeng:
*Issue #, if available:*

*Description of changes:*
Support Inferentia in Python SDK
- add inferentia target_instance_family validation
- add logic to create inferentia image uri
- check for regions and throw error
- check for framework/versions and throw error
- add warning for compiled model
- integ test

*Testing done:*
unit test
beta test, ran beta integ test
integ test (skip until containers are available prod)

## Merge Checklist

#### General

#### Tests


src/sagemaker/fw_utils.py
```diff
@@ -93,6 +93,13 @@
     "pytorch-serving-eia": [1, 3, 1],
 }
 
+INFERENTIA_VERSION_RANGES = {
```
yuanzhua: Why do we need to introduce the unbound and is_version_equal_or_lower? DLCs only have the lowest version. should we be consistent here? Is it because Neo/Inferentia doesn't support the latest version? EIA doesn't support the latest version either. I am okay with introducing a new map here, but maybe we should move EI's validation into the new map as well? downside of this approach is we have to keep updating the unbound every time Neo/EI supports a new version. We should put comments here about "Once Neo/EI has catched up, move them to the MERGED_FRAMEWORKS_LOWEST_VERSIONS map". What do you think?

ChuyangDeng: The reason why I introduce this upper limit is that Inferentia does not support latest versions and inferentia containers does not align with DLC versions very much.

Now we have tensorflow up to 2.1, but inferentia is 1.15.0, 3 versions behind (1.15.2, 2.0, 2.1). And for mxnet, DLC skipped 1.5.0 but inferentia only has 1.5.0. Plus inferentia is neo containers and follow neo image naming schema, so I think it’s better to have its own logic.

Also inferentia is neo containers, we should not move them to MERGRD_FRAMEWORKS_LOWEAST_VERSIONS because they are not merged dlc containers.

src/sagemaker/fw_utils.py
```diff
@@ -229,11 +270,25 @@ def create_image_uri(
...
+    if device_type == "inf":
+        if region not in INFERENTIA_SUPPORTED_REGIONS:
+            raise ValueError("Inferentia is not supported in region {}.".format(region))
+        if framework not in INFERENTIA_VERSION_RANGES:
+            raise ValueError("Inferentia does not support {}.".format(framework.split("-")[-1]))
+        if not _is_inferentia_supported(framework, framework_version):
+            raise ValueError(
+                "Inferentia is not supported with {} version {}.".format(
+                    framework.split("-")[-1], framework_version
+                )
```
ajaykarpur: It might be helpful to guide users on which regions, frameworks, and versions are supported in the error messages. For example, "Inferentia is not supported in region {}. The supported regions are {}." or something along those lines.

src/sagemaker/model.py
```diff
@@ -437,6 +464,11 @@ def deploy(
         if self.role is None:
             raise ValueError("Role can not be null for deploying a model")
 
+        if instance_type.startswith("ml.inf") and not self._is_compiled_model:
+            LOGGER.warning(
+                "Your model is not compiled, please compile your model before using Inferentia."
```
ajaykarpur:
```suggestion
                "Your model is not compiled. Please compile your model before using Inferentia."
```

PR #1381, opened 2020-03-31
infra: modify PR template and refactor test

knakad:
*Description of changes:* This change modifies the PR template to be slightly clearer and refactors an integ test.

*Testing done:* UTs/IT.

## Merge Checklist

#### General

- [X] I have passed the region in to any/all clients that I've initialized as part of this change.

#### Tests

- [X] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)

### DIFF
--- a/tests/integ/test_session.py
+++ b/tests/integ/test_session.py
@@ -24,9 +24,13 @@ CUSTOM_BUCKET_NAME = "this-bucket-should-not-exist"
 def test_sagemaker_session_does_not_create_bucket_on_init(
     sagemaker_client_config, sagemaker_runtime_config, boto_config
 ):
-    boto_session = (
-        boto3.Session(**boto_config) if boto_config else boto3.Session(region_name=DEFAULT_REGION)
-    )
+    if boto_config:
+        boto_session = boto3.Session(**boto_config)
+        s3 = boto3.resource("s3", region_name=boto_config["region_name"])
+    else:
+        boto_session = boto3.Session(region_name=DEFAULT_REGION)
+        s3 = boto3.resource("s3", region_name=DEFAULT_REGION)
+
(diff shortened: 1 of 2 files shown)


tests/integ/test_session.py
```diff
@@ -24,9 +24,13 @@
...
 ):
-    boto_session = (
-        boto3.Session(**boto_config) if boto_config else boto3.Session(region_name=DEFAULT_REGION)
-    )
+    if boto_config:
+        boto_session = boto3.Session(**boto_config)
+        s3 = boto3.resource("s3", region_name=boto_config["region_name"])
+    else:
+        boto_session = boto3.Session(region_name=DEFAULT_REGION)
+        s3 = boto3.resource("s3", region_name=DEFAULT_REGION)
```
laurenyu: I think I'd rather go with how the other tests do it and not have a fallback at all. That way we explicitly fail if the region is not set instead of potentially masking other errors (e.g. #1308)

knakad: I see what you're saying, and I think it's a better idea, but that would likely be best done as a refactor to every situation we currently have, because that's not how we currently do it.
I'd also say we should throw an exception in that case to make sure UTs/ITs catch it: https://github.com/aws/sagemaker-python-sdk/blob/d9f4dd162620814407604561927e04c37bfde54a/tests/conftest.py#L93

laurenyu: > a refactor to every situation we currently have
> I'd also say we should throw an exception in that case to make sure UTs/ITs catch it

so...these sound like good suggestions...

laurenyu:
refactor was addressed in #1407

PR #1434, opened 2020-04-22
fix: fix xgboost image incorrect latest version warning

balajitummala:
*Issue #, if available:*

*Description of changes:*
Incorrect xgboost latest version is being shown in the warning. Modified the warning to show the latest version 

*Testing done:*
tox -e py36 tests/unit/test_amazon_estimator.py
tox -e py27 tests/unit/test_amazon_estimator.py

## Merge Checklist

#### General

- [ ] I have passed the region in to any/all clients that I've initialized as part of this change.

#### Tests


src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -622,13 +633,25 @@ def get_image_uri(region_name, repo_name, repo_version=1):
...
+
+def _xgboost_version_compare(repo_version):
+    """Compare xgboost image version with latest version
+
+    Args:
+        repo_version:
+    """
+    if repo_version == 1:
+        return True
+    return _split(XGBOOST_LATEST_VERSION) > _split(repo_version)
```
laurenyu, line 648: this works for now, but it'd be better to convert the numbers to ints because of this kind of case:

```python
>>> "10" < "9"
True
>>> 10 < 9
False
```

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -622,13 +633,25 @@ def get_image_uri(region_name, repo_name, repo_version=1):
...
-            "'%s'. For example:\n"
-            "\tget_image_uri(region, 'xgboost', '%s').",
-            XGBOOST_VERSION_1,
-            XGBOOST_VERSION_1,
-        )
     repo = "{}:{}".format(repo_name, repo_version)
     return "{}/{}".format(registry(region_name, repo_name), repo)
+
+
+def _xgboost_version_compare(repo_version):
```
laurenyu, line 640: let's rename this to something that's a little more descriptive about what the return boolean actually means. how about `_is_not_latest_xgboost_version`? (or `_is_latest_xgboost_version` and then using a `not` for the check above?)

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -622,13 +633,25 @@ def get_image_uri(region_name, repo_name, repo_version=1):
...
+    """Compare xgboost image version with latest version
+
+    Args:
+        repo_version:
+    """
+    if repo_version == 1:
```
aws-patlin, line 646: `"latest"` is another tag that points to the legacy XGBoost image which we should check for.

src/sagemaker/amazon/amazon_estimator.py
aws-patlin: This comparison won't work if we have chars in the version, i.e. "1.x-1". It may be simpler to just do a direct string comparison between `XGBOOST_LATEST_VERSION` and the provided `repo_version`.

src/sagemaker/amazon/amazon_estimator.py
aws-patlin: It's easier to simply check if the version matches the latest version, so I think having this function as `_is_latest_xgboost_version` makes more sense. Then we can check `if not _is_latest_xgboost_version(repo_version)` above.

src/sagemaker/amazon/amazon_estimator.py
balajitummala: @laurenyu Versions might have characters in them so comparing the tuples instead of ints.

src/sagemaker/amazon/amazon_estimator.py
balajitummala: Added the "latest" check

balajitummala:
@aws-patlin Added couple of test cases to verify the version comparison

laurenyu:
the formatting check failed - please run `tox -e black-format` and then commit the changes

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -622,13 +637,25 @@ def get_image_uri(region_name, repo_name, repo_version=1):
...
+
+def _is_latest_xgboost_version(repo_version):
+    """Compare xgboost image version with latest version
+
+    Args:
+        repo_version:
+    """
+    if repo_version == 1 or repo_version == 'latest':
+        return True
+    return _split(XGBOOST_LATEST_VERSION) <= _split(repo_version)
```
aws-patlin, line 652: I forgot about this earlier, but we should also consider `{XGBOOST_LATEST_VERSION}-cpu-py3` as valid for the latest version. Let's make sure this value for `repo_version` returns true as well.

Also, let's use `==` instead of `<=`. `<=` would allow invalid versions to return `True` if they are "larger" than `XGBOOST_LATEST_VERSION`, which doesn't match what we're trying to do here.

src/sagemaker/amazon/amazon_estimator.py
laurenyu: @balajitummala right, forgot about that. that still doesn't address the fact that the string comparison with the tuples still has the potential to be inaccurate. I think @aws-patlin's suggestion of doing a direct string comparison is the right approach here:

```python
if repo_version in (1, "latest"):
    return True
return repo_version.startswith(XGBOOST_LATEST_VERSION)
```

src/sagemaker/amazon/amazon_estimator.py
```diff
@@ -622,13 +637,25 @@ def get_image_uri(region_name, repo_name, repo_version=1):
...
+
+
+def _is_latest_xgboost_version(repo_version):
+    """Compare xgboost image version with latest version
+
+    Args:
+        repo_version:
+    """
+    if repo_version == 1 or repo_version == 'latest':
+        return True
```
aws-patlin, line 651: We'll want to return False here

src/sagemaker/amazon/amazon_estimator.py
balajitummala: @laurenyu Modified comparison to just check if exists in the latest versions list

src/sagemaker/xgboost/defaults.py
```diff
@@ -17,3 +17,4 @@
 XGBOOST_VERSION_1 = "0.90-1"
 XGBOOST_LATEST_VERSION = "0.90-2"
 XGBOOST_SUPPORTED_VERSIONS = [XGBOOST_VERSION_1, XGBOOST_LATEST_VERSION]
+XGBOOST_LATEST_VERSION_EQUIVALENTS = [XGBOOST_LATEST_VERSION, XGBOOST_LATEST_VERSION + "-cpu-py3"]
```
aws-patlin, line 20: I like this idea. Since we need to check against `(version, version + "-cpu-py3")` for all versions, it may be better for us to have a helper function that returns the equivalents instead of a constant specifically for the latest version.

balajitummala: Added a helper to generate the equivalent versions.

laurenyu:
thanks for the contribution!

Issue #1354, opened 2020-03-12
Local deploy failure for MXNet estimator due to non-uniform output uri


ehsanmok:
**Describe the bug**

This is related to the already resolved issue #1349 . After using the local mode for training via `LocalSession()`, deploying the estimator locally repeatedly throws this error:

<details>
<summary> Click to see the error </summary>

```
algo-1-zu1b2_1  | 2020-03-13 17:36:32,149 [WARN ] W-9001-model-stderr com.amazonaws.ml.mms.wlm.WorkerLifeCycle -     in DEFAULT_MODEL_FILENAMES.items()]))
algo-1-zu1b2_1  | 2020-03-13 17:36:32,149 [WARN ] W-9001-model-stderr com.amazonaws.ml.mms.wlm.WorkerLifeCycle - ValueError: Failed to load model with default model_fn: missing file model-symbol.json.Expected files: ['model-symbol.json', 'model-0000.params', 'model-shapes.json']
algo-1-zu1b2_1  | 2020-03-13 17:36:32,149 [WARN ] W-9001-model com.amazonaws.ml.mms.wlm.BatchAggregator - Load model failed: model, error: Worker died.
```

I investigated where the issue might be and manually created the `MXNetModel` that worked with non-local mode but fails in local mode with

```python
---------------------------------------------------------------------------
ClientError                               Traceback (most recent call last)
<timed exec> in <module>()

~/anaconda3/envs/mxnet_p36/lib/python3.6/site-packages/sagemaker/model.py in deploy(self, initial_instance_count, instance_type, accelerator_type, endpoint_name, update_endpoint, tags, kms_key, wait, data_capture_config)
    440             self.name = "{}{}".format(name_prefix, compiled_model_suffix)
    441 
--> 442         self._create_sagemaker_model(instance_type, accelerator_type, tags)
    443         production_variant = sagemaker.production_variant(
    444             self.name, instance_type, initial_instance_count, accelerator_type=accelerator_type

~/anaconda3/envs/mxnet_p36/lib/python3.6/site-packages/sagemaker/model.py in _create_sagemaker_model(self, instance_type, accelerator_type, tags)
    175                 /api/latest/reference/services/sagemaker.html#SageMaker.Client.add_tags
    176         """
--> 177         container_def = self.prepare_container_def(instance_type, accelerator_type=accelerator_type)
    178         self.name = self.name or utils.name_from_image(container_def["Image"])
[...]

PR #1439, opened 2020-04-24
fix: allow output_path without trailing slash in Local Mode training jobs

laurenyu:
*Issue #, if available:*
#1354

*Description of changes:*
Local Mode's logic for constructing the final output path assumes that the estimator's `output_path` ended with a slash ("/"), which deviates from SageMaker's behavior (see #1354).

*Testing done:*
* unit tests
* Chainer integ test (see diff)

## Merge Checklist

#### General

- [ ] I have passed the region in to any/all clients that I've initialized as part of this change.

#### Tests


src/sagemaker/local/utils.py
```diff
@@ -72,6 +72,22 @@ def move_to_destination(source, destination, job_name, sagemaker_session):
...
+    Args:
+        path (str): the original path. If the path is only ``"/"``,
+            then it is ignored.
+        job_name (str): the job name to be appended to the path.
+
+    Returns:
+        str: an S3 prefix of the form ``"path/job_name"``
+    """
+    path = path.strip("/")
+    return job_name if path == "" else "/".join((path, job_name))
```
ehsanmok, line 88: Just curious to know instead of doing string manipulation, wouldn't `os.path.join` or `pathlib.Path` a better options for correctly making unified paths for local and non-local modes?

laurenyu: it would, except that S3 always uses "/" as a separator, so using `os.path.join` would fail for Windows users.

as far as I can tell, there's not an easy way of overriding the separator used in `os.path.join` (I wrote a little about this in #1435). The other option would be to use something like [urlunparse](https://docs.python.org/3.7/library/urllib.parse.html#urllib.parse.urlunparse), but I think it might be a little heavy-handed for this use-case.

laurenyu: `PurePosixPath` does actually look to solve the problem, except that it was introduced in Python 3.4 and never backported to Python 2.7. since this SDK still currently supports Python 2.7, we can't use it here either. however, I'll make a note in our internal backlog to use `PurePosixPath` after we do get around to dropping Python 2.7 support. Thanks for sharing that! We've had quite a few conversations on our team around finding a module that would work no matter the OS, but hadn't yet found `pathlib`.

ehsanmok: Sounds great! thanks for letting me know.

PR #1664, opened 2020-07-02
change: reshape Artifacts into data frame in ExperimentsAnalytics

yzhu0:
Reshape Artifacts into data frame in ExperimentsAnalytics

https://sim.amazon.com/issues/AML-71634

### DIFF
--- a/src/sagemaker/analytics.py
+++ b/src/sagemaker/analytics.py
@@ -431,6 +431,8 @@ class ExperimentAnalytics(AnalyticsMetricsBase):
         metric_names=None,
         parameter_names=None,
         sagemaker_session=None,
+        input_artifact_names=None,
+        output_artifact_names=None,
     ):
         """Initialize a ``ExperimentAnalytics`` instance.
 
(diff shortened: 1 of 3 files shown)


src/sagemaker/analytics.py
```diff
@@ -516,6 +520,36 @@ def _reshape_metrics(self, metrics):
                     out["{} - {}".format(metric_name, stat_type)] = stat_value
         return out
 
+    def _reshape_input_artifacts(self, input_artifacts):
```
jerrypeng7773, line 523: why can't we just have _reshape_artifacts() which can be applied to both input/output-artifacts, so we dont need two unit tests over this too =)

src/sagemaker/analytics.py
```diff
@@ -463,6 +465,8 @@ def __init__(
         self._sort_order = sort_order
         self._metric_names = metric_names
         self._parameter_names = parameter_names
+        self._input_artifact_names = input_artifact_names
+        self._output_artifact_name = output_artifact_names
```
leoroy, line 469: You might want to add comments for these 2 arguments in the above section as well.

src/sagemaker/analytics.py
danabens: +1 would eliminate some redundancy

yzhu0: thats true,

src/sagemaker/analytics.py
yzhu0: added

danabens:
looks good but not clear why unit test check is failing

nadiaya:
Pylint has failed:
```
src/sagemaker/analytics.py:455:0: C0301: ·[1mLine too long (119/100)·[0m (·[1mline-too-long·[0m)
src/sagemaker/analytics.py:457:0: C0301: ·[1mLine too long (119/100)·[0m (·[1mline-too-long·[0m)
```

src/sagemaker/analytics.py
```diff
@@ -450,6 +452,11 @@ def __init__(
             sagemaker_session (sagemaker.session.Session): Session object which manages interactions
                 with Amazon SageMaker APIs and any other AWS services needed. If not specified,
                 one is created using the default AWS configuration chain.
+            input_artifact_names(dict optional):The input artifacts for the experiment. Examples of
+             input artifacts are datasets, algorithms, hyperparameters, source code, and instance
```
ChuyangDeng, line 456: Could you correct the indentations in the docstring?

yzhu0: fixed

PR #1668, opened 2020-07-02
feature: add BaseSerializer and BaseDeserializer

bveeramani:
*Issue #, if available:*

*Description of changes:* 
* Added `sagemaker.serializers.BaseSerializer`
* Added `sagemaker.serializers.BaseDeserializer`
* Added `sagemaker.predictor.LegacySerialzier`
* Added `sagemaker.predictor.LegacyDeserializer`
* Changed implementation of `sagemaker.predictor.Predictor.__init__`
* Changed implementation of `sagemaker.predictor.Predictor._handle_response`
* Changed implementation of `sagemaker.predictor.Predictor._create_request_args`

*Testing done:*
* Created and added two tests to `tests.unit.sagemaker.test_serializers`
* Created and added two tests to `tests.unit.sagemaker.test_deserializers`

## Merge Checklist

#### General

#### Tests


src/sagemaker/deserializers.py
```diff
@@ -0,0 +1,43 @@
...
+            content_type (str): The MIME type of the data.
+
+        Returns:
+            object: The data deserialized into an object.
+        """
+        raise NotImplementedError
+
+    @property
+    @abc.abstractmethod
+    def ACCEPT(self):
```
metrizable, line 41: nit-pep8: [lowercase](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names) -- only [constants](https://www.python.org/dev/peps/pep-0008/#constants) should be all caps:

```suggestion
    def accept(self):
```

src/sagemaker/deserializers.py
```diff
@@ -0,0 +1,43 @@
...
+        """Deserialize data received from an inference endpoint.
+
+        Args:
+            data (object): Data to be deserialized.
+            content_type (str): The MIME type of the data.
+
+        Returns:
+            object: The data deserialized into an object.
+        """
+        raise NotImplementedError
```
metrizable, line 37: remove this line as not needed since `ABCMeta` will enforce on instantiation (`@abstractmethod` set `__isabstractmethod__` to `True` on the the `funcobj`)

src/sagemaker/deserializers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @property
+    @abc.abstractmethod
+    def ACCEPT(self):
+        """The content type that is expected from the inference endpoint."""
+        raise NotImplementedError
```
metrizable, line 43: remove

src/sagemaker/predictor.py
```diff
@@ -59,27 +61,23 @@ def __init__(
...
                 the deserializer (default: None).
         """
         self.endpoint_name = endpoint_name
         self.sagemaker_session = sagemaker_session or Session()
         self.serializer = serializer
         self.deserializer = deserializer
-        self.content_type = content_type or getattr(serializer, "content_type", None)
-        self.accept = accept or getattr(deserializer, "accept", None)
+        self.content_type = content_type or getattr(serializer, "CONTENT_TYPE", None)
+        self.accept = accept or getattr(deserializer, "ACCEPT", None)
```
metrizable, line 80: this attribute must exist on `deserializer` instances, so just `deserializer.accept`?

src/sagemaker/predictor.py
```diff
@@ -406,6 +404,88 @@ def _get_model_names(self):
...
+        Args:
+            data (object): Data to be serialized.
+
+        Returns:
+            object: Serialized data used for a request.
+        """
+        return self.serializer(data)
+
+    @property
+    def CONTENT_TYPE(self):
```
metrizable, line 442: lower

tests/unit/sagemaker/test_deserializers.py
```diff
@@ -0,0 +1,34 @@
...
+# language governing permissions and limitations under the License.
+from __future__ import absolute_import
+
+import pytest
+
+from sagemaker import deserializers
+
+
+class BaseDeserializerTest:
+    def test_instantiate_deserializer_without_accept_attribute(self):
```
metrizable, line 21: you're testing whether `ABCMeta` functionality works -- standard library functionality/core cpython implementation details. you don't need to test this. it's being tested at the language level.

`@abc.abstractmethod` sets the `__isabstractmethod__` [on the func object](https://github.com/python/cpython/blob/5c117dd227e1b4c4f0a62564d8592f1ba45c91eb/Lib/abc.py#L24) and [enforcement of override](https://github.com/python/cpython/blob/5c117dd227e1b4c4f0a62564d8592f1ba45c91eb/Lib/abc.py#L10-L12) is during instantiation

tests/unit/sagemaker/test_deserializers.py
```diff
@@ -0,0 +1,34 @@
...
+        class StubDeserializer(deserializers.BaseDeserializer):
+            def deserialize(self, data, content_type):
+                pass
+
+        with pytest.raises(TypeError):
+            StubDeserializer()
+
+    def test_instantiate_deserializer_without_deserialize_method(self):
```
metrizable, line 29: same, test not needed

src/sagemaker/deserializers.py
```diff
@@ -0,0 +1,43 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Implements methods for deserializing data returned from an inference endpoint."""
+from __future__ import absolute_import
+
+import abc
+
+
+class BaseDeserializer(abc.ABC):
+    """Abstract base class for creation of new deserializers.
```
metrizable, line 20: `@ignorable` just thinking out loud: a serializer/deserializer is in some sense a function along a special constant. i wonder if it's possible to reduce the "public interface" to just a function and "hide" the constant? maybe push the constants to a dict of these functions to content types.

src/sagemaker/predictor.py
```diff
@@ -59,27 +61,23 @@ def __init__(
...
+                object, used to encode data for an inference endpoint
+                (default: None).
+            deserializer (sagemaker.deserializers.BaseDeserializer): A
+                deserializer object, used to decode data from an inference
+                endpoint (default: None).
             content_type (str): The invocation's "ContentType", overriding any
-                ``content_type`` from the serializer (default: None).
+                ``CONTENT_TYPE`` from the serializer (default: None).
             accept (str): The invocation's "Accept", overriding any accept from
                 the deserializer (default: None).
```
laurenyu, line 73: [potentially out of scope of this PR] since the serializers/deserializers now enforce having the content/accept type set, I wonder if there really is any use with these two attributes. alternatively, there might not be much use to the `serializer`/`deserializer` args depending on how you want to address [Eric's earlier comment](https://github.com/aws/sagemaker-python-sdk/pull/1668/files#r449326977)

src/sagemaker/predictor.py
```diff
@@ -406,6 +404,88 @@ def _get_model_names(self):
...
+        """
+        return self.serializer(data)
+
+    @property
+    def CONTENT_TYPE(self):
+        """The MIME type of the data sent to the inference endpoint."""
+        return self.content_type
+
+
+class LegacyDeserializer(BaseDeserializer):
```
laurenyu, line 447: with this here, I don't think you need "breaking" for your commit label?

src/sagemaker/deserializers.py
bveeramani: ACCEPT is a constant. The use of @property with @abstractmethod is confusing, because it suggests that the user needs to implement a method rather than a class-level constant. But, defining a class-level constant also satisfies the interface. 

For example, the following code satisfies the `BaseSerializer` interface:
```
class FooDeserializer(BaseDeserialzier):
    ACCEPT = "application/json"
   
    def deserialize(self, data):
        pass
```

This is how I'm planning on implementing all the native serializers/deserializers, and also how I intend to encourage users to implement custom serializer. My reasons are two-fold:

1) I assume that there would be no reason to change accept and content-types at runtime, so it makes sense to specify them as a constant.
2) Using a class-level constant is a pattern seen in other Python packages (e.g., [see this PyTorch object](https://pytorch.org/tutorials/intermediate/pruning_tutorial.html#extending-torch-nn-utils-prune-with-custom-pruning-functions)).

src/sagemaker/deserializers.py
bveeramani: Fixed

src/sagemaker/deserializers.py
bveeramani: Fixed

src/sagemaker/predictor.py
bveeramani: Fixed

src/sagemaker/predictor.py
bveeramani: https://github.com/aws/sagemaker-python-sdk/pull/1668#discussion_r450385627

tests/unit/sagemaker/test_deserializers.py
bveeramani: Fixed

tests/unit/sagemaker/test_deserializers.py
bveeramani: Fixed

src/sagemaker/predictor.py
bveeramani: Actually, after re-running the unit tests, I realize that we need to keep the `getattr`s (at least for the time-being). By default, serializer and deserializer are None, in which case there will be no CONTENT_TYPE or ACCEPT attributes. If the user passes in no value for content_type and no value for serializer, then the program will crash.

src/sagemaker/deserializers.py
metrizable: makes sense re: class-level constants. my nit comments were more about the violating naming convention for the method in the base class, as it wasn't clear that your intent/desire was to have the derived classes implement the property by defining a constant.

src/sagemaker/predictor.py
```diff
@@ -406,6 +413,88 @@ def _get_model_names(self):
         return [d["ModelName"] for d in production_variants]
 
 
+class LegacySerializer(BaseSerializer):
+    """Wrapper that makes legacy serializers forward compatibile."""
+
+    def __init__(self, serializer):
+        """Placeholder docstring.
```
laurenyu, line 420: don't forget to fill this in :)

src/sagemaker/predictor.py
```diff
@@ -406,6 +413,88 @@ def _get_model_names(self):
...
+
+        Args:
+            serializer (callable): A legacy serializer.
+        """
+        self.serializer = serializer
+        self.content_type = getattr(serializer, "content_type", None)
+
+    def __call__(self, *args, **kwargs):
```
laurenyu, line 428: are the function signatures that different across legacy implementations? I think it would be better to avoid `*args` and `**kwargs` here if possible.

src/sagemaker/deserializers.py
```diff
@@ -0,0 +1,41 @@
...
+            content_type (str): The MIME type of the data.
+
+        Returns:
+            object: The data deserialized into an object.
+        """
+
+    @property
+    @abc.abstractmethod
+    def ACCEPT(self):
+        """The content type that is expected from the inference endpoint."""
```
metrizable, line 41: have you given any more thought to how to support serdes (serializers/deserializers) that can and should be able to accept and operate on a collection of content types?

bveeramani: Yeah, the user will be able to specify a string, list, or dictionary of accept types. Because adding support for multiple accept types necessitates changes to the sagemaker-inference-toolkit, I'm planning on first migrating the existing serializers/deserializers (which have only one accept type) and then adding support for multiple accept types.

src/sagemaker/predictor.py
bveeramani: Some of the classes have different function signatures. For example, `_NPYSerializer` has the signature `def __call__(self, data, dtype=None)`. If I were to not use *args and **kwargs, then I would need to make changes (or remove) some tests. Since LegacySerializer and LegacyDeserializer are temporary fixes so that I can make incremental changes, IMO I think it's fine.

PR #1682, opened 2020-07-08
breaking: Move _CsvDeserializer to sagemaker.deserializers and rename to CSVDeserializer

bveeramani:
*Issue #, if available:* #1694 

*Description of changes:*
Moved sagemaker.predictor._CsvDeserializer to sagemaker.deserializers.CSVDeserializer
*Testing done:*
Added three tests to test_deserializers.

## Merge Checklist

#### General

#### Tests

###


src/sagemaker/deserializers.py
```diff
@@ -60,3 +62,34 @@ def deserialize(self, data, content_type):
...
+        Args:
+            data (object): Data to be deserialized.
+            content_type (str): The MIME type of the data.
+
+        Returns:
+            list: The data deserialized into a list of lists representing the
+                contents of a CSV file.
+        """
+        try:
+            string = data.read().decode(self.encoding)
```
metrizable, line 92: You should probably use a different name, since `string` is the name of a [module in the standard library](https://docs.python.org/3/library/string.html).

src/sagemaker/deserializers.py
```diff
@@ -60,3 +62,34 @@ def deserialize(self, data, content_type):
...
+            encoding (str): The string encoding to use (default: "UTF-8").
+        """
+        self.encoding = encoding
+
+    def deserialize(self, data, content_type):
+        """Deserialize data from an inference endpoint into a list of lists.
+
+        Args:
+            data (object): Data to be deserialized.
+            content_type (str): The MIME type of the data.
```
metrizable, line 85: unused arg, here and elsewhere

src/sagemaker/deserializers.py
```diff
@@ -60,3 +62,34 @@ def deserialize(self, data, content_type):
...
+        Args:
+            encoding (str): The string encoding to use (default: "UTF-8").
+        """
+        self.encoding = encoding
+
+    def deserialize(self, data, content_type):
+        """Deserialize data from an inference endpoint into a list of lists.
+
+        Args:
+            data (object): Data to be deserialized.
```
metrizable, line 84: i wonder if there is there a stronger type that you can hint at here. this data object must respond to `.read()`, maybe BufferedIOBase or TextIOBase.

tests/unit/sagemaker/test_deserializers.py
```diff
@@ -23,3 +23,27 @@ def test_bytes_deserializer():
     result = deserializer.deserialize(io.BytesIO(b"[1, 2, 3]"), "application/json")
 
     assert result == b"[1, 2, 3]"
+
+
+def test_csv_deserializer_single_element():
+    deserializer = CSVDeserializer()
```
metrizable, line 29: it may be more pythonic to define fixture once

```
@pytest.fixture
def csv_deserializer():
    return CSVDeserializer()


def test_1(csv_deserializer):
    ...

def test_n(csv_deserializer):
    ...
```

you can even define scope if desired

tests/unit/sagemaker/test_deserializers.py
```diff
@@ -23,3 +23,27 @@ def test_bytes_deserializer():
...
+
+    result = deserializer.deserialize(io.BytesIO(b"1"), "text/csv")
+
+    assert result == [["1"]]
+
+
```
metrizable, line 35: you will also want to test for content that comes from posix-compliant [text files](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_392) where each [line](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_205) ends in a [newline](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_238) character. something like:

```
content = b"1,2,3\n4,5,6\n"
```
this would represent the content of a posix-compliant file comprised of two lines.

src/sagemaker/deserializers.py
bveeramani: Fixed

src/sagemaker/deserializers.py
bveeramani: Yeah. I've kept the content_type parameter in so that BaseDeserializer and all of it's subclasses have a consistent function signature

src/sagemaker/deserializers.py
bveeramani: Fixed

tests/unit/sagemaker/test_deserializers.py
bveeramani: Fixed

tests/unit/sagemaker/test_deserializers.py
bveeramani: Fixed

src/sagemaker/deserializers.py
```diff
@@ -60,3 +62,34 @@ def deserialize(self, data, content_type):
...
         finally:
             data.close()
+
+
+class CSVDeserializer(BaseDeserializer):
+    """Deserialize a stream of bytes into a list of lists."""
+
+    ACCEPT = "test/csv"
+
+    def __init__(self, encoding="UTF-8"):
```
metrizable, line 72: nit: probably most common [alias](https://docs.python.org/3/library/codecs.html#standard-encodings) used in the [docs](https://docs.python.org/3/library/stdtypes.html#bytearray.decode), in the wild, and even in the [cpython codecs implementation](https://github.com/python/cpython/blob/3.8/Lib/codecs.py#L1123) is lowercase `"utf-8"`

bveeramani: Fixed

src/sagemaker/deserializers.py
bveeramani: NumpyDeserializer needs to know the content type to properly deserialize a response. The content type would also be required to deserialize JSON lines and JSON-LD formats. Presumably, some custom user deserializers may also need the content-type of the returned data to conditionally deserialize data.

To your second point, I don't think there's any case where the user would directly use a deserializer. The deserializer is used in the Predict class at this line: https://github.com/aws/sagemaker-python-sdk/blob/zwei/src/sagemaker/predictor.py#L129. A content-type is always passed as the second argument.

src/sagemaker/deserializers.py
```diff
@@ -96,6 +98,37 @@ def deserialize(self, data, content_type):
             data.close()
 
 
+class CSVDeserializer(BaseDeserializer):
+    """Deserialize a stream of bytes into a list of lists."""
+
+    ACCEPT = "test/csv"
```
metrizable, line 104: should this be `"text/csv"`?

Issue #1694, opened 2020-07-09
Move SerDe to dedicated modules


bveeramani:
To improve discoverability, we should move and rename the following objects:

| Before | After | PR |
|-|-|-|
| `sagemaker.predictor._CsvDeserializer` | `sagemaker.deserializers.CSVDeserializer` | #1682  |
| `sagemaker.predictor._CsvSerializer` | `sagemaker.serializers.CSVSerializer` |  #1700 |
| `sagemaker.predictor.BytesDeserializer` | `sagemaker.deserializers.BytesDeserializers` |  #1674  |
| `sagemaker.predictor.StringDeserializer` | `sagemaker.deserializers.StringDeserializer` |  #1677 |
| `sagemaker.predictor.StreamDeserializer` | `sagemaker.deserializers.StreamDeserializer` |  #1679 |
| `sagemaker.predictor._JsonSerializer` | `sagemaker.serializers.JSONSerializer` |  #1698 |
| `sagemaker.predictor._NumpyDeserializer` | `sagemaker.deserializers.NumpyDeserializer` |  #1695 |
| `sagemaker.predictor._NPYSerializer` | `sagemaker.serializers.NumpySerializer` |  #1691 |
| `sagemaker.amazon.common.numpy_to_record_serializer` | `sagemaker.amazon.serializers.RecordSerializer` | #1690  |
| `sagemaker.amazon.common.record_deserializer` | `sagemaker.amazon.deserializers.RecordDeserializer` |  #1683 |
| `sagemaker.predictor._JsonDeserializer` | `sagemaker.deserializers.JSONDeserializer` |  #1697 |

PR #1735, opened 2020-07-22
feature: Add v2 SerDe compatability

bveeramani:
*Issue #, if available:* #1694

*Description of changes:*
* Changed `ASTTransformer` implementation such that a modified node, None, or a new node can be returned (previously only a modified node would be returned).
* Update all `Modifier`s to comply with new `ASTTransformer` implementation
* Added `serde` module that contains classes for upgrading SerDe-related code.

*Testing done:*
* Added `test_serde`

## Merge Checklist

#### General

#### Tests


src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+import ast
+
+import pasta
+
+from sagemaker.cli.compatibility.v2.modifiers import matching
+from sagemaker.cli.compatibility.v2.modifiers.modifier import Modifier
+
+# The values are tuples so that the object can be passed to matching.matches_any.
+OLD_CLASS_NAME_TO_NAMESPACES = {
+    "_CsvSerializer": ("sagemaker.predictor",),
```
metrizable, line 27: is it necessary to use tuples here? seems there are no tuples with more than one value.

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+OLD_OBJECT_NAME_TO_NEW_CLASS_NAME = {
+    "csv_serializer": "CSVSerializer",
+    "json_serializer": "JSONSerializer",
+    "npy_serializer": "NumpySerializer",
+    "csv_deserializer": "CSVDeserializer",
+    "json_deserializer": "JSONDeserializer",
+    "numpy_deserializer": "NumpyDeserializer",
+}
+
+OLD_AMAZON_CLASS_NAMES = set(
```
metrizable, line 78: don't need to construct a `set` from your comprehension, since it's already a set:

```
[ins] In [1]: foo = {1, 2, 2}

[ins] In [2]: type(foo)
Out[2]: set

[ins] In [3]: foo
Out[3]: {1, 2}
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+                see https://docs.python.org/3/library/ast.html#abstract-grammar.
+
+        Returns:
+            bool: If the ``ast.Call`` instantiates a SerDe class.
+        """
+        return matching.matches_any(node, OLD_CLASS_NAME_TO_NAMESPACES)
+
+    def modify_node(self, node):
+        """Modifies the ``ast.Call`` node to use the classes for SerDe
+        available in version 2.0 and later of the Python SDK:
```
metrizable, line 130: nit-[pep257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings): one line, blank line, body with detail

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+        if "sagemaker.amazon.common" in namespaces
+    }
+)
+NEW_AMAZON_CLASS_NAMES = set(
+    {
+        class_name
+        for class_name, namespaces in NEW_CLASS_NAME_TO_NAMESPACES.items()
+        if "sagemaker.amazon.common" in namespaces
+    }
+)
```
laurenyu, line 91: food for thought: it might be easier to hardcode which two classes are the "Amazon" serializers and build out the dictionaries from there. something like:

```python
OLD_CLASS_NAMES_TO_NAMESPACES = {class_name: "sagemaker.amazon.common" for class_name in OLD_AMAZON_CLASS_NAMES}
OLD_CLASS_NAMES_TO_NAMESPACES.update({class_name: "sagemaker.predictor" for class_name in OLD_PREDICTOR_CLASS_NAMES})
```

@metrizable might have thoughts on what's more Pythonic

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+OLD_OBJECT_NAMES = set(OLD_OBJECT_NAME_TO_NEW_CLASS_NAME.keys())
+
+
+class SerdeConstructorRenamer(Modifier):
+    """A class to rename SerDe classes."""
+
+    def node_should_be_modified(self, node):
+        """Checks if the ``ast.Call`` node instantiates a SerDe class.
+
+        This looks for the following calls:
```
laurenyu, line 105: you might want to note that it also looks for those calls without the namespace

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+        if class_name in OLD_AMAZON_CLASS_NAMES:
+            if isinstance(node.func, ast.Name):
+                node.func.id = new_class_name
+            elif isinstance(node.func, ast.Attribute):
+                node.func.attr = new_class_name
+            return node
+
+        namespace_name = NEW_CLASS_NAME_TO_NAMESPACES[new_class_name][0]
+        subpackage_name = namespace_name[namespace_name.find(".") + 1 :]
+        assert subpackage_name in {"serializers", "deserializers"}
```
laurenyu, line 163: what is the purpose of this?

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+        # We don't change the namespace for Amazon SerDe.
+        if class_name in OLD_AMAZON_CLASS_NAMES:
+            if isinstance(node.func, ast.Name):
+                node.func.id = new_class_name
+            elif isinstance(node.func, ast.Attribute):
+                node.func.attr = new_class_name
+            return node
+
+        namespace_name = NEW_CLASS_NAME_TO_NAMESPACES[new_class_name][0]
+        subpackage_name = namespace_name[namespace_name.find(".") + 1 :]
```
laurenyu, line 162: (optional) you could also do:

```suggestion
        subpackage_name = namespace_name.split(".")[1]
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
metrizable, line 163: in production code, if you want to fail in python on some condition, explicitly raise an exception on that condition, as assertions are stripped in [optimized mode](https://docs.python.org/3/using/cmdline.html#cmdoption-o):

```
$ python -c "assert False; print('foo')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AssertionError

$ python -O -c "assert False; print('foo')"
foo
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+        assert subpackage_name in {"serializers", "deserializers"}
+        return pasta.parse("%s.%s()" % (subpackage_name, new_class_name)).body[0].value
```
laurenyu, line 164: I'm not sure if there are any tradeoffs here other than readability, but my inclination is to say that it would be better to use the `ast.Call` and `ast.Attribute` constructors directly

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+            node (ast.ImportFrom): a node that represents a ``from ... import ... `` statement.
+                For more, see https://docs.python.org/3/library/ast.html#abstract-grammar.
+
+        Returns:
+            ast.ImportFrom: a node that represents a import statement, which has
+                been modified to remove imported serializers. If nothing is
+                imported, None is returned.
+        """
+        i = 0
+        while i < len(node.names):
```
laurenyu, line 251: does removing the name mess with for-loop iteration?

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+        Args:
+            node (ast.ImportFrom): a node that represents a ``from ... import ... `` statement.
+                For more, see https://docs.python.org/3/library/ast.html#abstract-grammar.
+
+        Returns:
+            bool: True if and only if the ``ast.ImportFrom`` imports a SerDe
+                from the ``sagemaker.predictor`` module.
+        """
+        return node.module == "sagemaker.predictor" and any(
+            [name.name in (OLD_CLASS_NAMES | OLD_OBJECT_NAMES) for name in node.names]
```
metrizable, line 235: no need to construct a `list` (i.e. remove the `[` `]`) as generator expressions are just fine for `any`:

```
[ins] In [1]: any(x for x in [False, True])
Out[1]: True
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,413 @@
...
+            name = node.names[i].name
+            if name in OLD_CLASS_NAMES | OLD_OBJECT_NAMES:
+                node.names.pop(i)
+            else:
+                i += 1
+
+        if not node.names:
+            return None
+
+        return node
```
metrizable, line 261: maybe positive statement instead?

```suggestion
        return node if node.names else None
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: It should right now. The unit tests for that method tests both calls with namespaces and without.

```
@pytest.mark.parametrize(
    "src, expected",
    [
        ("sagemaker.predictor._CsvSerializer()", True),
        ("sagemaker.predictor._JsonSerializer()", True),
        ("sagemaker.predictor._NpySerializer()", True),
        ("sagemaker.predictor._CsvDeserializer()", True),
        ("sagemaker.predictor.BytesDeserializer()", True),
        ("sagemaker.predictor.StringDeserializer()", True),
        ("sagemaker.predictor.StreamDeserializer()", True),
        ("sagemaker.predictor._NumpyDeserializer()", True),
        ("sagemaker.predictor._JsonDeserializer()", True),
        ("sagemaker.predictor.OtherClass()", False),
        ("sagemaker.amazon.common.numpy_to_record_serializer()", True),
        ("sagemaker.amazon.common.record_deserializer()", True),
        ("_CsvSerializer()", True),
        ("_JsonSerializer()", True),
        ("_NpySerializer()", True),
        ("_CsvDeserializer()", True),
        ("BytesDeserializer()", True),
        ("StringDeserializer()", True),
        ("StreamDeserializer()", True),
        ("_NumpyDeserializer()", True),
        ("_JsonDeserializer()", True),
        ("numpy_to_record_serializer()", True),
        ("record_deserializer()", True),
        ("OtherClass()", False),
    ],
)
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: The assertion? I left it as a sanity check, but it can easily be removed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: No. If I had used a for-loop the behavior would've been incorrect, but because I used a while loop it should be fine

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: The assertion is supposed to be a sanity check. As far as I know, `subpackage_name` should always be either `"serializers"` or `"deserializers"`

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Did not know that. That's useful to know. Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/ast_transformer.py
```diff
@@ -52,52 +62,85 @@ class ASTTransformer(ast.NodeTransformer):
...
 
         Args:
             node (ast.Call): a node that represents a function call.
 
         Returns:
-            ast.Call: a node that represents a function call, which has
-                potentially been modified from the original input.
+            ast.AST: if the returned node is None, the original node is removed
+            from its location. Otherwise, the original node is replaced with the
+            returned node.
```
laurenyu, line 74: use a hanging indent for argument/return value descriptions (applies throughout)

```suggestion
            ast.AST: if the returned node is None, the original node is removed
                from its location. Otherwise, the original node is replaced with the
                returned node.
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,407 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Classes to modify serializer and deserializer code to be compatible with
+version 2.0 and later of the SageMaker Python SDK.
+"""
```
laurenyu, line 15: I guess I've seen Eric quote [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) enough times now :joy: - this should be a one-liner

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,407 @@
...
+    "_CsvSerializer": ("sagemaker.predictor",),
+    "_JsonSerializer": ("sagemaker.predictor",),
+    "_NpySerializer": ("sagemaker.predictor",),
+    "_CsvDeserializer": ("sagemaker.predictor",),
+    "BytesDeserializer": ("sagemaker.predictor",),
+    "StringDeserializer": ("sagemaker.predictor",),
+    "StreamDeserializer": ("sagemaker.predictor",),
+    "_NumpyDeserializer": ("sagemaker.predictor",),
+    "_JsonDeserializer": ("sagemaker.predictor",),
+}
```
laurenyu, line 39: you could also use a comprehension here since these all have the same namespace

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,407 @@
...
+        Args:
+            node (ast.Name): a node that represents a Python identifier.
+
+        Returns:
+            ast.Call: a node that represents the instantiation of a SerDe object.
+        """
+        object_name = node.id if isinstance(node, ast.Name) else node.attr
+        new_class_name = OLD_OBJECT_NAME_TO_NEW_CLASS_NAME[object_name]
+        namespace_name = NEW_CLASS_NAME_TO_NAMESPACES[new_class_name][0]
+        subpackage_name = namespace_name[namespace_name.find(".") + 1 :]
```
laurenyu, line 208: same thing here with using `split()` instead

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,407 @@
...
+    def __init__(self, class_names, import_node):
+        """Initialize the ``class_names`` and ``import_node`` attributes.
+
+        Args:
+            class_names (set): If any of these class names are referenced in the
+                module, then ``import_node`` is inserted.
+            import_node (ast.ImportFrom): The AST node to insert.
+        """
+        self.class_names = class_names
+        self.import_node = import_node
```
laurenyu, line 315: I'm not sure if one way provides any benefit over the other, but you could also make these abstract property methods

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,407 @@
...
+    """
+
+    def __init__(self):
+        # Amazon SerDe are not defined in the sagemaker.serializers module.
+        class_names = {
+            class_name
+            for class_name in NEW_CLASS_NAMES - NEW_AMAZON_CLASS_NAMES
+            if "Serializer" in class_name
+        }
+        import_node = pasta.parse("from sagemaker import serializers\n").body[0]
```
laurenyu, line 378: I think it'd be better to avoid `pasta` here as well

same below

src/sagemaker/cli/compatibility/v2/ast_transformer.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: I thought about it, and I can't really decide which is better, so I'm leaving it as-is.

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
bveeramani: Fixed

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,410 @@
...
+        Args:
+            node (ast.ImportFrom): a node that represents a ``from ... import ... `` statement.
+                For more, see https://docs.python.org/3/library/ast.html#abstract-grammar.
+
+        Returns:
+            bool: True if and only if the ``ast.ImportFrom`` imports a SerDe from
+                the ``sagemaker.amazon.common`` module.
+        """
+        return node.module == "sagemaker.amazon.common" and any(
+            [alias.name in OLD_AMAZON_CLASS_NAMES for alias in node.names]
```
metrizable, line 277: remove `[` as unneeded as generator expression is fine
```suggestion
            alias.name in OLD_AMAZON_CLASS_NAMES for alias in node.names
```

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,410 @@
...
+    - ``sagemaker.serializers.NumpyDeserializer``
+    - ``sagemaker.serializer.JSONDeserializer``
+
+    Because ``DeserializerImportInserter`` is gauranteed to run after
+    ``SerdeConstructorRenamer`` and ``SerdeObjectRenamer``,
+    we only need to check for the new deserializer class names.
+    """
+
+    def __init__(self):
+        # Amazon SerDe are not defined in the sagemaker.serializers module.
```
metrizable, line 401: nit: just make this a docstring?

src/sagemaker/cli/compatibility/v2/modifiers/serde.py
```diff
@@ -0,0 +1,410 @@
...
+    - ``sagemaker.serializers.JSONSerializer``
+    - ``sagemaker.serializer.NumpySerializer``
+
+    Because ``SerializerImportInserter`` is gauranteed to run after
+    ``SerdeConstructorRenamer`` and ``SerdeObjectRenamer``,
+    we only need to check for the new serializer class names.
+    """
+
+    def __init__(self):
+        # Amazon SerDe are not defined in the sagemaker.serializers module.
```
metrizable, line 371: nit: just make this a docstring?

PR #1752, opened 2020-07-24
feature: Add Granular Target Description support for compilation (Neo)

apivovarov:
#### Description

Add Granular Target Description support for DL model compilation.
Example:
```
    model.compile(
        target_instance_family=None,
        input_shape={"data": [1, 3, 1024, 1024]},
        output_path="s3://output",
        role="role",
        framework="tensorflow",
        job_name="compile-model",
        target_platform_os="LINUX",
        target_platform_arch="ARM64",
        target_platform_accelerator="NVIDIA",
        compiler_options={"gpu-code": "sm_72", "trt-ver": "6.0.1", "cuda-ver": "10.1"}
    )
```

#### Tests

### DIFF
--- a/src/sagemaker/estimator.py
+++ b/src/sagemaker/estimator.py
@@ -512,6 +512,10 @@ class EstimatorBase(with_metaclass(ABCMeta, object)):
         framework_version=None,
         compile_max_run=15 * 60,
         tags=None,
+        target_platform_os=None,
+        target_platform_arch=None,
+        target_platform_accelerator=None,
+        compiler_options=None,
         **kwargs
     ):
         """Compile a Neo model using the input model.
@@ -536,6 +540,21 @@ class EstimatorBase(with_metaclass(ABCMeta, object)):
             tags (list[dict]): List of tags for labeling a compilation job. For
                 more, see
                 https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html.
+            target_platform_os (str): Target Platform OS, for example: 'LINUX'.
+                For allowed strings see
+                https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html.
+                It can be used instead of target_instance_family.
+            target_platform_arch (str): Target Platform Architecture, for example: 'X86_64'.
+                For allowed strings see
+                https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html.
+                It can be used instead of target_instance_family.
+            target_platform_accelerator (str, optional): Target Platform Accelerator,
+                for example: 'NVIDIA'. For allowed strings see
+                https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html.
+                It can be used instead of target_instance_family.
+            compiler_options (dict, optional): Additional parameters for compiler.
+                Compiler Options are TargetPlatform / target_instance_family specific. See
+                https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html for details.
             **kwargs: Passed to invocation of ``create_model()``.
                 Implementations may customize ``create_model()`` to accept
                 ``**kwargs`` to customize model creation during deploy. For
@@ -565,6 +584,10 @@ class EstimatorBase(with_metaclass(ABCMeta, object)):
             compile_max_run,
             framework=framework,
             framework_version=framework_version,
+            target_platform_os=target_platform_os,
+            target_platform_arch=target_platform_arch,
+            target_platform_accelerator=target_platform_accelerator,
+            compiler_options=compiler_options,
         )
         return self._compiled_models[target_instance_family]
 
--- a/src/sagemaker/model.py
+++ b/src/sagemaker/model.py
@@ -215,6 +215,10 @@ class Model(object):
         job_name,
         framework,
         tags,
+        target_platform_os=None,
+        target_platform_arch=None,
+        target_platform_accelerator=None,
+        compiler_options=None,
     ):
         """
         Args:
@@ -226,19 +230,45 @@ class Model(object):
             job_name:
             framework:
             tags:
+            target_platform_os:
+            target_platform_arch:
+            target_platform_accelerator:
+            compiler_options:
         """
         input_model_config = {
             "S3Uri": self.model_data,
-            "DataInputConfig": input_shape
-            if not isinstance(input_shape, dict)
-            else json.dumps(input_shape),
+            "DataInputConfig": json.dumps(input_shape)
+            if isinstance(input_shape, dict)
+            else input_shape,
             "Framework": framework,
         }
         role = self.sagemaker_session.expand_role(role)
         output_model_config = {
-            "TargetDevice": target_instance_type,
             "S3OutputLocation": output_path,
         }
+
+        if target_instance_type is not None:
+            output_model_config["TargetDevice"] = target_instance_type
+        else:
+            if target_platform_os is None and target_platform_arch is None:
+                raise ValueError(
+                    "target_instance_type or (target_platform_os and target_platform_arch) "
+                    "should be provided"
+                )
+            target_platform = {
+                "Os": target_platform_os,
+                "Arch": target_platform_arch,
+            }
+            if target_platform_accelerator is not None:
+                target_platform["Accelerator"] = target_platform_accelerator
+            output_model_config["TargetPlatform"] = target_platform
+
+        if compiler_options is not None:
+            output_model_config["CompilerOptions"] = (
+                json.dumps(compiler_options)
+                if isinstance(compiler_options, dict)
+                else compiler_options
+            )
 
         return {
             "input_model_config": input_model_config,
@@ -320,6 +350,10 @@ class Model(object):
         compile_max_run=5 * 60,
         framework=None,
         framework_version=None,
+        target_platform_os=None,
+        target_platform_arch=None,
+        target_platform_accelerator=None,
+        compiler_options=None,
     ):
         """Compile this ``Model`` with SageMaker Neo.
 
@@ -328,6 +362,9 @@ class Model(object):
                 run your model after compilation, for example: ml_c5. For allowed
                 strings see
                 https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html.
+                Alternatively, you can select an OS, Architecture and Accelerator using
+                ``target_platform_os``, ``target_platform_arch``,
+                and ``target_platform_accelerator``.
             input_shape (dict): Specifies the name and shape of the expected
                 inputs for your trained model in json dictionary form, for
                 example: {'data': [1,3,1024,1024]}, or {'var1': [1,1,28,28],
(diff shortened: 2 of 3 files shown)


src/sagemaker/estimator.py
```diff
@@ -512,6 +512,10 @@ def compile_model(
         framework_version=None,
         compile_max_run=15 * 60,
         tags=None,
+        target_platform_os=None,
+        target_platform_arch=None,
+        target_platform_accelerator=None,
+        compiler_options=None,
```
laurenyu, line 518: add these to the docstring as well

src/sagemaker/model.py
```diff
@@ -236,10 +244,27 @@ def _compilation_job_config(
...
             "S3OutputLocation": output_path,
         }
 
+        if target_instance_type is not None:
+            output_model_config["TargetDevice"] = target_instance_type
+        else:
+            assert target_platform_os is not None and target_platform_arch is not None, (
+                "target_instance_type or (target_platform_os and target_platform_arch) "
+                "should be provided"
+            )
```
laurenyu, line 256: it's better practice to avoid using `assert`s in source code, and instead raise a `ValueError` in this case

https://stackoverflow.com/a/945135
https://stackoverflow.com/a/1838411

src/sagemaker/model.py
```diff
@@ -328,6 +357,7 @@ def compile(
                 run your model after compilation, for example: ml_c5. For allowed
                 strings see
                 https://docs.aws.amazon.com/sagemaker/latest/dg/API_OutputConfig.html.
+                Alternatively, you can select OS, Arch and Accelerator using Target Platform
```
laurenyu, line 360:
```suggestion
                Alternatively, you can select an OS, Architecture, and Accelerator using
                ``target_platform_os``, ``target_platform_arch``, and ``target_platform_accelerator``.

File: src/sagemaker/model.py [hunk 4]
```diff
@@ -345,6 +375,16 @@ def compile(
                 model. Allowed values: 'mxnet', 'tensorflow', 'keras', 'pytorch',
                 'onnx', 'xgboost'
             framework_version (str):
+            target_platform_os (str): Target Platform OS. Allowed values: 'LINUX', 'ANDROID'.
+                It can be used instead of target_instance_family.
+            target_platform_arch (str): Target Platform Architecture. Allowed values: 'X86_64',
+                'X86', 'ARM64', 'ARM_EABIHF', 'ARM_EABI'.
+                It can be used instead of target_instance_family.
+            target_platform_accelerator (str): Target Platform Accelerator. Allowed values:
+                'INTEL_GRAPHICS', 'MALI', 'NVIDIA'. Optional.
```
laurenyu, line 384: can you instead link to docs that list the allowed versions? I assume they might change in the future, and if that happens, this docstring will inevitably be forgotten.

src/sagemaker/model.py
```diff
@@ -345,6 +375,16 @@ def compile(
...
+            compiler_options (str): Additional parameters for compiler (Optional, in JSON format).
```
laurenyu, line 385: if it needs to be in JSON, would it make more sense to accept a dict, and then let the Python SDK call `json.dumps`?

src/sagemaker/model.py
```diff
@@ -375,31 +415,41 @@ def compile(
...
+                    "please deploy the model manually.",
+                    target_instance_family,
+                )
         else:
             LOGGER.warning(
-                "The instance type %s is not supported to deploy via SageMaker,"
-                "please deploy the model manually.",
-                target_instance_family,
+                "Devices described by Target Platform Os, Arch and Accelerator are not supported"
+                "to deploy via SageMaker, please deploy the model manually."
```
laurenyu, line 452:
```suggestion
                "Deploying devices described by Target Platform OS, Architecture, and Accelerator "
                "is not supported via SageMaker. Please deploy the model manually."
```

src/sagemaker/model.py
```diff
@@ -375,31 +415,41 @@ def compile(
...
+                    target_instance_family,
+                    framework,
+                    framework_version,
+                )
+                self._is_compiled_model = True
+            else:
+                LOGGER.warning(
+                    "The instance type %s is not supported to deploy via SageMaker,"
+                    "please deploy the model manually.",
+                    target_instance_family,
```
laurenyu, line 447: (I know this isn't your change, but while you're here...)

```suggestion
                    "Deploying instance type %s is not supported via SageMaker."
                    "Please deploy the model manually.",
                    target_instance_family,
```

tests/unit/sagemaker/model/test_neo.py
```diff
@@ -95,6 +95,45 @@ def test_compile_model_for_edge_device_tflite(sagemaker_session):
...
+        output_path="s3://output",
+        role="role",
+        framework="tensorflow",
+        job_name="compile-model",
+        target_platform_os="LINUX",
+        target_platform_arch="ARM64",
+        target_platform_accelerator="NVIDIA",
+        compiler_options="{'gpu-code': 'sm_72', 'trt-ver': '6.0.1', 'cuda-ver': '10.1'}",
+    )
+    assert model._is_compiled_model is False
```
laurenyu, line 115: why wouldn't the model be considered compiled?

tests/unit/sagemaker/model/test_neo.py
```diff
@@ -95,6 +95,45 @@ def test_compile_model_for_edge_device_tflite(sagemaker_session):
...
+
+
+def test_compile_model_android_armv7(sagemaker_session):
+    sagemaker_session.wait_for_compilation_job = Mock(
+        return_value=DESCRIBE_COMPILATION_JOB_RESPONSE
+    )
+    model = _create_model(sagemaker_session)
+    model.compile(
+        target_instance_family=None,
+        input_shape={"data": [1, 3, 1024, 1024]},
```
mchoi8739, line 125: input_shape={'data': [1, 3, 1024, 1024]}

^ suggesting to use single-quote (based on what we discussed earlier)

tests/unit/sagemaker/model/test_neo.py
```diff
@@ -95,6 +95,45 @@ def test_compile_model_for_edge_device_tflite(sagemaker_session):
...
 
 
+def test_compile_model_linux_arm64_nvidia(sagemaker_session):
+    sagemaker_session.wait_for_compilation_job = Mock(
+        return_value=DESCRIBE_COMPILATION_JOB_RESPONSE
+    )
+    model = _create_model(sagemaker_session)
+    model.compile(
+        target_instance_family=None,
+        input_shape={"data": [1, 3, 1024, 1024]},
```
mchoi8739, line 105: input_shape={'data': [1, 3, 1024, 1024]}

^ suggesting to use single-quote (based on what we discussed earlier)

src/sagemaker/model.py
```diff
@@ -215,6 +215,10 @@ def _compilation_job_config(
         job_name,
         framework,
         tags,
+        target_platform_os=None,
+        target_platform_arch=None,
+        target_platform_accelerator=None,
+        compiler_options=None,
```
mchoi8739, line 221: Maybe we want to have the doc strings here in `model.compile`. Need to point out that the target_instance_family param and the target_platform_* params are not usable together.

src/sagemaker/model.py
mchoi8739: The API reference is updated.
[OutputConfig API](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_OutputConfig.html)
[TargetPlatform API](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_TargetPlatform.html)

src/sagemaker/model.py
mchoi8739: The API reference is updated.
OutputConfig API
TargetPlatform API

src/sagemaker/model.py
mchoi8739, line 385: We may need to point out here as well that `target_instance_family` and `target_platform_*` are not usable together?

tests/unit/sagemaker/model/test_neo.py
apivovarov: According to `model.py` code it should be `True` only if `target_instance_family` starts with `ml_` and `inf_`. Instances supported by Sagemaker.

src/sagemaker/estimator.py
apivovarov: added

src/sagemaker/model.py
apivovarov: Replaced with `ValueError`

src/sagemaker/model.py
apivovarov: fixed as suggested

src/sagemaker/model.py
apivovarov: I left just one example value and added link to allowed values.

src/sagemaker/model.py
apivovarov: Fixed, now it support both `dict` and `string`. Similar to `DataInputConfig`

apivovarov: For each `target_platform_*` field documentation I added - `It can be used instead of target_instance_family.`

src/sagemaker/model.py
apivovarov: Can we keep old comment? I like it better.

src/sagemaker/model.py
apivovarov: Can we keep old comment? I like it better.

tests/unit/sagemaker/model/test_neo.py
apivovarov: Yes, we can use single quotes here. Fixed

tests/unit/sagemaker/model/test_neo.py
apivovarov: fixed

src/sagemaker/model.py
laurenyu: "not supported to deploy" is the part that was bothering me. maybe "not supported for deployment"?

src/sagemaker/model.py
```diff
@@ -236,10 +244,32 @@ def _compilation_job_config(
...
+            }
+            if target_platform_accelerator is not None:
+                target_platform["Accelerator"] = target_platform_accelerator
+            output_model_config["TargetPlatform"] = target_platform
+
+        if compiler_options is not None:
+            output_model_config["CompilerOptions"] = (
+                compiler_options
+                if not isinstance(compiler_options, dict)
+                else json.dumps(compiler_options)
```
laurenyu, line 270: let's flip this around to put the positive condition first, i.e. `json.dumps(compiler_options) if isinstance(compiler_options, dict) else compiler_options`

src/sagemaker/model.py
apivovarov: ok, changed to `not supported for deployment`.

src/sagemaker/model.py
apivovarov: changed to `not supported for deployment`

src/sagemaker/model.py
apivovarov: fixed here and for `DataInputConfig` as well.

PR #1901, opened 2020-09-17
Add more aliases in XGBoost versions

iyerr3:

#### General

#### Tests


ChuyangDeng (approved):

src/sagemaker/image_uri_config/xgboost.json
```diff
@@ -1,7 +1,9 @@
 {
     "scope": ["inference", "training"],
     "version_aliases": {
-        "latest": "1"
+        "latest": "1", 
+        "0.90": "0.90-2", 
+        "1.0": "1.0-1" 
     },
     "versions": {
```
ChuyangDeng, line 8: When you update version_aliases, you will also need to update versions with the corresponding regional accounts, py_versions, repository name etc.

ChuyangDeng (changes requested):

src/sagemaker/image_uri_config/xgboost.json
iyerr3: I'm confused on this point. I interpret the `version_aliases` as a means to allow users type the alias instead of the full version (e.g input `1.0` instead of `1.0-1`). Adding a new alias should not affect the destination version number, isn't it? In this case, `0.90-2` and `1.0-1` remain the same as before.

src/sagemaker/image_uri_config/xgboost.json
```diff
@@ -1,7 +1,9 @@
 {
     "scope": ["inference", "training"],
     "version_aliases": {
-        "latest": "1"
+        "latest": "1", 
+        "0.90": "0.90-2",
```
edwardjkim, line 5: In SDK v1, `0.90` pointed to `0.90-1` not `0.90-2`, because 1) 0.90 was released before we finalized our versioning story, and 2) the alias gets created when `-1` version is released, so for backwards compatibility, we are forced to alias the bare xgboost version to the very first container version. For backwards compatibility, the alias `0.90` needs to be pointed to `0.90-1` forever even if there is a `0.90-2`. With the major version bump to SDK v2, this may be okay for `0.90` or `1.0`, but this is why I don't think it's best to use the xgboost versions as aliases.

Issue #1977, opened 2020-10-29
[Feature] Offer accept argument in all Deserializers


athewsey:
**Describe the feature you'd like**

Ability to customize `Accept` header for as many `Deserializer`s as possible, but **in particular** the "identity" deserializers like `BytesDeserializer` and `StreamDeserializer`.

Sometimes we need to explicitly specify an `Accept`, but don't want the SageMaker SDK to do any processing beyond loading the bytestream.

In my particular case, I noticed that SDKv2 is materially harder to use than v1 with the built-in Semantic Segmentation algorithm - where we used to be able to simply:

```python
# SDK v1 code:
ss_predictor.content_type = 'image/jpeg'
ss_predictor.accept = 'image/png'
return_img = ss_predictor.predict(img)
```

Now I had to do the below to get the code working:

```python
# Eww:
class MyPNGDeserializer(sagemaker.deserializers.BytesDeserializer):
    ACCEPT = ('image/png',)

ss_predictor.serializer = sagemaker.serializers.IdentitySerializer('image/jpeg')
ss_predictor.accept = MyPNGDeserializer()
return_img = ss_predictor.predict(img)
```

(Need to specify the Accept type for this algorithm, because default appears not to be PNG)

**How would this feature be used? Please describe.**

Directly re-use existing deserializers while being able to specify particular `Accept`:

```python
dictor.serializer = sagemaker.serializers.IdentitySerializer('image/jpeg')
ss_predictor.accept = sagemaker.deserializers.BytesDeserializer('image/png')
return_img = ss_predictor.predict(img)
```

**Describe alternatives you've considered**

Perhaps this is a deliberate omission, to encourage users to get familiar with implementing proper Deserializers rather than doing it after the `predictor.predict()` call? I could appreciate the rationale for that, it just doesn't feel very beginner-friendly.

**Additional context**

N/A

metrizable:
Hello @athewsey ,
    
Thank you for using Amazon SageMaker. Great suggestion. When I read the beginning of your issue, I thought the same thing myself re: passing in an `accept` to the constructor or, say, an associated `classmethod`.

Your current use of the implementation is correct, although the UX, as you noted, can objectively be improved. We are always re-evaluating our backlog of features based on customer requests, so we appreciate the feedback on this feature.

ajaykarpur:
Merged in https://github.com/aws/sagemaker-python-sdk/pull/1993

PR #1993, opened 2020-11-19
feature: all de/serializers support content type

athewsey:
**Issue #, if available:** #1977

**Description of changes:**

Add support for `content_type` constructor argument in all provided `Serializer`s, and for `accept` constructor argument in all provided `Deserializer`s.

This change will reduce the frequency with which users need to create custom de/serializers, by increasing the flexibility of the pre-provided ones.

For example: consider the built-in [SageMaker Semantic Segmentation algorithm](https://docs.aws.amazon.com/sagemaker/latest/dg/semantic-segmentation.html#semantic-segmentation-inputoutput), where `BytesDeserializer` may be appropriate for receiving an `image/png` result - but the `Accept` header must explicitly indicate that this result is preferred rather than an `application/x-recordio-protobuf`.

I've strived for maximal backward compatibility in the proposed changeset, and believe the change is non-breaking:

- The `SimpleBase*` classes are introduced as a compromise between wanting to keep code DRY / implementing the API exactly once, versus not wanting to change the advertised `Base*` APIs.
- The new arguments are always optional, and appended to existing lists where necessary in case anybody's been using existing kwargs by position.
- Although I don't really like the ambiguity of implementing constant-looking readonly props `ACCEPT` and `CONTENT_TYPE` backed by non-private-marked instance variables `accept` and `content_type` (rather than e.g. `_accept`, and maybe making these lowercase names the actual read/writable properties) - I kept the pattern for consistency with existing implementations in e.g. `NumpyDeserializer` and `IdentitySerializer`.

I guess one small doc change introduced is that the `NumpyDeserializer` in particular used to carry a warning that multiple accept types may not work with legacy images. That warning is no longer present and it's now possible to assign multiple content types to this Deserializer... Maybe could consider adding some generic wording to *every* `accept` docstring?

Also added a bit of a note on the `CSVDeserializer` docstring as I only realised while making this change that the `NumpyDeserializer` supports additional content-types and is a good choice for text/csv->ndarray!

**Testing done:**

Because the change is implemented via optional constructor args, no new code paths are introduced and so no new tests have been added.

I've checked the existing unit tests all pass, and have done some anecdotal experiments with a couple of classes to check they seem to work as expected.

## Merge Checklist

#### General

- [ ] ~I have passed the region in to all S3 and STS clients that I've initialized as part of this change.~

#### Tests

- [ ] ~I have added tests that prove my fix is effective or that my feature works (if appropriate)~
- [ ] ~I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)~


athewsey:
^ Looking at the log, I think that CI job had some bigger problem than my changeset? But happy to look at fixing it if somebody disagrees!

athewsey:
OK the second one (integration) looks like it was my bad as some tests are (incorrectly I might add!) trying to set `CONTENT_TYPE` - will get on that

src/sagemaker/deserializers.py
```diff
@@ -55,17 +56,44 @@ def ACCEPT(self):
...
+        Args:
+            accept (union[str, tuple[str]]): The MIME type (or tuple of allowable MIME types) that
+                is expected from the inference endpoint (default: "*/*").
+        """
+        super(SimpleBaseDeserializer, self).__init__()
+        self.accept = accept
+
+    @property
+    def ACCEPT(self):
+        """The tuple of possible content types that are expected from the inference endpoint."""
```
icywang86rui, line 79:
```suggestion
        """Content type or the tuple of possible content types that are expected from the inference endpoint.
            
            Returns: MIME type or tuple of MIME types.
        """
```

icywang86rui (commented):
Looks good. Thanks for contributing.

src/sagemaker/deserializers.py
athewsey: But in the current implementation, `ACCEPT` property always returns a tuple, right? It's the`.accept` instance variable that could be str or Tuple[str]?

ajaykarpur (approved):

Issue #1937, opened 2020-10-02
[Request] Give warning/error when job ends in 'Stopped' rather than 'Completed'


athewsey:
**Describe the feature you'd like**

`Stopped` jobs (which could have been `Completed`) should show some kind of warning or even an error: Not just silence as they do today.

**How would this feature be used? Please describe.**

Common reasons a job might be `Stopped` rather than `Completed` include:

- The job timed out (training may or may not still have exported a checkpoint or final model)
- A custom detective control in the environment specifically terminated the job via a Stop*Job call (e.g. out of budget, security policy violation, etc)

In many such cases the job termination was **not healthy**, and in the case where the job *was* healthy, the developer must have taken explicit steps to achieve that (e.g. implementing checkpointing, etc).

Therefore the current pattern in the SDK of treating `Stopped` as a success is misleading to inexperienced users ("The `.fit()` cell ran with no errors right? Everything must be fine") or experienced users who might not realise they're working in an environment with detective controls implemented ("Why does it keep not saving the model!? I do it right there in the script!").

**Describe alternatives you've considered**

1. **Current behaviour** (Stopped == Completed)
    - Not ideal for the reasons described above, but backward-compatible
2. **print() a warning** message on 'Stopped'
    - Still easy to ignore, particularly if the job generated a lot of logs already before stopping and the warning is just added on below. Still doesn't interrupt code execution.
    - ...but simple and not breaking
3. **Raise a Python warning** on 'Stopped'
    - Nice and visible in display: IPython will render in a red box much like uncaught errors. Doesn't break existing code flows.
    - ...but default `warnings` settings are a bit weird in notebook kernels: Easy for users to have the warning set to "once", in which case it will only display the first time it's triggered - which could be even more confusing. Still doesn't interrupt code execution.
4. **Raise a specific error** on 'Stopped'
    - Breaking change in the (unusual?) case of code flows that use job timeout as standard (rather than other stopping conditions)
    - ...but sets a nice intuitive behaviour that your notebook cell will terminate nicely if your model/processing runs successfully, and error otherwise.
    - Would also not pollute logs/warnings in the event that the condition is explicitly expected and handled, which it could be easily for users who expect the condition.

(4) seems like a nice solution, so long as the logic to catch that specific error (and not `Failed`) is reasonably intuitive.

**Additional context**

[...]

PR #2000, opened 2020-11-24
feature: warn on 'Stopped' (non-Completed) jobs

athewsey:
**Issue #, if available:**  #1937

**Description of changes:**

Jobs ending with status "Stopped" rather than "Completed" now trigger a warning, because they're likely to have been terminated in a non-happy-path (e.g. aborted or stopped early).

As discussed on the associated issue, a user typically needs to have taken actual conscious steps (e.g. saving intermediate output) for a job terminating in this way to be considered successful: So warning as default behaviour should help to protect less-experienced users.

Some caveats / possible points of feedback:

- This PR implements the change directly in the shared `_check_job_status()` function because, as far as I've been able to ascertain, this significance appears to be shared between the different types of jobs it's shared between (which includes AutoML, Training, Processing, Compilation, HyperParameterTuning, Transform).
- In my search of the docs, I couldn't find anywhere that the current 'silent' behaviour was explicitly referenced to update... And since it's only a warning I didn't feel it merited adding something explicit.
- The current draft uses `LOGGER.warning()` to show a warning message, not `warnings.warn` (a pattern which is already used in some other places) - but I couldn't see any existing test tooling to assert on this: Just tests like`with pytest.warns(UserWarning):` which seem to work with the `warnings` module only. Open to the idea of using the warnings module instead, but I think the default filter behaviour would be different? (Showing once only vs showing each triggered warning).

**Testing done:**

Verified the unit tests run, and imported the modified library in SageMaker (Studio) notebook to check manually that attaching to `Stopped` jobs triggers a visible warning as expected.

## Merge Checklist

#### General

- [ ] ~I have passed the region in to all S3 and STS clients that I've initialized as part of this change.~
- [ ] ~I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)~

#### Tests

- [ ] ~I have added tests that prove my fix is effective or that my feature works (if appropriate)~
- [ ] ~I have checked that my tests are not configured for a specific region or account (if appropriate)~
- [ ] ~I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)~


athewsey:
Is there a way to re-kick the CI checks without adding more commits? Seems like there was some sort of availability problem

ajaykarpur (approved):

PR #2025, opened 2020-12-10
fix: Use CUDA 11 DLC images for custom profiler config

ndodda-amazon:
*Issue #, if available:*

*Description of changes:*
If a profiler config is specified in the estimator while the framework version and python version are specified, default to using the CUDA 11 images.

*Testing done:*
Ran the unit tests locally.

## Merge Checklist

#### General

#### Tests


src/sagemaker/image_uris.py
```diff
@@ -253,7 +258,7 @@ def _should_auto_select_container_version(instance_type, distribution):
     if distribution:
         smdistributed = "smdistributed" in distribution
 
-    return p4d or smdistributed
+    return p4d or smdistributed or has_custom_profiler_config
```
metrizable, line 261: This is a one-way door. Just confirming that rather than support profiler on cu10, we want to always use cu11 in this case. If people author their code for cu11, then a revert happens, it could potentially break them.

ndodda-amazon: Understood, I updated the PR description. Can we introduce a `cuda_version` parameter in the estimator so that we don't force the user to specify image URI if they want to choose a different CUDA version? Ideally all relevant DLC images should be available through the SDK without the user needing to specify the image URI.

We don't want users to have to specify image URI for CUDA 10 DLC images if they want to enable profiling using a CUDA 10 image (once we have profiling support for CUDA 10 images).

metrizable: The purpose of the SDK is to keep things simple. For the most common use cases (and the most common use cases won't satisfy every condition) three dimensions is more than enough. By including cuda as an argument, dimensionality of image uri config validation multiplies, cognitive overhead for users goes up. The SDK historically has purposely avoided inclusion of cuda (and os) to keep cognitive decisions at a minimum. Given that os is essentially the only other dimension for full image uri resolution, the expense of adding cuda is dubious.

PR #2047, opened 2020-12-19
feat: Add support for actions in debugger rules.

ndodda-amazon:
*Issue #, if available:*

*Description of changes:*
These actions allow users to specify actions along with debugger rules during training. If the rule job finds issues during training, then the actions specified by the user will be triggered. 

As of right now, support for actions is limited to debugger rules. 

These changes include a dependency on smdebug-rulesconfig==1.0.1, which includes the action classes to be used as input to `Rule.sagemaker` via the `actions` parameters.

*Testing done:*

## Merge Checklist

#### General

#### Tests


metrizable (approved):

src/sagemaker/debugger/debugger.py
```diff
@@ -268,6 +271,12 @@ def sagemaker(
                 """
             )
 
+        if actions is not None and not rule_configs.is_valid_action_object(actions):
+            raise RuntimeError(
+                """`actions` must be of type `Action` or `ActionList`!
+                """
```
metrizable, line 277:
```suggestion
                """`actions` must be of type `Action` or `ActionList`!"""
```

src/sagemaker/debugger/debugger.py
```diff
@@ -352,6 +363,12 @@ def custom(
             :class:`~sagemaker.debugger.Rule`: The instance of the custom rule.
 
         """
+        if actions is not None and not rule_configs.is_valid_action_object(actions):
+            raise RuntimeError(
+                """`actions` must be of type `Action` or `ActionList`!
+                """
```
metrizable, line 369:
```suggestion
                """`actions` must be of type `Action` or `ActionList`!"""
```

setup.py
```diff
@@ -39,7 +39,7 @@ def read_version():
     "numpy>=1.9.0",
     "protobuf>=3.1",
     "protobuf3-to-dict>=0.1.5",
-    "smdebug_rulesconfig==1.0.0",
+    "smdebug_rulesconfig==1.0.1",
```
NihalHarish, line 42: If we change the requirement to `>=` can we avoid the need to submit a PR every time we update rulesconfig?

metrizable (changes requested):

setup.py
ndodda-amazon: Understood but the point of pinning a version is so that we submit a PR and verify that the sagemaker SDK tests pass with the new version of rulesconfig. This accomplishes two things:
- If a rulesconfig release occurs and it turns out to be incompatible with the current sagemaker SDK version, then the sagemaker SDK functionality won't be broken immediately (because the version is pinned to an older version of rulesconfig). If sagemaker SDK changes are required with a new rulesconfig release, we mandate that the rulesconfig gets released first, then make a new PR in sagemaker SDK with the changes and bump in rulesconfig version (like this PR).
- A new sagemaker SDK release should happen for each rulesconfig update, regardless of whether SDK changes are required for the rulesconfig update. We don't want users to have any direct interaction with rulesconfig, installing the latest sagemaker SDK should always install the latest rulesconfig. If we use `>=`, user may have to manually reinstall latest rulesconfig even with latest sagemaker SDK.

src/sagemaker/debugger/debugger.py
ndodda-amazon: Done.

src/sagemaker/debugger/debugger.py
ndodda-amazon: Done.

setup.py
NihalHarish: -- For point 1; nothing stops us from reversing this logic if there is a need to update the sdk before releasing rulesconfig.

-- For point 2; How will this scenario occur? Every-time a the sagemaker sdk is built; it will pull the latest rulesconfig ensuring that it is always built with the latest rulesconfig.

ndodda-amazon: > -- For point 1; nothing stops us from reversing this logic if there is a need to update the sdk before releasing rulesconfig.

Why would there be a need to update the SDK before releasing rulesconfig?

> -- For point 2; How will this scenario occur? Every-time a the sagemaker sdk is built; it will pull the latest rulesconfig ensuring that it is always built with the latest rulesconfig.

If the user installed sagemaker SDK before a rulesconfig release. Then to have the latest rulesconfig features, they would have to manually install the latest version of rulesconfig, instead of the latest version of sagemaker (which is pinned to latest version of rulesconfig). So, user should have no direct interaction with rulesconfig.

metrizable (commented):
looks good! can you add an integ test for the new actions support?

metrizable (approved):
looks good! thanks for the contribution!

PR #2071, opened 2021-01-06
change: add a 1 second sleep for List operations when a LimitExceededException is thrown

yzhu0:
Add throttle to list methods

### DIFF
--- a/src/sagemaker/apiutils/_base_types.py
+++ b/src/sagemaker/apiutils/_base_types.py
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 """Provides utilities for custom boto type objects."""
 from __future__ import absolute_import
-
+import time
 from sagemaker.apiutils import _boto_functions, _utils
 
 
@@ -143,6 +143,7 @@ class Record(ApiObject):
                     yield list_item_factory(item)
                 if not next_token:
                     break
+                time.sleep(1)
         except StopIteration:
             return
 
--- a/src/sagemaker/apiutils/_base_types.py
+++ b/src/sagemaker/apiutils/_base_types.py
@@ -12,7 +12,9 @@
 # language governing permissions and limitations under the License.
 """Provides utilities for custom boto type objects."""
 from __future__ import absolute_import
+import logging
 import time
+import botocore
 from sagemaker.apiutils import _boto_functions, _utils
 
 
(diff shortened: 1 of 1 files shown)


icywang86rui:
I am not sure this is the correct way to fix throttling problems. This will add 1 second delay in all the ListApi calls wherever this is used. This might break our customers that has a higher rate for their specific account.

icywang86rui (changes requested):
Not sure this change is the right fix.

jerrypeng7773 (dismissed):

danabens:
fix unit tests
`[Container] 2021/01/06 18:45:23 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: tox -e flake8,pylint,docstyle,black-check,twine --parallel all. Reason: exit status 1`

yzhu0:
> I am not sure this is the correct way to fix throttling problems. This will add 1 second delay in all the ListApi calls wherever this is used. This might break our customers that has a higher rate for their specific account.

since we have rate =2 under ListApi call. We should let it sleep for 1s to prevent API throttling
https://tiny.amazon.com/xh7mx9uw/IronmanUtilizationService/342d6fd1/default-rules.json#L225

danabens (approved):

danabens (commented):
Rui's comments are correct - should normally list at full speed, once there is a throttling exception list at a slower rate.

yzhu0:
> Rui's comments are correct - should normally list at full speed, once there is a throttling exception list at a slower rate.

I update the pr to sleep only when the method is throttled

danabens:
confirm tested end to end either via integ test or manually

src/sagemaker/apiutils/_base_types.py
```diff
@@ -130,19 +132,28 @@ def _list(
...
+                    if not next_token:
+                        break
+                except botocore.exceptions.ClientError as error:
+                    if error.response["Error"]["Code"] == "LimitExceededException":
+                        logging.getLogger().warning(
+                            "API call limit exceeded; backing off and retrying..."
+                        )
+                        time.sleep(1)
+                    else:
+                        raise error
```
ChoiByungWook, line 156: Can we update unit tests to exercise this line?

ChoiByungWook (changes requested):

ChoiByungWook (commented):
Would it be better to configure the retry strategy in the boto client instead of sleeping?

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

icywang86rui:
There has not been any actions for a few weeks. Closing for now. Feel free to reopen.

PR #2087, opened 2021-01-13
friendly names for short URIs

yzhu0:
Shorten the short uri as name 

Test
tox -e py38 --  tests/integ
tox -e py38 --  tests/unit


danabens:
looks like the unit test failure is related

PR #2133, opened 2021-02-10
feature: Add Framework Version support for PyTorch compilation (Neo)

prateekch91:
*Description of changes:*
Add Framework Version support for PyTorch compilation (Neo), also added image URI support for 1.5 and 1.6. Updated 1.4.0 to 1.4 in image URI for consistency.

```
Example:
model.compile(
        target_instance_family="ml_c4",
        input_shape={"data": [1, 3, 1024, 1024]},
        output_path="s3://output",
        role="role",
        framework="pytorch",
        framework_version="1.6",
        job_name="compile-model",
)
```
#### Testing
- Added test cases to test for different scenarios.
- Tested Sagemaker compile for PyTorch versions - 1.4/1.5/1.6 with new Boto changes.

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)

#### Tests

- [x ] I have added tests that prove my fix is effective or that my feature works (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2134, opened 2021-02-11
fix: disable profiler by default for regions not support it

guoren66:
In China regions (cn-north-1/cn-northwest-1), if sdk is updated from 1 to latest one and start a training job, an error occurrs:
```
ClientError: An error occurred (ValidationException) when calling the CreateTrainingJob operation: ProfilerConfig is currently not supported in cn-northwest-1.
```

Many customers have reported this error (No direct solution to it in the error log) and we have to help resolve it one by one, and this also brings potential customer's churn if we haven't covered them yet.

*Description of changes:*
add logic of judging whether region supports profiler, if not, set 'disable_profiler' to True.

*Testing done:*
Unit Testing
Integ Testing

## Merge Checklist

#### General

#### Tests


src/sagemaker/estimator.py
```diff
@@ -349,6 +349,9 @@ def __init__(
         self.profiler_config = profiler_config
         self.disable_profiler = disable_profiler
 
+        if not _region_supports_profiler(self.sagemaker_session.boto_region_name):
```
metrizable, line 352: Do you think a test exercising this branch is warranted?

guoren66: Have been passed in Unit and integ test (mxnet using default profiler confit in test profiler) from my side. Do you think it's necessary to test the branch in environment of China regions?

guoren66:
@metrizable can you help check the failed test of './amazon-sagemaker-examples/sagemaker-python-sdk/mxnet_onnx_eia/mxnet_onnx_eia.ipynb' in sagemaker-python-sdk-notebook-tests? (container not passing health check), as this PR is not for deployment.
same issue is also found in this PR https://github.com/aws/sagemaker-python-sdk/pull/2133. Maybe it's a general one.

I have tested 'amazon-sagemaker-examples/sagemaker-python-sdk/mxnet_onnx_eia/mxnet_onnx_eia.ipynb' on 2.24.5 in us-west-2, it's ok.

guoren66:
> ### AWS CodeBuild CI Report
> * CodeBuild project: sagemaker-python-sdk-notebook-tests
> * Commit ID: [7a91c3f](https://github.com/aws/sagemaker-python-sdk/commit/7a91c3f5f305d67bdc4ddc0afad12b520a78c28b)
> * Result: FAILED
> * [Build Logs](https://jr6hkjvk5d.execute-api.us-west-2.amazonaws.com/Prod/buildlogs?key=14cbdb49-b3f5-44d9-842f-4872fe847546%2Fbuild.log) (available for 30 days)
> 
> _Powered by [github-codebuild-logs](https://serverlessrepo.aws.amazon.com/applications/arn:aws:serverlessrepo:us-east-1:277187709615:applications~github-codebuild-logs), available on the [AWS Serverless Application Repository](https://aws.amazon.com/serverless/serverlessrepo/)_

@metrizable can you help check the failed test of './amazon-sagemaker-examples/sagemaker-python-sdk/mxnet_onnx_eia/mxnet_onnx_eia.ipynb' in sagemaker-python-sdk-notebook-tests? (container not passing health check), as this PR is not for deployment.
same issue is also found in this PR https://github.com/aws/sagemaker-python-sdk/pull/2133. Maybe it's a general one.

I have tested 'amazon-sagemaker-examples/sagemaker-python-sdk/mxnet_onnx_eia/mxnet_onnx_eia.ipynb' on 2.24.5 in us-west-2, it's ok.

ChoiByungWook (commented):
Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?

I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.

guoren66:
> Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?
> 
> I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.

Thanks for the concern.

I have received more than 10 cases from customers in China and the error log cannot directly show them how to handle this fail. I also worry about those who are new to SageMaker and we haven't touched directly in China and encounter this fail when start a very simple training job. It's better to change this parameter to 'False' for the region not supporting Profiler. 

I also agree with your concern about those who set this to 'True' but not enabled. So I recommend that we can add a print log inside this 'if' logic, to tell the user that this region is not supporting Profiler and we automatically set this to ‘False’.

ChoiByungWook:
> > Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?
> > I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.
> 
> Thanks for the concern.
> 
> I have received more than 10 cases from customers in China and the error log cannot directly show them how to handle this fail. I also worry about those who are new to SageMaker and we haven't touched directly in China and encounter this fail when start a very simple training job.(this may lead to user churn) It's better to change this parameter to 'False' for the region not supporting Profiler.
> 
> I also agree with your concern about those who set this to 'True' but not enabled. So I recommend that we can add a print log inside this 'if' logic, to tell the user that this region is not supporting Profiler and we automatically set this to ‘False’, to eliminate their confusion.

Is profiler on by default for all regions or is it because users are attempting to use it in an unsupported region?

guoren66:
> > > Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?
> > > I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.
> > 
> > 
> > Thanks for the concern.
> > I have received more than 10 cases from customers in China and the error log cannot directly show them how to handle this fail. I also worry about those who are new to SageMaker and we haven't touched directly in China and encounter this fail when start a very simple training job.(this may lead to user churn) It's better to change this parameter to 'True' for the region not supporting Profiler.
> > I also agree with your concern about those who set this to 'False' but not enabled. So I recommend that we can add a print log inside this 'if' logic, to tell the user that this region is not supporting Profiler and we automatically set this to ‘True’, to eliminate their confusion.
> 
> Is profiler on by default for all regions or is it because users are attempting to use it in an unsupported region?

(I correct my above comment for the value of this parameter)
Yes. The root cause is profiler is on by default for all the regions. (the 'disable_profiler' is False by default). So this PR is for changing this parameter to 'True' when in unsupported region. I recommend I can add another print in this 'if' logic to tell user the profiler is off automatically even though you configure it to 'False' manually and explicitly.

ChoiByungWook:
> > > > Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?
> > > > I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.
> > > 
> > > 
> > > Thanks for the concern.
> > > I have received more than 10 cases from customers in China and the error log cannot directly show them how to handle this fail. I also worry about those who are new to SageMaker and we haven't touched directly in China and encounter this fail when start a very simple training job.(this may lead to user churn) It's better to change this parameter to 'True' for the region not supporting Profiler.
> > > I also agree with your concern about those who set this to 'False' but not enabled. So I recommend that we can add a print log inside this 'if' logic, to tell the user that this region is not supporting Profiler and we automatically set this to ‘True’, to eliminate their confusion.
> > 
> > 
> > Is profiler on by default for all regions or is it because users are attempting to use it in an unsupported region?
> 
> (I correct my above comment for the value of this parameter)
> Yes. The root cause is profiler is on by default for all the regions. (the 'disable_profiler' is False by default). So this PR is for changing this parameter to 'True' when in unsupported region. I recommend I can add another print in this 'if' logic to tell user the profiler is off automatically even though you configure it to 'False' manually and explicitly.

Well... since it is on by default, I don't think you can discern easily whether or not a user wants to enable this feature in an unsupported region.

At least how I understand it is, if we add print statement in here, everyone in an unsupported region will get this print region, even if they aren't aware of this feature.

guoren66:
> > > > > Would it be better to throw an exception on the client side if the user attempts to use the profiler explicitly in an unsupported region?
> > > > > I would rather the user's job request fail fast if they attempt to use a feature in an unsupported region, instead of defaulting everyone off and having the job succeed. The user may get confused about why the profiler wasn't enabled in their job, even though they set the flag.
> > > > 
> > > > 
> > > > Thanks for the concern.
> > > > I have received more than 10 cases from customers in China and the error log cannot directly show them how to handle this fail. I also worry about those who are new to SageMaker and we haven't touched directly in China and encounter this fail when start a very simple training job.(this may lead to user churn) It's better to change this parameter to 'True' for the region not supporting Profiler.
> > > > I also agree with your concern about those who set this to 'False' but not enabled. So I recommend that we can add a print log inside this 'if' logic, to tell the user that this region is not supporting Profiler and we automatically set this to ‘True’, to eliminate their confusion.
> > > 
> > > 
> > > Is profiler on by default for all regions or is it because users are attempting to use it in an unsupported region?
> > 
> > 
> > (I correct my above comment for the value of this parameter)
> > Yes. The root cause is profiler is on by default for all the regions. (the 'disable_profiler' is False by default). So this PR is for changing this parameter to 'True' when in unsupported region. I recommend I can add another print in this 'if' logic to tell user the profiler is off automatically even though you configure it to 'False' manually and explicitly.
> 
> Well... since it is on by default, I don't think you can discern easily whether or not a user wants to enable this feature in an unsupported region in the current if statement. You would have to do it where the defaults happen.
> 
> At least how I understand it is, if we add print statement in here, everyone in an unsupported region will get this print region, even if they aren't aware of this feature.
> 
> If the feature doesn't exist in that region, then disabling it by default seems fine. I am wondering why we enabled this feature by default and didn't make it an opt-in feature at release.

I totally agree with you about why we enable this feature by default. I don't know the specific reason maybe have some other considerations. So in this PR I propose the commits after the default setting for unsupported regions, instead of changing the default parameter from 'False' to 'None' in the init function.

Changing the default opt is an important issue I don't think we can do it in this PR.

ChoiByungWook (approved):

PR #2146, opened 2021-02-15
fix: mask creds from docker commands in local mode. Closes #2118

### LINKED ISSUE #2118: Mask creds in docker compose file for local mode
Currently, [line 673](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/local/image.py#L673) in `sagemaker/local/image.py` logs AWS creds in the docker compose file. This could cause a credentials leak if it is running in a ci pipeline, during presentations/demos and for live streamers or youtubers creating sagemaker content. 

I don't feel this is a outright security issue, just unexpected logging that could have adverse impact in certain situations. 

One has to explicitly provide fake creds and local data to get around this. This limits the ability to use data from s3 and use other aws services in jobs.

Credentials should be masked in the logs to prevent a possible leak. This could be accomplished with a regex replace of the yaml content.  I'm happy to submit a PR for this, but wanted to get thoughts because there may be better alternatives to regex. I'm not a fan of regex but it was the best I could come up with since the container needs the creds.

Currently patching out this logging in a ci pipeline.

jmahlik:
*Issue #, if available:*
#2118 

*Description of changes:*
Mask credentials in logging of docker compose commands. Change env vars to "[Masked]".
Pipe stderr of docker login to `subprocess.DEVNULL` so creds are not printed to error logs.

*Testing done:*
Add caplog to train and serve to ensure they contain "[Masked]". 

## Merge Checklist

#### General

- [ x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [x ] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)
- [ x] I have passed the region in to all S3 and STS clients that I've initialized as part of this change.
- [x ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)

#### Tests

- [x ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ x] I have checked that my tests are not configured for a specific region or account (if appropriate)
- [x ] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)


jmahlik:
Let me know if anything needs changing or there are better ideas. I was on a windows machine so had limited ability to pass the full test suite. I ran the tests for `tests/unit/test_image.py`, the altered tests passed.

Side note, another scenario where this came up: commiting AWS credentials to git if notebook output is not cleared when using local mode. This is probably the most impactful situation this PR will fix.

jmahlik:
Looks like there's quite a few unrelated pylint errors that already existed. Not sure how to proceed.

src/sagemaker/local/image.py
```diff
@@ -1067,7 +1074,7 @@ def _ecr_login_if_needed(boto_session, image):
     ecr_url = auth["authorizationData"][0]["proxyEndpoint"]
 
     cmd = "docker login -u AWS -p %s %s" % (token, ecr_url)
-    subprocess.check_output(cmd.split())
+    subprocess.check_output(cmd.split(), stderr=subprocess.DEVNULL)
```
ChoiByungWook, line 1077: Why do we set stderr to DEVNULL?

jmahlik: [Suppressing the error](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) if it errors. It will error if docker isn't configured properly ex. behind a proxy but don't have docker set up for it. 

The error has possibly sensitive creds in it. It was printing to build logs and triggering credential scanners. Also, might cause a notebook with the creds output to get commited.

I'm open to other ideas of how to suppress it.

ChoiByungWook: Gotcha,

That makes a lot of sense... I'm wondering if suppressing the error will cause issues for users to quickly debug/figure out why something failed.

Would it be possible to modify what is being sent to stderr? What are your thoughts?

jmahlik: Good thought. 

Initially, I thought piping it might be a good idea but the docs recommend against using pipes with check_call. We could switch to using `Popen(cmd).communicate() ` with pipes, manually check the outputs and raise etc... 

It might be better to see if there's a different docker command to use for this? The only alternative I'm privy to is `["aws", "ecr", "get-login-password"]` (that's what I patch this out with currently). It relies on users having the ecr credentials helper, which I do not believe is configured on notebook instances. For studio, that might be even more of a pain to expect it to be in all images.

Maybe this belongs in a separate PR?

ChoiByungWook: Ahhh this solution looks like it requires a bit more time to figure out properly.

I agree, thank you so much for your efforts!

Can we remove `stderr=subprocess.DEVNULL` for now?

jmahlik: Just pushed it. I'll open a separate issue for discussion.

ChoiByungWook (approved):
Awesome!!!

Issue #2197, opened 2021-03-09
Workflow ProcessingStep run arguments


jaggaer-mtrninic:
**Describe the feature you'd like**
ProcessingStep at the moment as per my knowledge doesn't cover all the arguments that run function of concrete Processor has. 
Feature would be ability to pass in an run arguments for concrete processor from the ProcessingStep object.

**How would this feature be used? Please describe.**
For example PySparkProcessor has a submit_jars argument but this can not be passed to ProcessingStep as part of the pipeline. 
It would be nice if there could be a run arguments for ProcessingStep where we could client pass the arguments of concrete Procesor.

**Describe alternatives you've considered**
For PySpark processor and submit jars only solution in my mind is to rewrite the code in Scala and make an uber jar with all jars required for the job.

jerrypeng7773:
https://github.com/aws/sagemaker-python-sdk/issues/2857

jerrypeng7773:
Reprinted from the answer for #2857 

@jaggaer-mtrninic We recently released a SDK change to support this. 

```
from sagemaker.workflow.pipeline_context import PipelineSession

session = PipelineSession()

processor = FrameworkProcessor(...)

step_args = processor.run(inputs=..., source_dir="...")

step_sklearn = ProcessingStep(
    name="MyProcessingStep",
    step_args=step_args,
)
```

In summary, we introduced the `PipelineSession`. This special session does not trigger a processing job immediately when you call `processor.run` , instead, it captures the request arguments required to run a processing job, and delegate it to the processing step to start the job later during pipeline execution. 

This works for estimator, transformer, and tuner too.

You should be able to use this with the latest sdk release 2.89.0

jerrypeng7773:
close for now, feel free to re-open if you have any other concern.

PR #2209, opened 2021-03-13
feature: Support all processor types in ProcessingStep

staubhp:
*Issue #, if available:*
https://github.com/aws/sagemaker-python-sdk/issues/2197

*Description of changes:*
SageMaker ML Pipelines was launched at reInvent 2020. These pipelines are composed of steps representing various SageMaker jobs/actions, and are built using the SageMaker Python SDK.

The ProcessingStep constructor mostly matches the `run()` method on the `ScriptProcessor` class. This works for processors that inherit `ScriptProcessor` directly like `SKLearnProcessor`, but presents a problem for processors that have an additional layer of abstraction, like `PySparkProcessor`. The latter take special arguments in their `run()` method which are transformed in that method and passed to their parent class. Because of this disconnect, users don't have an easy way of passing arguments for such processors to `ProcessingStep`.

This change adds a `get_run_args()` method to the `ScriptProcessor` base class which subclasses can override. This method takes the same arguments as the processor's `run()` method and does the necessary transformation, returning a `RunArgs` object with properties matching those on the `ProcessingStep` constructor.

*Testing done:*
This change adds unit tests and integ tests for the new method. In addition, it was tested manually in a local notebook.

## Merge Checklist

#### General

#### Tests


src/sagemaker/processing.py
```diff
@@ -442,6 +442,34 @@ def __init__(
             network_config=network_config,
         )
 
+    def get_run_args(
```
ChoiByungWook, line 445: Does this method need to be an abstract method?

ChoiByungWook (commented):
LGTM!

I am mainly concerned about ensuring that extending classes implement their additional layer. I'm wondering if an [abstractmethod](https://docs.python.org/3/library/abc.html) might be the best solution for this?

However, looking at this base class, it looks like there isn't an abstract base class, so it might be a lot more work. So I am okay with this default method definition.

src/sagemaker/spark/processing.py
```diff
@@ -738,7 +837,6 @@ def run(
                 user code file (default: None).
         """
         self._current_job_name = self._generate_current_job_name(job_name=job_name)
-        self.command = [_SparkProcessorBase._default_command]
```
ChoiByungWook, line 741: Is this technically a breaking change?

ChoiByungWook: Oh nevermind. I see it is still called.

src/sagemaker/workflow/steps.py
```diff
@@ -320,6 +320,7 @@ def __init__(
         code: str = None,
         property_files: List[PropertyFile] = None,
         cache_config: CacheConfig = None,
+        kms_key=None,
```
ChoiByungWook, line 323: integ test?

src/sagemaker/processing.py
```diff
@@ -442,6 +442,34 @@ def __init__(
...
+        self,
+        code=None,
```
icywang86rui, line 447: Is this optional?

src/sagemaker/processing.py
```diff
@@ -1144,6 +1172,41 @@ def _to_request_dict(self):
...
+    """Accepts parameters that correspond to ScriptProcessors.
+
+    An instance of this class is returned from the ``get_run_args()`` method on processors,
+    and is used for normalizing the arguments so that they can be passed to
+    :class:`~sagemaker.workflow.steps.ProcessingStep`
+    """
+
+    def __init__(
+        self,
+        code=None,
```
icywang86rui, line 1185: Is this optional?

src/sagemaker/spark/processing.py
```diff
@@ -685,6 +718,72 @@ def __init__(
...
+        spark_event_logs_s3_uri=None,
+    ):
+        """Returns a RunArgs object.
+
+        This object contains the normalized inputs, outputs
+            and arguments needed when using a ``PySparkProcessor``
+            in a :class:`~sagemaker.workflow.steps.ProcessingStep`.
+
+        Args:
+            submit_app (str): Path (local or S3) to Python file to submit to Spark
```
icywang86rui, line 741: Can we just call this `code` to be consistent?

src/sagemaker/spark/processing.py
```diff
@@ -685,6 +718,72 @@ def __init__(
...
+                be published to.
+        """
+        self._current_job_name = self._generate_current_job_name(job_name=job_name)
+
+        if not submit_app:
+            raise ValueError("submit_app is required")
+
+        extended_inputs, extended_outputs = self._extend_processing_args(
+            inputs,
+            outputs,
```
icywang86rui, line 772: nit - always pass named args.

src/sagemaker/processing.py
```diff
@@ -1144,6 +1172,41 @@ def _to_request_dict(self):
         return s3_output_request
 
 
+class RunArgs(object):
```
icywang86rui, line 1175: This seems like a good class to just use `attr` to define

```suggestion
class RunArgs(object):
    code = attr.ib()
    inputs = attr.ib(default=None)
    outputs = attr.ib(default=None)
    arguments = attr.ib(default=None)
```

src/sagemaker/processing.py
staubhp: I tried but don't see a way of making this work.

The class inheritance looks like this: `Processor>ScriptProcessor>_SparkProcessorBase>PySparkProcessor`

ScriptProcessor can't be an abstract class because it's instantiated and used on its own in some cases. I can make the Processor class abstract and give it an abstract `get_run_args()`, but that only requires `ScriptProcessor` to implement it, I see no way of forcing subclasses (ie, _SparkProcessorBase, PySparkProcessor) to implement an abstract method for the super of its super. Correct me if I'm wrong. 

Moreover, there are cases like SKLearn where it doesn't need to be implemented. by the subclass. Its inheritance is `Processor>ScriptProcessor>SKLearnProcessor`, so they `get_run_args()` on `ScriptProcessor` works fine for it.

src/sagemaker/workflow/steps.py
staubhp: Let me remove this for the time being. It's an additional miss on the ProcessingStep as you can't specify a KMS key for the code uploaded by `_normalize_inputs`, but it's not related to the problem this PR is solving. I'll make a new PR for this.

src/sagemaker/spark/processing.py
staubhp: I'd prefer this match the signature of the processor's `run()` method, which also takes `submit_app` which eventually becomes `code` in the ScriptProcessor. Users previously expressed confusion about this, they weren't sure what "code" meant in the context of a Spark processor.

src/sagemaker/spark/processing.py
staubhp: Ack

src/sagemaker/processing.py
staubhp: It is not, let me fix that.

icywang86rui (approved):

ahsan-z-khan (approved):

src/sagemaker/spark/processing.py
guoqiaoli1992: +1 using submit_app, but could you add reference to RunArgs.code here

staubhp: Will do

ahsan-z-khan (approved):

PR #2264, opened 2021-04-01
fix: Disable debugger when checkpointing is enabled with distributed training

NihalHarish:
*Issue #, if available:*

*Description of changes:*
- update `_validate_and_set_debugger_configs` to disable debugger when checkpointing is enabled with distributed training

*Testing done:*
- Added new unit test

## Merge Checklist

#### General

#### Tests


src/sagemaker/estimator.py
```diff
@@ -2219,7 +2219,20 @@ def _validate_and_set_debugger_configs(self):
...
         elif not self.debugger_hook_config:
-            self.debugger_hook_config = None
+            self.debugger_hook_config = False
+
+        # Disable debugger if checkpointing is enabled by the customer
+        _should_disable_debugger = False
+        if self.checkpoint_s3_uri and self.checkpoint_local_path and self.debugger_hook_config:
+            if self.instance_count > 1:
+                _should_disable_debugger = True
+            if hasattr(self, "distribution") and self.distribution is not None:
```
ndodda-amazon, line 2229: Why can't we just consolidate all of this in one if check?

```
if self.instance_count > 1 or (hasattr(self, "distribution") and self.distribution is not None):
    logger.info(
                "SMDebug Does Not Currently Support Distributed Training Jobs With Checkpointing Enabled"
            )
    self.debugger_hook_config = False
```

src/sagemaker/estimator.py
```diff
@@ -2219,7 +2219,20 @@ def _validate_and_set_debugger_configs(self):
         ):
             self.debugger_hook_config = DebuggerHookConfig(s3_output_path=self.output_path)
         elif not self.debugger_hook_config:
-            self.debugger_hook_config = None
+            self.debugger_hook_config = False
```
ndodda-amazon, line 2222: Why are we changing this variable to `False`? Is this intentional? If so, why?

ndodda-amazon (changes requested):
Changes look good, left some comments

src/sagemaker/estimator.py
NihalHarish: I did this for readability.

src/sagemaker/estimator.py
NihalHarish: The intent of the line above is to disable the debug hook config.
`False` is a more explicit indicator of that than `None

tests/integ/test_debugger.py
```diff
@@ -351,6 +353,95 @@ def test_mxnet_with_debugger_hook_config(
...
+            sagemaker_session=sagemaker_session,
+            debugger_hook_config=debugger_hook_config,
+            checkpoint_local_path="/opt/ml/checkpoints",
+            checkpoint_s3_uri=os.path.join(s3_output_path, "checkpoints"),
+        )
+        mx._prepare_for_training()
+        # Debug Hook should be disabled
+        assert mx.debugger_hook_config is False
+
+        # Estimator with checkpointing enabled and Model Parallel Enabled
```
ndodda-amazon, line 407: nit: comment is wrong, this is Data Parallel

ndodda-amazon (approved):
LGTM! Left a minor comment but otherwise changes look good.

tests/integ/test_debugger.py
NihalHarish: Nice catch, fixed.

ndodda-amazon (approved):

src/sagemaker/estimator.py
```diff
@@ -2219,7 +2219,20 @@ def _validate_and_set_debugger_configs(self):
         ):
             self.debugger_hook_config = DebuggerHookConfig(s3_output_path=self.output_path)
         elif not self.debugger_hook_config:
```
ahsan-z-khan, line 2221: Just to understand it, so it's possible for `self.debugger_hook_config` to be `None` or `False`. And you're changing it to `False` intentionally to disable it?

NihalHarish: The default value of `self.debugger_hook_config` is None.
The if-statement above checks if the debugger is supported in the region in which it is invoked and if not it should be set to `False` instead of `None` in my opinion.

tests/integ/test_debugger.py
```diff
@@ -351,6 +354,115 @@ def test_mxnet_with_debugger_hook_config(
...
+        pt._prepare_for_training()
+        # Debug Hook should be disabled
+        assert pt.debugger_hook_config is False
+
+        # Estimator with checkpointing enabled and SMModelParallel Enabled
+        tf = TensorFlow(
+            base_job_name="tf-smdataparallel-mnist",
+            entry_point=script_path,
+            role="SageMakerRole",
+            framework_version="2.3.1",
```
ahsan-z-khan, line 433: Any reason not to use the latest TF version here? (2.4.1/2.3.2) As we are using the latest PyTorch version

src/sagemaker/estimator.py
ajaykarpur: `None` is already a falsy value in Python.

It doesn't make sense to change this to a boolean, since the `debugger_hook_config` variable can hold an actual config dict. You can simply check if `debugger_hook_config` is `None` instead of checking if it's `False`.

tests/integ/test_debugger.py
NihalHarish: There's no reason. In fact this only verifies the sanity of the estimator and does not run training

src/sagemaker/estimator.py
NihalHarish: [Documentation](https://sagemaker.readthedocs.io/en/stable/amazon_sagemaker_debugger.html#default-behavior-and-opting-out) however indicates to customers that they need to pass `False` to opt-out of debugger.

`None` is the values of this variable before we reach this code-block.

I'm making sure that when we check the value of post the `_validate` function call, we're indicative that we want debugger off instead of `None` which might come to mean that it has not yet been initialized.

tests/integ/test_debugger.py
NihalHarish: I have updated the version

src/sagemaker/estimator.py
ajaykarpur: Yes, you're right: https://github.com/aws/sagemaker-python-sdk/blob/b66cb98b9223b437b02973197d62511776ccd9cb/src/sagemaker/tensorflow/estimator.py#L361-L365

Thank you for clarifying.

ajaykarpur (approved):

ahsan-z-khan (approved):

PR #2518, opened 2021-07-13
feature: Add `LamdaModel` and `LambdaPredictor`

bveeramani:
*Issue #, if available:*

*Description of changes:*
- Added `ModelBase`
- Added `PredictorBase`
- Changed `Predictor` base class to `PredictorBase`
- Changed `Model` base class to `ModelBase`
- Added `LambdaModel`
- Added `LambdaPredictor`

*Testing done:*
Unit tests and an integration test.

## Merge Checklist

#### General

#### Tests


shreyapandit (changes requested):

tests/data/serverless/classes.txt
```diff
@@ -0,0 +1,1000 @@
+tench
+goldfish
+great white shark
+tiger shark
+hammerhead
+electric ray
+stingray
```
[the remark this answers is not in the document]
bveeramani: Yeah. The container uses a pre-trained ResNet model. The pre-trained model has been trained on ImageNet, with has  1000 classes. If we were to remove classes from the list, then the model would not classify images correctly. 

An alternative would be to manually train the model on a dataset with fewer classes (e.g., CIFAR). However, given that ImageNet is an extremely standard dataset, I don't think that's necessary.

src/sagemaker/serverless/model.py
```diff
@@ -0,0 +1,90 @@
...
+            Timeout=timeout,
+            MemorySize=memory_size,
+        )
+
+        if not wait:
+            return LambdaPredictor(function_name, client=self._client)
+
+        # Poll function state.
+        polling_interval = 5
+        while response["State"] == "Pending":
```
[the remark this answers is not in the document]
bveeramani: I disagree that moving the polling interval to a module-level constant would improve reusability and readability.
1. The polling interval is specific to the deploy method, so it would not be used elsewhere in the module. 
2. Variables should generally be defined close to where they're used; if we move the variable to the top of the module, it would be far away from where it's used, and that might hurt readability.
3. On a related note, if we define the polling interval at the top of the module, it's unclear how that constant is used until much later (unless we add a note or name it something verbose like `DEPLOYMENT_POLLING_INTERVAL`)

I don't feel as strongly about the Lambda state constants, but I'm thinking it might be better not to use module-level constants fom them too.

1. I don't think `response["State"] == LAMBDA_PENDING_STATE` is clearer than `response["State"] == "Pending"`, although I do feel the former is more verbose.
2. The name of Lambda states won't ever change (e.g., the active state will always be "Active"), so we don't need to concern ourselves with changing the value of the constant (i.e., maintainability isn't that big of a factor).

tests/integ/test_serverless.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+
+from sagemaker.serverless import LambdaModel
+from sagemaker.utils import unique_name_from_base
+
+# See tests/data/serverless for the image source code.
+IMAGE_URI = "142577830533.dkr.ecr.us-west-2.amazonaws.com/serverless-integ-test:latest"
+ROLE = "arn:aws:iam::142577830533:role/lambda_basic_execution"
+URL = "https://c.files.bbci.co.uk/12A9B/production/_111434467_gettyimages-1143489763.jpg"
```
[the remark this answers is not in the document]
bveeramani: I'm unsure how to do this. Can you link me to documentation so I can learn more?

src/sagemaker/model.py
```diff
@@ -805,6 +823,8 @@ def delete_model(self):
             )
         self.sagemaker_session.delete_model(self.name)
 
+    destroy = delete_model
```
apogupta2018, line 826: Why not name the abstract method delete_model? Same for the predictor.

src/sagemaker/serverless/predictor.py
```diff
@@ -0,0 +1,81 @@
...
+        This function is synchronous. It will only return after the function
+        has produced a prediction.
+
+        Arguments:
+            data: The data sent to the Lambda function as input.
+
+        Returns:
+            The data returned by the Lambda function.
+        """
+        response = self._client.invoke(
```
apogupta2018, line 54: Did we decide not to support any sort of error handling even for retryable exceptions?

tests/unit/sagemaker/serverless/test_model.py
```diff
@@ -0,0 +1,52 @@
...
+IMAGE_URI = "123456789012.dkr.ecr.us-west-2.amazonaws.com/my-lambda-image:latest"
+ROLE = "arn:aws:iam::123456789012:role/MyLambdaExecutionRole"
+
+
+@pytest.fixture
+def mock_client():
+    return Mock()
+
+
+@pytest.mark.parametrize("wait", [False, True])
```
apogupta2018, line 29: How about adding a test for when the deployment fails?

src/sagemaker/serverless/predictor.py
bveeramani: Yeah, that's correct. If there's an error, we'll just pass up whatever Boto throws

src/sagemaker/model.py
bveeramani: Conciseness. We could use `Model.delete_model` and `Predictor.delete_predictor`, but IMO that's unnecessarily verbose. 

Ultimately, I think this issue is sort-of bike shedding. If you feel strongly about this suggestion (or if there are other factors I'm not considering), I'm happy to change it. Otherwise, we can just disagree and commit

tests/unit/sagemaker/serverless/test_model.py
bveeramani: Good idea. I'll do so

src/sagemaker/model.py
apogupta2018: So its destroy vs delete_model, I wouldn't say its verbose, more explicit rather. But more than anything because we choose to have a different name, it just adds another lookup in the code "destroy = delete_model".

tests/integ/test_serverless.py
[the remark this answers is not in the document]
shreyapandit: Pointing this test to a hosting link external to Amazon results in a dependency and point of failure which is outside our control (the test would start to fail if that image is taken down and we won't be notified for the same). I was suggesting hosting something on Amazon Corp Drive or similar, where we have better control over the image being used

src/sagemaker/model.py
shreyapandit: I do think that while too long variable names are a code smell, given the context, in this scenario delete_model offers itself to better readability without trading the verbosity too much.

src/sagemaker/serverless/model.py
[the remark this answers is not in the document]
shreyapandit: 👍  We can revisit this when/if the file grows bigger

tests/data/serverless/classes.txt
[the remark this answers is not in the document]
shreyapandit: Makes sense, thanks for clarifying.

src/sagemaker/serverless/model.py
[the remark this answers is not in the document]
shreyapandit: To add, Regards point (2) in the first section, yes agreed with defining variable names close to where they are used, however as this codebase will grow, it will become harder to find these values buried among the relevant sections in case we need to change it. Referring to [YAGNI](https://martinfowler.com/bliki/Yagni.html), I do think we can revisit this when/if the file grows bigger 👍

src/sagemaker/model.py
bveeramani: Yeah, that makes sense to me. I hadn't really considered those as much. I'll make the change

tests/unit/sagemaker/serverless/test_model.py
bveeramani: Addressed

src/sagemaker/model.py
bveeramani: Addressed

apogupta2018 (approved):
LGTM. Merge after the slow-tests pass.

shreyapandit (approved):

PR #2533, opened 2021-07-22
fix: update `sagemaker.serverless` integration test

bveeramani:
*Issue #, if available:*

*Description of changes:*
`"CODEBUILD_BUILD_ID" not in os.environ` was evaluating `True` during builds, so the test was getting skipped.

Made two fixes:
- Stored cat image in S3 (it was previously hosted at some third-party website)
- Changed the condition for skipping the test

Note that this branch is based on the branch `bveeramani:rename-delete-endpoint`. See #2529.

*Testing done:*
Integration test.

## Merge Checklist

#### General

#### Tests


shreyapandit (dismissed):
Pending refactoring to see how we can automate ingestion of the ECR URL (which needs to be refactored anyways across the codebase), these changes look good to me

shreyapandit (approved):

tests/conftest.py
```diff
@@ -97,6 +97,16 @@ def boto_session(request):
         return boto3.Session(region_name=DEFAULT_REGION)
 
 
+@pytest.fixture(scope="session")
+def account(boto_session):
+    return boto_session.client("sts").get_caller_identity()["Account"]
+
+
```
ahsan-z-khan, line 104: Do we need to create these two fixtures?

Can't we just use `boto_session` to access `account` and `region` ?

Any benefit for creating this one?

bveeramani: For `account`, it's to reduce code duplication. While it's true we could use `boto_session.client("sts").get_caller_identity()["Account"]` every time we want to access the account ID, IMO setting up a fixture for the account is a cleaner implementation.

With respect to the `region` fixture, there's less of a case for that. I subjectively feel it simplifies the implementation, but I also see the argument for why it may be unnecessary. If you feel that we should get rid of it, I'd be okay with doing so.

ahsan-z-khan (approved):

PR #2548, opened 2021-08-02
feature: add LambdaStep support for SageMaker Pipelines

rohangujarathi:
*Issue #, if available:*
AML-116688

*Description of changes:*
Added a new Lambda_helper class and boto3 wrappers to create, update, delete and invoke lambdas.
Added support for LambdaStep
Added relevant unit tests and integ tests
*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/tests/integ/test_workflow.py
+++ b/tests/integ/test_workflow.py
@@ -845,7 +845,8 @@ def test_two_step_lambda_pipeline_with_output_reference(
     step_lambda2 = LambdaStep(
         name="lambda-step2",
         lambda_func=Lambda(
-            function_arn="arn:aws:lambda:us-west-2:123456789012:function:sagemaker_test_lambda"
+            function_arn="arn:aws:lambda:us-west-2:123456789012:function:sagemaker_test_lambda",
+            session=sagemaker_session
         ),
         inputs={"arg1": outputParam1},
         outputs=[],
--- a/tests/integ/test_workflow.py
+++ b/tests/integ/test_workflow.py
@@ -52,6 +52,7 @@ from sagemaker.spark.processing import PySparkProcessor, SparkJarProcessor
 from sagemaker.workflow.conditions import ConditionGreaterThanOrEqualTo, ConditionIn
 from sagemaker.workflow.condition_step import ConditionStep
 from sagemaker.workflow.callback_step import CallbackStep, CallbackOutput, CallbackOutputTypeEnum
+from sagemaker.workflow.lambda_step import LambdaStep, LambdaOutput, LambdaOutputTypeEnum
 from sagemaker.wrangler.processing import DataWranglerProcessor
 from sagemaker.dataset_definition.inputs import DatasetDefinition, AthenaDatasetDefinition
 from sagemaker.workflow.execution_variables import ExecutionVariables
@@ -70,6 +71,7 @@ from sagemaker.workflow.steps import (
 )
 from sagemaker.workflow.step_collections import RegisterModel
 from sagemaker.workflow.pipeline import Pipeline
+from sagemaker.lambda_helper import Lambda
 from sagemaker.feature_store.feature_group import FeatureGroup, FeatureDefinition, FeatureTypeEnum
 from tests.integ import DATA_DIR
 from tests.integ.kms_utils import get_or_create_kms_key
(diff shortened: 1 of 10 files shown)


tests/integ/test_workflow.py
```diff
@@ -781,6 +783,96 @@ def test_two_step_callback_pipeline_with_output_reference(
...
+    )
+
+    pipeline = Pipeline(
+        name=pipeline_name,
+        parameters=[instance_count],
+        steps=[step_lambda1, step_lambda2],
+        sagemaker_session=sagemaker_session,
+    )
+
+    try:
```
shreyapandit, line 862: Thanks for adding the test. 

This tests the success scenario, I think there is worth in adding a test that catches errors on when a pipeline may not be created successfully, to ensure we are failing gracefully? What is the expected error in that case? Also, will pipeline creation include spinning up the lambda, if so, can you add a test to check the expected behavior upon executing it?

tests/unit/sagemaker/workflow/test_lambda_step.py
```diff
@@ -0,0 +1,171 @@
...
+
+def test_pipeline_interpolates_lambda_outputs(sagemaker_session):
+    parameter = ParameterString("MyStr")
+    outputParam1 = LambdaOutput(output_name="output1", output_type=LambdaOutputTypeEnum.String)
+    outputParam2 = LambdaOutput(output_name="output2", output_type=LambdaOutputTypeEnum.String)
+    lambda_step1 = LambdaStep(
+        name="MyLambdaStep1",
+        depends_on=["TestStep"],
+        lambda_func=Lambda(
+            function_arn="arn:aws:lambda:us-west-2:123456789012:function:sagemaker_test_lambda",
```
shreyapandit, line 99: nit: For us as a whole team, we should in general rethink about refactoring and reusing duplicated arns/resource objects across tests

rohangujarathi: Agree. This is just a dummy ARN though. It will create a pipeline which has that ARN in the functionArn field of Lambda step in the pipeline definition. We just need to verify if the pipeline gets created with the correct function arn. Staring a pipeline execution invokes the lambda function which will require actual resource, but that functionality is not tested here since no change in that behavior and the same has been tested in existing workflow tests

shreyapandit: That'a fair. 

There is some behavior here though that is specific to LambdaStep, for example the Outputs can only be of certain types. What happens if the Output generated is of another type? Do we throw an error?

Could you point me to existing tests that check this scenario

tests/integ/test_workflow.py
rohangujarathi: If the pipeline is not created, it throws a generalized pipeline creation failure error from the service backend which we bubble up in SDK. This is similar to all the other existing steps and there is no change in that.

This test is not creating a actual lambda function but Pipeline creation has an option to create a lambda internally, but we are not testing that since that needs extra lambda related permissions to the execution role which the current role running the tests does not have. But that functionality has been tested locally, so we should be fine there

shreyapandit (changes requested):
As discussed in comments

tests/unit/sagemaker/workflow/test_lambda_step.py
rohangujarathi: Sure. Here are some of those. There are more
https://github.com/aws/sagemaker-python-sdk/blob/master/tests/integ/test_workflow.py#L476
https://github.com/aws/sagemaker-python-sdk/blob/master/tests/integ/test_workflow.py#L578
https://github.com/aws/sagemaker-python-sdk/blob/master/tests/integ/test_workflow.py#L675

rohangujarathi: If the output generated is of different type, and is referenced in another step, tioga will throw an error while validation it. This will be an error while running the pipeline and pipeline execution will be in failed status. We don't have a test for this specific scenario in sdk since this is a pipeline execution error and we are not throwing any error in SDK

shreyapandit: Makes sense, thanks for clarifying!

shreyapandit (approved):

PR #2552, opened 2021-08-03
fix: deprecate tag logic on Association

danabens:
associations don't have arns so this tagging logic doesn't make sense

## Merge Checklist

#### General

#### Tests


src/sagemaker/lineage/association.py
```diff
@@ -64,28 +64,6 @@ def delete(self):
         """Delete this Association from SageMaker."""
         self._invoke_api(self._boto_delete_method, self._boto_delete_members)
 
-    def set_tag(self, tag=None):
-        """Add a tag to the object.
```
shreyapandit, line 68: > associations don't have arns so this tagging logic doesn't make sense

Looking at the [description](https://github.com/danabens/sagemaker-python-sdk/blob/remove-set-tag/src/sagemaker/lineage/association.py#L33) for this class, an association object still has a source and destination arn as attributes. Has this behavior changed?

Also, can we deprecate this instead and communicate this change instead of deleting it first, and then delete it eventually ? I am concerned with downstream consumers of the lineage APIs, who might be currently setting tags as part of their workflows and might find those workflows failing when they pull in the release with this change. Thanks so much!

danabens: > Has this behavior changed?

No, association does have  a source and dest but does not have an arn itself so it never made sense to have tagging behavior on this class in the first place. Clients tag the source and dest in other classes or by calling add_tags directly.

> can we deprecate this instead 

Ya, good idea.

jerrypeng7773 (approved):

ahsan-z-khan (approved):

src/sagemaker/lineage/association.py
shreyapandit: Thanks for the clarification! Waiting for tests to pass before merge

shreyapandit (approved):

PR #2564, opened 2021-08-07
Revert #2560

verdimrc:
*Issue #, if available:*

*Description of changes:*
- Framework processor uses a python3 bootstrap script
- Re-enable the new SKLearn processor

*Testing done:* unit, integ.

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)
- [ ] I have passed the region in to all S3 and STS clients that I've initialized as part of this change.
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have checked that my tests are not configured for a specific region or account (if appropriate)
- [ ] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2251, opened 2021-03-29
feature: processors that support multiple Python files, requirements.txt, and dependencies.

verdimrc:
*Issue #, if available:* #1248, #2117

*Description of changes:* Propose processing classes that are feature-parity with estimator. These classes allow SDK users to  runn a Python job that consists of multiple Python scripts, `requirements.txt` and additional dependencies.

Documentation provided as docstrings.

*Testing done:* on my own AWS account, ran processing jobs using the proposed classes (`FrameworkProcessor` and its subclasses) -- the testing scripts are located [here](https://github.com/aws-samples/smallmatter-package/commit/79706e93e2599fb236902784afb506d630f9bc3c), and usage is as outlined as in #1248  ([this comment](https://github.com/aws/sagemaker-python-sdk/issues/1248#issuecomment-779544372)). 

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)
- [ ] I have passed the region in to all S3 and STS clients that I've initialized as part of this change.
- [x] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have checked that my tests are not configured for a specific region or account (if appropriate)
- [ ] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2633, opened 2021-09-14
Restore SKLearn FrameworkProcessor via _normalize_args

athewsey:
**Issue #, if available:** #2656 

**Description of changes:**

An alternative proposal to #2564 for restoring `FrameworkProcessor` to `SKLearnProcessor` and fixing how `FrameworkProcessor` works with SageMaker Pipelines (which currently seems broken).

- Standardize on existing `Processor._normalize_args()` (as used by `ProcessingStep`) as the interface point to handle user and extra code inputs
  - Integrate `FrameworkProcessor`'s handling of user code and framework code upload into this function, instead of hooking in at `run()` or `get_run_args()`
  - Support plain (non-tarballed and no requirements.txt) `source_dir` functionality even in `ScriptProcessor`
- Migrate `FrameworkProcessor` entrypoint from `/bin/bash` to POSIX-compliant `/bin/sh`, because some lightweight Linux distributions optimized for containers (e.g. Alpine) may not install bash by default: This should be a more flexible option for bringing custom containers with FrameworkProcessor
- Make some minor improvements to the documentation for FrameworkProcessor: Explicitly mentioning `source_dir` on the Processing user guide page, and ensuring *Processor classes get listed in framework API directories.

**Additional context & design:**

- #2251 added support for Processing Jobs to consume folders of code (instead of single files) and install user code dependencies at runtime (via requirements.txt in the folder) similarly to what users are used to with `Framework`-based training jobs.
  - User's submitted code/folder is bundled to a tarball on S3 as a `code` input to the processing job
  - A small shell script is uploaded to S3 as an `entrypoint` input and overrides the entrypoint of the job: The shell script extracts the bundle and installs requirements.txt, then kicks off the user's actual requested entrypoint/script.
- ...But it seemed to introduce bugs in existing SM Pipelines where processing jobs would try to execute Python scripts as shell scripts (e.g. `import: command not found`) - so specifically `SKLearnProcessor` was rolled back to its old implementation by #2560.
- Seeking to fix this symptom, #2564 has been in review for a while to restore `SKLearnProcessor`'s framework functionality by changing `FrameworkProcessor` to use a python entrypoint instead of shell script.

From my investigations, it seems like #2564 proposes to patch a symptom of the issue rather than the root cause: I've not seen any reports that the entrypoint script being shell-based is an issue per-se (e.g. no complaints about custom containers not having `/bin/bash` installed) - the problem is the inconsistency of trying to use the command for the (shell-script) framework entrypoint against the file for the user's (Python) script.

In fact, switching to a Python-based framework entrypoint script would:
- Risk masking the issue by making the user's code run OK but the requirements.txt be ignored
- Introduce a hard dependency on Python in the target container that could make life harder for any users wanting to derive new FrameworkProcessors for other languages e.g. `R`, `Julia`, etc.

In this alternative solution, we **integrate `FrameworkProcessor` more closely** with `ScriptProcessor` by overriding parent private methods (including `_normalize_args`, `_include_code_in_inputs`, etc) rather than adding extra new methods hooked in to specific events as before.

This allows `FrameworkProcessor` to work correctly with SM Pipelines without changing the API between them, with the trade-off that several closely coupled overrides are needed with some code duplication: Every new parameter a child class adds to `run()` (e.g. `git_config`, `dependencies`, etc) needs to be pushed through a few internal methods where the only difference is the number of args transmitted.

**Alternatives** I think could include:

- Creating a new standard API on Processor similar to `_normalize_args` for external classes needing to generate the actual job arguments; having existing Processor/ScriptProcessor pass this through to `_normalize_args`, and having FrameworkProcessor add in the additional required hook for preparing code bundle.
- Refactoring `Processor.get_run_args()` to be this API - but I think this would be a breaking change
- A more significant refactor of the hierarchy of internal methods in *Processor, which could accidental risk breaking changes or finding that not much could be improved without breaking changes

**Testing done:**

Some minor additions to test coverage (including checking ScriptProcessor with a folder works), but mostly captured under the scope of existing tests - just tweaked arguments for some diversity between providing a script or a full folder + requirements.txt to `SKLearnProcessor`.

Verified the fix against the [pipelines sample notebook](https://github.com/aws/amazon-sagemaker-examples/blob/master/sagemaker-pipelines/tabular/abalone_build_train_deploy/sagemaker-pipelines-preprocess-train-evaluate-batch-transform.ipynb) that was known to be broken before.

## Merge Checklist

#### General

- [X] ~I have passed the region in to all S3 and STS clients that I've initialized as part of this change.~
  - Improved presentation of framework processing classes in the API doc and SageMaker Processing user guide.

#### Tests

  - Tweaked existing test configurations to try and get good coverage of possible axes
  - Added a unit test to specifically validate SKLearnProcessor (and by extension FrameworkProcessor) behaves as expected by pipelines `ProcessingStep`
- [X] ~I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)~


verdimrc:
A note on alternative (1): with this PR, practically `get_run_args()` can be simply an alias to `_normalize_args()`.

athewsey:
(Pushed to draft in favour of #2664)

ahsan-z-khan (commented):
@athewsey 
Is this related to #2664 ? Are these dependent on each other? OR can we close this?

trajanikant:
@athewsey 
This has been open since 2021. Kindly close this or share an update related to this.

mufaddal-rohawala:
Closing as this has been in draft state for too long! Please reopen if the change is still relevant.

PR #2645, opened 2021-09-17
fix: fix _repack_model script used in pipelines not supporting depend…

staubhp:
…encies or source_dir arguments, which are used for script-mode in models

*Issue #, if available:*
N/A

*Description of changes:*
When creating a model, users can optionally specify a custom entry point as well as a source directory and a list of dependencies. These custom scripts are "repacked" into the model tarball by the SDK.

Because Pipelines execute outside of the context of the SDK, they add a "RepackModel" step when users are registering a model that uses script-mode. This repack step starts a training job that runs the `_repack_model.py` script which puts the custom scripts into the tarball.

Previously, this script only handled custom entry_points, allowing users to customize inference code but not allowing them to add custom dependencies. This change adds support for repacking the `dependencies` and `source_dir` args the same way that the SDK does.

*Testing done:*
Unit, manual.

Note that the code being tested is a training job script that works with absolute root filepaths. For that reason, the unit tests are not enabled because they will fail in most cases due to permissions to the `/opt/` directory.

## Merge Checklist

#### General

#### Tests


src/sagemaker/workflow/_repack_model.py
```diff
@@ -34,15 +34,19 @@
...
-    parser.add_argument("--inference_script", type=str, default="inference.py")
-    parser.add_argument("--model_archive", type=str, default="model.tar.gz")
-    args = parser.parse_args()
+def repack(inference_script, model_archive, dependencies=None, source_dir=None):
+    """Repack custom dependencies and code into an existing model TAR archive
+
+    Args:
+        inference_script (str): The path to the custom entry point.
+        model_archive (str): The name of the model TAR archive.
+        dependencies (str): A space-delimited string of paths to custom dependencies.
```
danabens, line 43: nit this description makes more sense on a cli argument than a method parameter

src/sagemaker/workflow/_repack_model.py
```diff
@@ -51,17 +55,59 @@
...
+        if dependencies:
+            for dependency in dependencies.split(" "):
+                actual_dependency_path = os.path.join("/opt/ml/code", dependency)
+                lib_dir = os.path.join(code_dir, "lib")
+                if not os.path.exists(lib_dir):
+                    os.mkdir(lib_dir)
+                if os.path.isdir(actual_dependency_path):
+                    shutil.copytree(
+                        actual_dependency_path,
+                        os.path.join(lib_dir, os.path.basename(actual_dependency_path)),
```
danabens, line 91: why is `os.path.join` needed here but not on line 77?

src/sagemaker/workflow/_repack_model.py
```diff
@@ -51,17 +55,59 @@
...
         # custom inference script, to the output of this training job
         copy_tree(src_dir, "/opt/ml/model")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--inference_script", type=str, default="inference.py")
+    parser.add_argument("--dependencies", type=str, default=None)
+    parser.add_argument("--source_dir", type=str, default=None)
+    parser.add_argument("--model_archive", type=str, default="model.tar.gz")
```
danabens, line 106: this also has a `help` param if that would benefit users

tests/unit/sagemaker/workflow/test_repack_model_script.py
```diff
@@ -0,0 +1,174 @@
...
+    # create files that will be added to model.tar.gz
+    create_file_tree(
+        "/opt/ml/code",
+        [
+            "inference.py",
+        ],
+    )
+
+    # repack
+    _repack_model.repack(inference_script="inference.py", model_archive=model_tar_name)
```
danabens, line 57: nice, would be nice to also cover `__main__`

src/sagemaker/workflow/_repack_model.py
staubhp: `os.path.join` is being used for both calls, but on line 77 it's being done in the assignment of the `s` and `d` variables

src/sagemaker/workflow/_repack_model.py
staubhp: Ack. I don't think it would be much use since this script is intended for, and will only work in, a training job

tests/unit/sagemaker/workflow/test_repack_model_script.py
staubhp: One of the reasons I moved the logic to a function was because I couldn't figure out how to invoke the `__main__` of a module from within python :) As far as I understand,`__name__` is only set to `__main__` if it's called from a command line. Suggestions welcome

danabens (dismissed):

Issue #2736, opened 2021-10-27
upserting a pipeline kills caching


neilmcguigan:
**Describe the bug**
Upserting a pipeline always causes cache misses

**To reproduce**
Create a pipeline with caching enabled. Run the pipeline. Upsert the pipeline. Run the pipeline again. You'll get a cache miss every time

```
cache_config = CacheConfig(
    enable_caching=True,
    expire_after = "P30M"
)

# define the pre-processor:
processor = PySparkProcessor(
    role=role,
    instance_type="ml.m5.large",
    instance_count=1,
    framework_version="3.0"
)

step1 = ProcessingStep(
    name="step1",
    processor=processor,
    code="processor1.py",
    cache_config=cache_config
)

# define the pipeline
pipeline = Pipeline(
    name="pipeline1",
    steps=[step1]
)

pipeline.upsert(role_arn="...")

pipeline.start()

```
**Expected behavior**
I expect a cache hit, even if I upsert a pipeline

**System information**
A description of your system. Please provide:
- **SageMaker Python SDK version**: 2.66.2 (latest)
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: any
- **Framework version**: any
- **Python version**: 3.8.10
- **CPU or GPU**: any
- **Custom Docker image (Y/N)**: no

**Additional context**

purak24:
This still results in a cache miss. I believe the reason is that even the code script ("abalone/preprocessing.py") will be a `ProcessingInput` behind the scenes. The file key on S3 where the code script is uploaded is also getting appended with timestamp resulting in different inputs and hence cache miss. Note this happens even if we provide the `name` in `ProcessingStep` as that is treated as a `base_job_name` instead of `_current_job_name`. Is there any way we can set the `_current_job_name` from the `ProcessingStep` to circumvent this?

staubhp:
You're right, it turns out [we are not passing the `job_name`](https://github.com/aws/sagemaker-python-sdk/blob/dev/src/sagemaker/workflow/steps.py#L516-L521) from the processor when normalizing the args in the ProcessingStep, which will result in a generated job name each time. 

This should be a simple fix which we'll prioritize. The only workaround I can offer now is providing an S3 URI for your `code` arg instead of a local file to prevent the SDK from uploading it with the generated job name.

giovannirescia-upwork:
@staubhp what about for the training job? I am trying the Abalon example and still getting cache misses.

[...]

PR #2786, opened 2021-12-06
fix: Normalizing job_name in the ProcessingStep.

purak24:
*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/2736

*Description of changes:* Passing the `job_name` from the [processor](https://github.com/aws/sagemaker-python-sdk/blob/dev/src/sagemaker/workflow/steps.py#L516-L521) when normalizing the args in the `ProcessingStep`, to avoid resulting in a generated job name each time.

*Testing done:* Ran test suite

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/workflow/steps.py
+++ b/src/sagemaker/workflow/steps.py
@@ -496,6 +496,7 @@ class ProcessingStep(ConfigurableRetryStep):
         self.job_arguments = job_arguments
         self.code = code
         self.property_files = property_files
+        self.job_name = name
 
         # Examine why run method in sagemaker.processing.Processor mutates the processor instance
         # by setting the instance's arguments attribute. Refactor Processor.run, if possible.
@@ -514,6 +515,7 @@ class ProcessingStep(ConfigurableRetryStep):
         ProcessingJobName and ExperimentConfig cannot be included in the arguments.
         """
         normalized_inputs, normalized_outputs = self.processor._normalize_args(
+            job_name=self.job_name,
             arguments=self.job_arguments,
             inputs=self.inputs,
             outputs=self.outputs,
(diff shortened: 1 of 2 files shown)


src/sagemaker/workflow/steps.py
```diff
@@ -496,6 +496,7 @@ def __init__(
         self.job_arguments = job_arguments
         self.code = code
         self.property_files = property_files
+        self.job_name = name
```
staubhp, line 499: If we use the step name as the job name then all input and output locations will be overwritten when the args are normalized. In practical terms that means any pipeline with a step named `MyProcessingStep` will start using the code for this processor.

Things get a little better if we include the `pipeline_name` in the `job_name`, but that still means the code will be overwritten which will prevent recreatibility (ie, looking at a previous execution and going to S3 to see the script that was run).

Let me give this some thought. We need some signal of intentionality from the user here to maintain backwards compatibility.

staubhp: I think the best approach here is to hash the caller's script contents and include that in the `job_name`. This ensures there will be cache hits/misses as appropriate while not requiring any input from the user. I can work on this change since it's a little heavier weight than initially expected

purak24: Good callout! How do you feel about using the enable caching boolean as the signal of intentionality from the user? Something like:

```
name = pipeline_name + job_name
self.job_name = name if cache_config.enable_caching else None
```

purak24: Your comment loaded later for me - I like your idea as well. This will go against the documentation [here](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-caching.html)  

> Pipelines doesn't check whether the data or code that the arguments point to has changed

As a user though, I do believe it will be nice to evolve caching pipeline steps functionality to detect data/code changes!

staubhp: https://github.com/aws/sagemaker-python-sdk/pull/2790

I'm going to close this PR

PR #2855, opened 2022-01-19
feature: jumpstart vulnerability and deprecated check

evakravi:
*Issue #, if available:*

*Description of changes:*
This PR adds adds support for the deprecated and vulnerability fields of the JumpStart model specs. If specs for a vulnerable or deprecated model are requested, a new exception type will be surfaced. There is a user option to override these safety checks, so they may use a vulnerable/deprecated model if they wish.  

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/image_uris.py
+++ b/src/sagemaker/image_uris.py
@@ -45,6 +45,8 @@ def retrieve(
     training_compiler_config=None,
     model_id=None,
     model_version=None,
+    tolerate_vulnerable_model=None,
+    tolerate_deprecated_model=None,
 ) -> str:
     """Retrieves the ECR URI for the Docker image matching the given arguments.
 
@@ -79,6 +81,10 @@ def retrieve(
             (default: None).
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
+        tolerate_vulnerable_model (bool): True if vulnerable models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
 
     Returns:
         str: the ECR URI for the corresponding SageMaker Docker image.
@@ -106,6 +112,8 @@ def retrieve(
             distribution,
             base_framework_version,
             training_compiler_config,
+            tolerate_vulnerable_model,
+            tolerate_deprecated_model,
         )
 
     if training_compiler_config is None:
--- a/src/sagemaker/image_uris.py
+++ b/src/sagemaker/image_uris.py
@@ -93,7 +93,11 @@ def retrieve(
         str: the ECR URI for the corresponding SageMaker Docker image.
 
     Raises:
+        NotImplementedError: If the scope is not supported.
         ValueError: If the combination of arguments specified is not supported.
+        VulnerableJumpStartModelError: If any of the dependencies required by the script have
+            known security vulnerabilities.
+        DeprecatedJumpStartModelError: If the version of the model is deprecated.
     """
     if is_jumpstart_model_input(model_id, model_version):
 
--- a/src/sagemaker/image_uris.py
+++ b/src/sagemaker/image_uris.py
@@ -81,10 +81,13 @@ def retrieve(
             (default: None).
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
-        tolerate_vulnerable_model (bool): True if vulnerable models should be tolerated (exception
-            not raised). False if these models should raise an exception. (Default: None).
-        tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
-            not raised). False if these models should raise an exception. (Default: None).
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model specifications
+            should be tolerated (exception not raised). False or None, raises an exception if
+            the script used by this version of the model has dependencies with known security
+            vulnerabilities. (Default: None).
+        tolerate_deprecated_model (bool): True if deprecated versions of model specifications
+            should be tolerated (exception not raised). False or None, raises an exception
+            if the version of the model is deprecated. (Default: None).
 
     Returns:
         str: the ECR URI for the corresponding SageMaker Docker image.
--- a/src/sagemaker/image_uris.py
+++ b/src/sagemaker/image_uris.py
@@ -82,9 +82,9 @@ def retrieve(
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
         tolerate_vulnerable_model (bool): True if vulnerable models should be tolerated (exception
-            not thrown). False if these models should throw an exception. (Default: None).
+            not raised). False if these models should raise an exception. (Default: None).
         tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
-            not thrown). False if these models should throw an exception. (Default: None).
+            not raised). False if these models should raise an exception. (Default: None).
 
     Returns:
         str: the ECR URI for the corresponding SageMaker Docker image.
--- a/src/sagemaker/image_uris.py
+++ b/src/sagemaker/image_uris.py
@@ -45,8 +45,8 @@ def retrieve(
     training_compiler_config=None,
     model_id=None,
     model_version=None,
-    tolerate_vulnerable_model=None,
-    tolerate_deprecated_model=None,
+    tolerate_vulnerable_model=False,
+    tolerate_deprecated_model=False,
 ) -> str:
     """Retrieves the ECR URI for the Docker image matching the given arguments.
 
@@ -82,12 +82,12 @@ def retrieve(
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
         tolerate_vulnerable_model (bool): True if vulnerable versions of model specifications
-            should be tolerated (exception not raised). False or None, raises an exception if
+            should be tolerated (exception not raised). If False, raises an exception if
             the script used by this version of the model has dependencies with known security
-            vulnerabilities. (Default: None).
+            vulnerabilities. (Default: False).
         tolerate_deprecated_model (bool): True if deprecated versions of model specifications
-            should be tolerated (exception not raised). False or None, raises an exception
-            if the version of the model is deprecated. (Default: None).
+            should be tolerated (exception not raised). If False, raises an exception
+            if the version of the model is deprecated. (Default: False).
 
     Returns:
         str: the ECR URI for the corresponding SageMaker Docker image.
@@ -100,10 +100,6 @@ def retrieve(
         DeprecatedJumpStartModelError: If the version of the model is deprecated.
     """
     if is_jumpstart_model_input(model_id, model_version):
-
-        # adding assert statements to satisfy mypy type checker
-        assert model_id is not None
-        assert model_version is not None
 
         return artifacts._retrieve_image_uri(
             model_id,
--- a/src/sagemaker/jumpstart/artifacts.py
+++ b/src/sagemaker/jumpstart/artifacts.py
@@ -41,8 +41,8 @@ def _retrieve_image_uri(
     distribution: Optional[str],
     base_framework_version: Optional[str],
     training_compiler_config: Optional[str],
-    tolerate_vulnerable_model: Optional[bool],
-    tolerate_deprecated_model: Optional[bool],
+    tolerate_vulnerable_model: bool,
+    tolerate_deprecated_model: bool,
 ):
     """Retrieves the container image URI for JumpStart models.
 
@@ -75,12 +75,12 @@ def _retrieve_image_uri(
         distribution (dict): A dictionary with information on how to run distributed training
         training_compiler_config (:class:`~sagemaker.training_compiler.TrainingCompilerConfig`):
             A configuration class for the SageMaker Training Compiler.
-        tolerate_vulnerable_model (Optional[bool]): True if vulnerable versions of model
-            specifications should be tolerated (exception not raised). False or None, raises an
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model
+            specifications should be tolerated (exception not raised). If False, raises an
             exception if the script used by this version of the model has dependencies with known
             security vulnerabilities.
-        tolerate_deprecated_model (Optional[bool]): True if deprecated versions of model
-            specifications should be tolerated (exception not raised). False or None, raises
+        tolerate_deprecated_model (bool): True if deprecated versions of model
+            specifications should be tolerated (exception not raised). If False, raises
             an exception if the version of the model is deprecated.
 
     Returns:
@@ -95,8 +95,6 @@ def _retrieve_image_uri(
     if region is None:
         region = JUMPSTART_DEFAULT_REGION_NAME
 
-    assert region is not None
-
     model_specs = verify_model_region_and_return_specs(
         model_id=model_id,
         version=model_version,
@@ -109,7 +107,6 @@ def _retrieve_image_uri(
     if image_scope == JumpStartScriptScope.INFERENCE:
         ecr_specs = model_specs.hosting_ecr_specs
     elif image_scope == JumpStartScriptScope.TRAINING:
-        assert model_specs.training_ecr_specs is not None
         ecr_specs = model_specs.training_ecr_specs
 
     if framework is not None and framework != ecr_specs.framework:
@@ -172,8 +169,8 @@ def _retrieve_model_uri(
     model_version: str,
     model_scope: Optional[str],
     region: Optional[str],
-    tolerate_vulnerable_model: Optional[bool],
-    tolerate_deprecated_model: Optional[bool],
+    tolerate_vulnerable_model: bool,
+    tolerate_deprecated_model: bool,
 ):
     """Retrieves the model artifact S3 URI for the model matching the given arguments.
 
@@ -185,12 +182,12 @@ def _retrieve_model_uri(
         model_scope (str): The model type, i.e. what it is used for.
             Valid values: "training" and "inference".
         region (str): Region for which to retrieve model S3 URI.
-        tolerate_vulnerable_model (Optional[bool]): True if vulnerable versions of model
-            specifications should be tolerated (exception not raised). False or None, raises an
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model
+            specifications should be tolerated (exception not raised). If False, raises an
             exception if the script used by this version of the model has dependencies with known
             security vulnerabilities.
-        tolerate_deprecated_model (Optional[bool]): True if deprecated versions of model
-            specifications should be tolerated (exception not raised). False or None, raises
+        tolerate_deprecated_model (bool): True if deprecated versions of model
+            specifications should be tolerated (exception not raised). If False, raises
             an exception if the version of the model is deprecated.
     Returns:
         str: the model artifact S3 URI for the corresponding model.
@@ -204,8 +201,6 @@ def _retrieve_model_uri(
     if region is None:
         region = JUMPSTART_DEFAULT_REGION_NAME
 
-    assert region is not None
-
     model_specs = verify_model_region_and_return_specs(
         model_id=model_id,
         version=model_version,
@@ -218,7 +213,6 @@ def _retrieve_model_uri(
     if model_scope == JumpStartScriptScope.INFERENCE:
         model_artifact_key = model_specs.hosting_artifact_key
     elif model_scope == JumpStartScriptScope.TRAINING:
-        assert model_specs.training_artifact_key is not None
         model_artifact_key = model_specs.training_artifact_key
 
     bucket = get_jumpstart_content_bucket(region)
@@ -233,8 +227,8 @@ def _retrieve_script_uri(
     model_version: str,
     script_scope: Optional[str],
     region: Optional[str],
-    tolerate_vulnerable_model: Optional[bool],
-    tolerate_deprecated_model: Optional[bool],
+    tolerate_vulnerable_model: bool,
+    tolerate_deprecated_model: bool,
 ):
     """Retrieves the script S3 URI associated with the model matching the given arguments.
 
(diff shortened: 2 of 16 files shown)


src/sagemaker/image_uris.py
```diff
@@ -45,6 +45,8 @@ def retrieve(
     training_compiler_config=None,
     model_id=None,
     model_version=None,
+    tolerate_vulnerable_model=None,
+    tolerate_deprecated_model=None,
```
JGuinegagne, line 49: could you default these two `bool` to `False`?

src/sagemaker/image_uris.py
```diff
@@ -79,6 +81,10 @@ def retrieve(
             (default: None).
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
+        tolerate_vulnerable_model (bool): True if vulnerable models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
```
JGuinegagne, line 87: nit: `[...] (exception not raised). [...]`

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""This module stores exceptions related to SageMaker JumpStart."""
+
+from typing import List, Optional
+
```
JGuinegagne, line 16: question: do you need to import a logger here so that this works in some execution environment which don't have a default stdout?

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+    """Exception raised for errors with vulnerable JumpStart models."""
+
+    def __init__(
+        self,
+        model_id: Optional[str] = None,
+        version: Optional[str] = None,
+        vulnerabilities: Optional[List[str]] = None,
+        inference: Optional[bool] = None,
+        message: Optional[str] = None,
+    ):
```
JGuinegagne, line 28: docstring please.

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+
+class VulnerableJumpStartModelError(Exception):
+    """Exception raised for errors with vulnerable JumpStart models."""
+
+    def __init__(
+        self,
+        model_id: Optional[str] = None,
+        version: Optional[str] = None,
+        vulnerabilities: Optional[List[str]] = None,
+        inference: Optional[bool] = None,
```
JGuinegagne, line 26: for extensibility, would it be possible to rename this argument to `scope` and pass an `enum` here instead of a `bool` ?

That way, we could support packaging for edge, or batch in the future for example.

You can raise a `NotImplementedError` if the scope isn't mapped yet in the SDK.

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""This module stores exceptions related to SageMaker JumpStart."""
+
+from typing import List, Optional
+
+
+class VulnerableJumpStartModelError(Exception):
+    """Exception raised for errors with vulnerable JumpStart models."""
```
JGuinegagne, line 19:
```py
"""Exception raised when trying to access a JumpStart model specs flagged as vulnerable.

Raise this exception only if the scope of attributes accessed in the specifications have vulnerabilities. For example, a model training script may have vulnerabilities, but not the hosting scripts. In such a case, raise a
``VulnerableJumpStartModelError`` only when accessing the training specifications.
"""
```

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+                    f"JumpStart model '{model_id}' and version '{version}' has at least 1 "
+                    "vulnerable dependency in the training scripts. "
+                    f"List of vulnerabilities: {', '.join(vulnerabilities)}"
+                )
+
+        super().__init__(self.message)
+
+
+class DeprecatedJumpStartModelError(Exception):
+    """Exception raised for errors with deprecated JumpStart models."""
```
JGuinegagne, line 54:
```py
"""Exception raised when trying to access a JumpStart model deprecated specifications.

A deprecated specification for a JumpStart model does not mean the whole model is deprecated.
There may be more recent specifications available for this model. For example, all specification before
version ``2.0.0`` may be deprecated, in such a case, the SDK would raise this exception only when
specifications ``1.*`` are accessed.
""""
```

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+        model_id: Optional[str] = None,
+        version: Optional[str] = None,
+        message: Optional[str] = None,
+    ):
+        if message:
+            self.message = message
+        else:
+            if None in [model_id, version]:
+                raise ValueError("Must specify `model_id` and `version` arguments.")
+            self.message = f"JumpStart model '{model_id}' and version '{version}' is deprecated."
```
JGuinegagne, line 67: Nit:
```py
f"Version '{version}' of JumpStart model '{model_id}' is deprecated."
```

We could also add: `[...] Please try targeting a higher version of the model.`, but it might be confusing if we deprecated the whole model. WDYT ?

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+            self.message = message
+        else:
+            if None in [model_id, version, vulnerabilities, inference]:
+                raise ValueError(
+                    "Must specify `model_id`, `version`, `vulnerabilities`, "
+                    "and inference arguments."
+                )
+            if inference is True:
+                self.message = (
+                    f"JumpStart model '{model_id}' and version '{version}' has at least 1 "
```
JGuinegagne, line 39: nit: 
```py
f"Version '{version}' of JumpStart model '{model_id}' has at least 1 "
```

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+                    "vulnerable dependency in the inference scripts. "
+                    f"List of vulnerabilities: {', '.join(vulnerabilities)}"
+                )
+            else:
+                self.message = (
+                    f"JumpStart model '{model_id}' and version '{version}' has at least 1 "
```
JGuinegagne, line 45: same comment.

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+        else:
+            if None in [model_id, version, vulnerabilities, inference]:
+                raise ValueError(
+                    "Must specify `model_id`, `version`, `vulnerabilities`, "
+                    "and inference arguments."
+                )
+            if inference is True:
+                self.message = (
+                    f"JumpStart model '{model_id}' and version '{version}' has at least 1 "
+                    "vulnerable dependency in the inference scripts. "
```
JGuinegagne, line 40: `[...] inference script.` (singular)

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +140,86 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+
+
+def verify_model_region_and_return_specs(
+    model_id: Optional[str],
+    version: Optional[str],
+    scope: Optional[str],
+    region: str,
+    tolerate_vulnerable_model: Optional[bool] = None,
+    tolerate_deprecated_model: Optional[bool] = None,
+):
```
JGuinegagne, line 152: return typing please.

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +140,86 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+            obtains specs.
+        version (Optional[str]): version of the JumpStart model to verify and
+            obtains specs.
+        scope (Optional[str]): scope of the JumpStart model to verify.
+        region (Optional[str]): region of the JumpStart model to verify and
+            obtains specs.
+        tolerate_vulnerable_model (Optional[bool]): True if vulnerable models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
+        tolerate_deprecated_model (Optional[bool]): True if deprecated models should be tolerated (exception
+            not thrown). False if these models should throw an exception. (Default: None).
```
JGuinegagne, line 169: add a `Raises:` section please.

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +140,86 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+    version: Optional[str],
+    scope: Optional[str],
+    region: str,
+    tolerate_vulnerable_model: Optional[bool] = None,
+    tolerate_deprecated_model: Optional[bool] = None,
+):
+    """Verifies that an acceptable model_id, version, scope, and region combination is provided.
+
+    If the scope is not supported, the model id/region/version has no spec, or the model is vulnerable
+    or deprecated, an exception will be raised.
```
JGuinegagne, line 156: remove and address in the `Raises:` section please.

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +140,86 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+        tolerate_deprecated_model = False
+
+    if scope is None:
+        raise ValueError(
+            "Must specify `model_scope` argument to retrieve model "
+            "artifact uri for JumpStart models."
+        )
+
+    if scope not in constants.SUPPORTED_JUMPSTART_SCOPES:
+        raise ValueError(
```
JGuinegagne, line 185: can you raise a `NotImplementedError` please?

src/sagemaker/model_uris.py
```diff
@@ -28,6 +28,8 @@ def retrieve(
     model_id=None,
     model_version: Optional[str] = None,
     model_scope: Optional[str] = None,
+    tolerate_vulnerable_model: Optional[bool] = None,
+    tolerate_deprecated_model: Optional[bool] = None,
```
JGuinegagne, line 32: remove `Optional` and use `False` as default please.

src/sagemaker/script_uris.py
```diff
@@ -27,6 +28,8 @@ def retrieve(
     model_id=None,
     model_version=None,
     script_scope=None,
+    tolerate_vulnerable_model: Optional[bool] = None,
+    tolerate_deprecated_model: Optional[bool] = None,
```
JGuinegagne, line 32: same comment.

tests/unit/sagemaker/jumpstart/test_utils.py
```diff
@@ -112,3 +117,93 @@ def test_get_sagemaker_version(patched_parse_sm_version: Mock):
...
+        utils.verify_model_region_and_return_specs(
+            model_id="pytorch-eqa-bert-base-cased", version="*", scope=TRAINING, region="us-west-2"
+        )
+    assert (
+        "JumpStart model 'pytorch-eqa-bert-base-cased' and version '*' has "
+        "at least 1 vulnerable dependency in the training scripts. List of vulnerabilities: "
+        "some, vulnerability" == str(e.value.message)
+    )
+
+    assert (
```
JGuinegagne, line 171: non-blocking: ideally, make this a different test case.

tests/unit/sagemaker/jumpstart/test_utils.py
```diff
@@ -112,3 +117,93 @@ def test_get_sagemaker_version(patched_parse_sm_version: Mock):
...
+        utils.verify_model_region_and_return_specs(
+            model_id="pytorch-eqa-bert-base-cased",
+            version="*",
+            scope=INFERENCE,
+            region="us-west-2",
+            tolerate_deprecated_model=True,
+        )
+        is not None
```
JGuinegagne, line 208: same comment here.

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -0,0 +1,69 @@
...
+                    f"List of vulnerabilities: {', '.join(vulnerabilities)}"
+                )
```
JGuinegagne, line 42: how about adding: `[...] Please try targeting a higher version of this model.`

JGuinegagne (changes requested):
Two main comments:
- please use default `False` for the `tolerate` boolean instead of using `Optional`
- please add more explanation to the error messages.

src/sagemaker/image_uris.py
evakravi: I'd rather keep this `None`, keeping the pattern of this function.

src/sagemaker/jumpstart/exceptions.py
evakravi: I don't think so, since I don't print anything in this module.

src/sagemaker/jumpstart/exceptions.py
evakravi: I think we could offer a suggestion, it doesn't hurt.

src/sagemaker/model_uris.py
evakravi: I'd rather keep the pattern of setting these params to `None`.

src/sagemaker/script_uris.py
evakravi: I'd rather keep the pattern of setting these params to None.

JGuinegagne:
Question: shall we also add this vulnerability check in the hyperparameters and environments utilities?

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -72,39 +75,36 @@ def _retrieve_image_uri(
...
+            not raised). False if these models should raise an exception.
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
+            not raised). False if these models should raise an exception.
 
     Returns:
         str: the ECR URI for the corresponding SageMaker Docker image.
 
     Raises:
         ValueError: If the combination of arguments specified is not supported.
+        VulnerableJumpStartModelError: If the model is vulnerable.
```
JGuinegagne, line 88: nit: If any of the dependencies required by the script have known security vulnerabilities

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -72,39 +75,36 @@ def _retrieve_image_uri(
...
+        DeprecatedJumpStartModelError: If the model is deprecated.
```
JGuinegagne, line 89: nit: If the version of the model is deprecated.

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -72,39 +75,36 @@ def _retrieve_image_uri(
         distribution (dict): A dictionary with information on how to run distributed training
         training_compiler_config (:class:`~sagemaker.training_compiler.TrainingCompilerConfig`):
             A configuration class for the SageMaker Training Compiler.
+        tolerate_vulnerable_model (bool): True if vulnerable models should be tolerated (exception
+            not raised). False if these models should raise an exception.
```
JGuinegagne, line 79: nit: 
```py
"""True if vulnerable versions of model specifications should be tolerated (exception not raised). 
False or omitted, raises an exception if the script used by this version of the model has dependencies 
with known security vulnerabilities. (Default: None)."""
```

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -72,39 +75,36 @@ def _retrieve_image_uri(
...
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated (exception
+            not raised). False if these models should raise an exception.
```
JGuinegagne, line 81: nit:
```py
"""True if deprecated versions of model specifications should be tolerated (exception not raised). 
False or omitted, raises an exception if the version of the model is deprecated. (Default: None).
"""
```

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -72,39 +75,36 @@ def _retrieve_image_uri(
...
+        model_id=model_id,
+        version=model_version,
+        scope=image_scope,
+        region=region,
+        tolerate_vulnerable_model=tolerate_vulnerable_model,
+        tolerate_deprecated_model=tolerate_deprecated_model,
     )
 
-    if image_scope == INFERENCE:
+    if image_scope == JumpStartScriptScope.INFERENCE.value:
```
JGuinegagne, line 105: nit: no need for `.value` since the enum class inherits from `str` as its first parent.

src/sagemaker/image_uris.py
```diff
@@ -106,6 +112,8 @@ def retrieve(
             distribution,
             base_framework_version,
             training_compiler_config,
+            tolerate_vulnerable_model,
```
JGuinegagne, line 115: Apologies for not thinking about this earlier, but here, should we pass `tolerate_vulnerable_model=True` and remove the argument from the utility function signature ?

The idea is that only the script may have a vulnerable flag, not the image itself. So if a customer wants to invoke this utility, but use their own scripts URI, should we block them? The same goes for the model artifact URI. Let's discuss, this is an important point.

JGuinegagne (changes requested):
- let's make the docstring a bit more descriptive
- consider whether to always tolerate vulnerability for `image_uris` and `model_uris`

src/sagemaker/image_uris.py
evakravi: Good point. After discussing on Slack, we determined that the model and version should be entirely vulnerable for all of inference or training. This means that accessing model artifact URIs or image URIs which corresponds to a vulnerable script URI should also raise an Exception.

JGuinegagne (approved):

src/sagemaker/image_uris.py
```diff
@@ -79,12 +81,23 @@ def retrieve(
             (default: None).
         model_version (str): Version of the JumpStart model for which to retrieve the
             image URI (default: None).
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model specifications
+            should be tolerated (exception not raised). False or None, raises an exception if
+            the script used by this version of the model has dependencies with known security
+            vulnerabilities. (Default: None).
+        tolerate_deprecated_model (bool): True if deprecated versions of model specifications
+            should be tolerated (exception not raised). False or None, raises an exception
+            if the version of the model is deprecated. (Default: None).
```
mufaddal-rohawala, line 90: nit: can we default to False for clarity?

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +144,101 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+
+    if scope not in constants.SUPPORTED_JUMPSTART_SCOPES:
+        raise NotImplementedError(
+            "JumpStart models only support scopes: "
+            f"{', '.join(constants.SUPPORTED_JUMPSTART_SCOPES)}."
+        )
+
+    assert model_id is not None
+    assert version is not None
+
```
mufaddal-rohawala, line 204: Just a caution to emit specific Errors here rather than AssertionErrors, unless this is handled upstream elsewhere and this is just a sanity check.

src/sagemaker/image_uris.py
mufaddal-rohawala: +1 to Janathan's comment, is there a specific reason to keep the pattern if this function intact?

src/sagemaker/jumpstart/exceptions.py
mufaddal-rohawala: +1

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +144,101 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+            f"JumpStart model ID '{model_id}' and version '{version}' " "does not support training."
+        )
+
+    if model_specs.deprecated and not tolerate_deprecated_model:
+        raise DeprecatedJumpStartModelError(model_id=model_id, version=version)
+
+    if (
+        scope == constants.JumpStartScriptScope.INFERENCE.value
+        and model_specs.inference_vulnerable
+        and not tolerate_vulnerable_model
```
mufaddal-rohawala, line 223: if not raise `VulnerableJumpStartModelError` on `tolerate_vulnerable_model = True`, should we atleast log as a warning?

src/sagemaker/jumpstart/utils.py
```diff
@@ -136,3 +144,101 @@ def is_jumpstart_model_input(model_id: Optional[str], version: Optional[str]) ->
...
+            model_id=model_id,
+            version=version,
+            vulnerabilities=model_specs.inference_vulnerabilities,
+            scope=constants.JumpStartScriptScope.INFERENCE,
+        )
+
+    if (
+        scope == constants.JumpStartScriptScope.TRAINING.value
+        and model_specs.training_vulnerable
+        and not tolerate_vulnerable_model
```
mufaddal-rohawala, line 235: Same comment as above

src/sagemaker/jumpstart/utils.py
```diff
@@ -214,31 +209,33 @@ def verify_model_region_and_return_specs(
...
+    if scope == constants.JumpStartScriptScope.INFERENCE.value and model_specs.inference_vulnerable:
+        if not tolerate_vulnerable_model:
+            raise VulnerableJumpStartModelError(
+                model_id=model_id,
+                version=version,
+                vulnerabilities=model_specs.inference_vulnerabilities,
+                scope=constants.JumpStartScriptScope.INFERENCE,
+            )
+        LOGGER.warning(
+            "Using vulnerable JumpStart model '%s' and version '%s' (inference).", model_id, version
```
JGuinegagne, line 226: non-blocking: while I agree that logging a warning is good practice, there is one downside to this approach. The typical use case of our utilities is:

```py
model_id = "pytorch-ic-mobilenet-v2"
version = "1.1.2"

train_image = image_uris.retrieve(model_id=model_id, version=version, tolerate_deprecated_model=True)
train_model = model_uris.retrieve(model_id=model_id, version=version, tolerate_deprecated_model=True)
train_script = script_uris.retrieve(model_id=model_id, version=version, tolerate_deprecated_model=True)

estimator = Estimator(image_uri=train_image, source_dir=train_script, model_uri=train_model)
[...]
```

In that case, we would log 3 times the same warning, correct?

This may be the correct tradeoff, we can discuss. One possibility would be to add another state-holding class like [SageMakerSetting](https://github.com/aws/sagemaker-python-sdk/blob/d9d8c68b4bfba9b1adf1d8356de092331d70039b/src/sagemaker/jumpstart/accessors.py#L21) to handle such warnings. The class could store some metadata of the warnings (eg model+version) used and associate a time stamp, and bypass the warning if one was raised less than 10 minutes ago, for example.

In any case, this case be addressed in a separate PR.

evakravi: I don't think it's worth it right now to engineer a solution to this problem. After all, most people will not be using deprecated/vulnerable models, and no models right now have this flag.

mufaddal-rohawala (approved):

PR #3019, opened 2022-03-22
more logging info for static pipeline test data setup

danabens:
more insight into build failures https://us-west-2.console.aws.amazon.com/codesuite/codebuild/142577830533/projects/sagemaker-python-sdk-pr/build/sagemaker-python-sdk-pr%3A2afc29c4-ef6f-4422-b180-8123bd066b9e/?region=us-west-2#

### DIFF
--- a/tests/integ/sagemaker/lineage/conftest.py
+++ b/tests/integ/sagemaker/lineage/conftest.py
@@ -539,7 +539,7 @@ def _get_static_pipeline_execution_arn(sagemaker_session):
         _deploy_static_endpoint(
             execution_arn=pipeline_execution_arn, sagemaker_session=sagemaker_session
         )
-
+    logging.info(f"Using static pipeline {pipeline_execution_arn}")
     return pipeline_execution_arn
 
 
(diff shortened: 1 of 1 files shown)


tests/integ/sagemaker/lineage/conftest.py
```diff
@@ -651,6 +659,8 @@ def static_endpoint_context(sagemaker_session, static_pipeline_execution_arn):
         )
         endpoint_arn = get_endpoint_arn_from_static_pipeline(sagemaker_session)
 
+    endpoint_arn = get_endpoint_arn_from_static_pipeline(sagemaker_session)
```
navinsoni, line 662: It looks like we are overwriting endpoint_arn here

navinsoni (changes requested):

tests/integ/sagemaker/lineage/conftest.py
Snoyark, line 662: Should the above `get_endpoint_arn_from_static_pipeline(sagemaker_session)` be moved out of the if block? Since the same code is running in either case with this change.

danabens: removed

danabens: it is needed within the block since it should return a value only after the endpoint is deployed within the block

staubhp (approved):

tests/integ/sagemaker/lineage/conftest.py
```diff
@@ -709,27 +721,41 @@ def static_model_artifact(sagemaker_session, static_pipeline_execution_arn):
...
+    logging.info(
+        f"Found {len(dataset_associations)} associated with model artifact {model_artifact_arn}"
     )
     if len(dataset_associations["AssociationSummaries"]) == 0:
         # no directly associated dataset. work backwards from the model
         model_associations = sagemaker_session.sagemaker_client.list_associations(
-            DestinationArn=static_model_artifact.artifact_arn, SourceType="Model"
-        )
+            DestinationArn=model_artifact_arn, SourceType="Model"
+        )["AssociationSummaries"]
```
jkasiraj, line 742: Won't this throw an exception if list_associations is empty or null?

jkasiraj (approved):

tests/integ/sagemaker/lineage/conftest.py
danabens: `list_associations` here is the auto generated boto method for calling `ListAssociations`. as long as the request is successful a response will be returned with a field of AssociationSummaries - which may be empty or not but it will be there resulting in this being successful. so the response being null here is unlikely.

navinsoni (approved):

danabens:
investigating build failures, should be fixed

PR #3076, opened 2022-04-22
feature: Add ModelStep for SageMaker Model Building Pipeline

qidewenwhen:
*Description of changes:* feature: Add ModelStep for SageMaker Model Building Pipeline

*Testing done:* unit tests and integ tests

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [X] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [X] I certify that the changes I am introducing will be backword compatible, and I have discussed concerns about this, if any, with the Python SDK team
- [X] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)
- [X] I have passed the region in to all S3 and STS clients that I've initialized as part of this change.
- [ ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)

#### Tests

- [X] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [X] I have added unit and/or integration tests as appropriate to ensure backward compatibility of the changes
- [X] I have checked that my tests are not configured for a specific region or account (if appropriate)
- [X] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3107, opened 2022-05-13
fix: Add script mode env vars to model packages created from Register…

staubhp:
…ModelStep

*Description of changes:*
When an `estimator` is passed to `RegisterModelStep`, the resulting `CreateModelPackage` args will be missing the environment variables needed for script mode to work. Add those env vars to the default container definition.

*Testing done:*
Unit, manual

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -48,6 +48,8 @@ from sagemaker.session_settings import SessionSettings
 LOGGER = logging.getLogger("sagemaker")
 
 NOTEBOOK_METADATA_FILE = "/opt/ml/metadata/resource-metadata.json"
+DIR_PARAM_NAME = "SAGEMAKER_SUBMIT_DIRECTORY"
+SCRIPT_PARAM_NAME = "SAGEMAKER_PROGRAM"
 
 _STATUS_CODE_TABLE = {
     "COMPLETED": "Completed",
@@ -4209,6 +4211,7 @@ def get_model_package_args(
     drift_check_baselines=None,
     customer_metadata_properties=None,
     validation_specification=None,
+    entry_point=None,
 ):
     """Get arguments for create_model_package method.
 
@@ -4239,6 +4242,7 @@ def get_model_package_args(
         drift_check_baselines (DriftCheckBaselines): DriftCheckBaselines object (default: None).
         customer_metadata_properties (dict[str, str]): A dictionary of key-value paired
             metadata properties (default: None).
+        entry_point (str): The path (absolute or relative) to the custom entry point.
     Returns:
         dict: A dictionary of method argument names and values.
     """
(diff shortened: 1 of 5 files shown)


src/sagemaker/session.py
```diff
@@ -4249,6 +4253,15 @@ def get_model_package_args(
...
         }
+        if entry_point is not None:
+            # This is a script mode model and its container def wasn't set
+            # (this happens when an estimator is passed to RegisterModelStep instead
+            # of a model). Set the env vars needed for script mode.
+            env_map = {
+                DIR_PARAM_NAME: "/opt/ml/model/code",
+                SCRIPT_PARAM_NAME: os.path.basename(entry_point),
+            }
+            container["Environment"] = env_map
```
qidewenwhen, line 4264: Seems we're appending the env vars in `session.py` so I have some concerns on this:
- Once we release the new interface ModelStep and deprecate the RegisterModel, this logic will be redundant
- As we're changing outside of the Pipeline scope, it may break some of the model registry without pipeline use like:
  - For TensorFlowModel, even if `entry_point` is given, it does not append DIR_PARAM_NAME or SCRIPT_PARAM_NAME in the env vars (see [here](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/tensorflow/model.py#L337)
  - For Model type,  if only `entry_point` is given without the `source_dir`, the `DIR_PARAM_NAME` is a s3 path. However, here we override it to be a local path. This may not work for the case outside of pipeline scope.
You can find more env cases in [unit tests here](https://github.com/aws/sagemaker-python-sdk/pull/3076/files#diff-2019ff8a82e7a8af31bee2df6c7df55901496e5cb8b8b383f4c7b6a9b051000eR686-R772)

Note: if the link does not work, please check the parameters of `test_request_compare_of_register_model_under_different_sessions` in `test_model_step.py`

qidewenwhen: Have discussed this with Zhankui previously, as the to be deprecated `RegisterModel` is not well defined and it's bothering us with endless patching, can we simply remove its entry for user script input? 
If there're any user reports this, we can direct them to the ModelStep with the public doc prepared by Jerry.

Sorry that I didn't engage you when reviewing the ModelStep given the tight timeline.

staubhp: Those are good callouts; if we go this route we'll need to add logic for all the various arg combinations and frameworks.

I don't think we should break backwards compatibility by removing any args though. The script mode still works in most cases, but it's reasonable to prefer deprecating this and moving to ModelStep instead of spending time fixing (the latest) issue with it.

I see your ModelStep PR is ready for merge. Let's go that route instead of spending more time fixing RegisterModel step.

PR #3158, opened 2022-06-03
fix: Fix processing image uri param

jerrypeng7773:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

- [ ] I certify that the changes I am introducing will be backword compatible, and I have discussed concerns about this, if any, with the Python SDK team

#### Tests


src/sagemaker/utils.py
```diff
@@ -99,9 +100,17 @@ def base_name_from_image(image):
     Returns:
         str: Algorithm name, as extracted from the image name.
     """
-    m = re.match("^(.+/)?([^:/]+)(:[^:]+)?$", image)
-    algo_name = m.group(2) if m else image
-    return algo_name
+    if is_pipeline_variable(image):
+        if is_pipeline_parameter_string(image) and image.default_value:
+            image_str = image.default_value
```
qidewenwhen, line 105: The base_name_from_image will appear in s3 uri eventually.
- How about we simply return "base_name" directly in terms of pipeline variable? As the image_uri can be changed in execution time, then s3 uri generated as per image is not making sense and may cause confusions.
- Plus as the base_name_from_image would show up in s3 uri, can we make "base_name" more meaningful? like "processing-job" etc.

tests/unit/sagemaker/workflow/test_processing_step.py
```diff
@@ -328,17 +328,27 @@ def test_processing_step_with_processor(pipeline_session, processing_input):
...
+    [
+        IMAGE_URI,
+        ParameterString(name="MyImage"),
+        ParameterString(name="MyImage", default_value="my-image-uri"),
+        Join(on="/", values=["docker", "my-fake-image"]),
+    ],
+)
+def test_processing_step_with_processor_and_step_args(
+    pipeline_session, processing_input, image_uri
+):
```
qidewenwhen, line 342: This test is the fault case regarding the exclusive relationship between `processor` and `step_args`. 
Why we add the image_uri test case here?

jerrypeng7773: not following ...

src/sagemaker/utils.py
jerrypeng7773: > How about we simply return "base_name" directly in terms of pipeline variable?

Make sense. However, I wonder the frequency of the change of the `image_uri`, if the underlying algo. does not change, the necessity of change the `image_uri` is not obvious.

> Plus as the base_name_from_image would show up in s3 uri, can we make "base_name" more meaningful? like "processing-job" etc.

Yea we probably should do this. Thought about this before, this would make the code looks ugly, basically, where we call this function (from estimator, processor, model etc), we will need to pass an additional class name, and it is only used if the image name value given is a pipeline variable.

tests/unit/sagemaker/workflow/test_processing_step.py
qidewenwhen: I meant usually we would separate the tests of happy case (e.g. parameterized image uri) and the failed case (e.g. the invalid input arguments of ProcessingStep) in different tests. But that's fine. It's just a Nit suggestion.

src/sagemaker/utils.py
```diff
@@ -90,18 +91,27 @@ def unique_name_from_base(base, max_length=63):
...
 
     Returns:
         str: Algorithm name, as extracted from the image name.
     """
-    m = re.match("^(.+/)?([^:/]+)(:[^:]+)?$", image)
-    algo_name = m.group(2) if m else image
-    return algo_name
+    if is_pipeline_variable(image):
+        if is_pipeline_parameter_string(image) and image.default_value:
+            image_str = image.default_value
```
navaj0, line 106: The default value could be None, empty strings or garbage strings. What the constructed base name in this case?

navaj0 (commented):
The workaround looks hacky. We need to rethink about the s3 artifacts layout under pipeline context. For example, we should use the pipeline names, step names. But that is a much bigger problem to solve.

src/sagemaker/utils.py
jerrypeng7773: if parameter default is None or empty string, it will be the `default_base_name` 

if parameter default is garbage string, it will be the garbage string.

navaj0 (approved):

mufaddal-rohawala (approved):

navinsoni (approved):

Issue #2910, opened 2022-02-08
`s3:ListAllMyBuckets` or `s3:CreateBucket` permission required for Session with default bucket


janbe-ts:
**Describe the bug**
When I create a Session like this: `Session(default_bucket=my_bucket)`, the library conveniently creates the bucket if it doesn't already exists. To check whether the bucket exists is based on getting the creation date of the bucket which is an operation that requires the IAM permission `s3:ListAllMyBuckets`. This permission shouldn't be necessary and is often not allowed in environments following the least privileges principle.

**To reproduce**
Create a Session with a `default_bucket` that already exists. Creating any Sagemaker processing job will fail if the used IAM role does not have `s3:ListAllMyBuckets` and `s3:CreateBucket` permission. 

**Expected behavior**
I should be able to create a Sagemaker processing job without having to assign `s3:ListAllMyBuckets` or `s3:CreateBucket` to my IAM role.

**System information**
A description of your system. Please provide:
- **SageMaker Python SDK version**: 2.75.1
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: PySparkProcessor
- **Framework version**: 2.4
- **Python version**: 3.9.4
- **CPU or GPU**: CPU
- **Custom Docker image (Y/N)**: Y

**Additional context**
This code in the function `_create_s3_bucket_if_it_does_not_exist` in `session.py` creates the bucket if it doesn't already exists:
```
if bucket.creation_date is None:
  try:
      if region == "us-east-1":
          # 'us-east-1' cannot be specified because it is the default region:
          # https://github.com/boto/boto3/issues/125
          s3.create_bucket(Bucket=bucket_name)
      else:
          s3.create_bucket(
              Bucket=bucket_name, CreateBucketConfiguration={"LocationConstraint": region}
          )
  
      LOGGER.info("Created S3 bucket: %s", bucket_name)
  except ClientError as e:
      error_code = e.response["Error"]["Code"]
      message = e.response["Error"]["Message"]
  
      if error_code == "BucketAlreadyOwnedByYou":
          pass
      elif (
          error_code == "OperationAborted"
          and "conflicting conditional operation" in message
      ):
          # If this bucket is already being concurrently created, we don't need to create
          # it again.
          pass
      else:
          raise
```

[...]

PR #3165, opened 2022-06-09
fix: bucket exists check for session.default_bucket

mufaddal-rohawala:
*Issue #, if available:*
https://github.com/aws/sagemaker-python-sdk/issues/2910

*Description of changes:*
Current check for bucket existence in `session.py` requires `s3:ListAllMyBuckets`. This permission shouldn't be necessary and is often not allowed in environments following the least privileges principle. Also Without `s3:ListAllMyBuckets` permissions the case of bucket existence, still requires `s3:CreateBucket` perms otherwise the check doesn't work properly.

Expected behavior
Users should be able to create a Sagemaker jobs without having to assign s3:ListAllMyBuckets or s3:CreateBucket to the IAM role. 

With this change users can now use a `s3:ListBucket` on a specific bucket, which can support the case of same or cross account specific bucket access.

*Testing done:*
For backwards compatibility the creation_date check for `s3:ListAllMyBuckets` would also be supported.
Added units

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/session.py
+++ b/src/sagemaker/session.py
@@ -412,29 +412,47 @@ class Session(object):  # pylint: disable=too-many-public-methods
         bucket = s3.Bucket(name=bucket_name)
         if bucket.creation_date is None:
             try:
-                if region == "us-east-1":
-                    # 'us-east-1' cannot be specified because it is the default region:
-                    # https://github.com/boto/boto3/issues/125
-                    s3.create_bucket(Bucket=bucket_name)
-                else:
-                    s3.create_bucket(
-                        Bucket=bucket_name, CreateBucketConfiguration={"LocationConstraint": region}
-                    )
-
-                LOGGER.info("Created S3 bucket: %s", bucket_name)
+                # trying head bucket call
+                s3.meta.client.head_bucket(Bucket=bucket.name)
             except ClientError as e:
+                # bucket does not exist or forbidden to access
                 error_code = e.response["Error"]["Code"]
                 message = e.response["Error"]["Message"]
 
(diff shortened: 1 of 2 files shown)


src/sagemaker/session.py
```diff
@@ -412,29 +412,47 @@ def _create_s3_bucket_if_it_does_not_exist(self, bucket_name, region):
...
-                if error_code == "BucketAlreadyOwnedByYou":
-                    pass
-                elif (
-                    error_code == "OperationAborted"
-                    and "conflicting conditional operation" in message
-                ):
-                    # If this bucket is already being concurrently created, we don't need to create
-                    # it again.
-                    pass
+                if error_code == "404" and message == "Not Found":
```
BasilBeirouti, line 422: nit: no need for the message check. error_code == "404" is enough. Based on this documentation :
https://docs.aws.amazon.com/cli/latest/reference/s3api/head-bucket.html

_If the bucket does not exist or you do not have permission to access it, the HEAD request returns a generic `404 Not Found` or `403 Forbidden` code. A message body is not included, so you cannot determine the exception beyond these error codes._

BasilBeirouti:
Thank you @mufaddal-rohawala for writing such a clear description of the problem and for linking the PR issue

src/sagemaker/session.py
mufaddal-rohawala: yeah agreed only error code should be fine, but I think this way the code is more verbose as to actual error message response, this would help better for debugging in future. Having both checks shouldn't hurt?

shreyapandit (approved):

BasilBeirouti (approved):

Issue #3141, opened 2022-05-27
Exception: instance_type should not be a pipeline variable (<class 'sagemaker.workflow.parameters.ParameterString'>)


qidewenwhen:
**Describe the bug**
After the release https://github.com/aws/sagemaker-python-sdk/releases/tag/v2.92.0, pipelines which was working before started failing with this exception, which relates to [this commit](https://github.com/aws/sagemaker-python-sdk/commit/19b5560ab0811d3833d3132acc630fa9869ffacd)

```
Exception: instance_type should not be a pipeline variable (<class 'sagemaker.workflow.parameters.ParameterString'>)
```

**Note**
This is a breaking change which can break some existing customer codes, so open this issue, post the explanation and fix options here, in case any one searches for it.

qidewenwhen:
**The reason for the breaking change**
The error is thrown when we pass in any kind of pipeline variables e.g. `Parameter`, `Properties `etc. to the `image_uris.retrieve()`
The pipeline variables are placeholders which are parsed in execution time but the `image_uris.retrieve()`is retrieving `image_uris` in compile time (PySDK stage). In other words, the parameterization here won't work at all. And we have received several internal/external customers complaining on this e.g. [GitHub issue](https://github.com/aws/sagemaker-python-sdk/issues/3024).
Thus we decided to explicitly forbid passing in any pipeline variables to `image_uris.retrieve()`.

**To resolve this issue**
- If you directly put any pipeline variables to `image_uris.retrieve()`, please replace them with python primitives e.g. plain string
- If you see this error thrown from any estimator/processor, that's because `image_uri` is not provided (i.e. None) here so that the `instance_type` is used to retrieve `image_uri` under the hood. The quick fix can be either:
  - Provide `image_uri` to the estimator/processor 
  - Or **Just simply update the `instance_type` to be a plain string** (This can be an easier fix)

PR #3180, opened 2022-06-18
change: Add override_pipeline_parameter_var decorator to give grace period to update invalid pipeline var args

qidewenwhen:
*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/3141

*Description of changes:*
- Add `override_pipeline_parameter_var` decorator to `image_uris.retrieve` to give grace period for users to remove passing pipeline variables to this method
  - Note this only provides grace period to passing `ParameterString` with default value to the `image_uris.retrieve` which are all the issue reports we've received. And for other types of pipeline variables, they did not work previously so should not be impacted by the Git issue 3141.
- Update the doc string of the `definition` field of `ProcessingOutput` as it's not correct in Pipeline context.

*Testing done:* Unit tests

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/processing.py
+++ b/src/sagemaker/processing.py
@@ -1195,7 +1195,9 @@ class ProcessingOutput(object):
             source (str): The source for the output.
             destination (str): The destination of the output. If a destination
                 is not provided, one will be generated:
-                "s3://<default-bucket-name>/<job-name>/output/<output-name>".
+                "s3://<default-bucket-name>/<job-name>/output/<output-name>"
+                (Note: this does not apply when used with
+                :class:`~sagemaker.workflow.steps.ProcessingStep`).
             output_name (str): The name of the output. If a name
                 is not provided, one will be generated (eg. "output-1").
             s3_upload_mode (str): Valid options are "EndOfJob" or "Continuous".
--- a/src/sagemaker/workflow/utilities.py
+++ b/src/sagemaker/workflow/utilities.py
@@ -13,23 +13,29 @@
 """Utilities to support workflow."""
 from __future__ import absolute_import
 
+import inspect
+import logging
+from functools import wraps
 from pathlib import Path
-from typing import List, Sequence, Union, Set
+from typing import List, Sequence, Union, Set, TYPE_CHECKING
 import hashlib
 from urllib.parse import unquote, urlparse
 from _hashlib import HASH as Hash
 
+from sagemaker.workflow.parameters import Parameter
 from sagemaker.workflow.pipeline_context import _StepArguments
-from sagemaker.workflow.step_collections import StepCollection
 from sagemaker.workflow.entities import (
     Entity,
     RequestType,
 )
 
+if TYPE_CHECKING:
+    from sagemaker.workflow.step_collections import StepCollection
+
 BUF_SIZE = 65536  # 64KiB
 
 
-def list_to_request(entities: Sequence[Union[Entity, StepCollection]]) -> List[RequestType]:
+def list_to_request(entities: Sequence[Union[Entity, "StepCollection"]]) -> List[RequestType]:
     """Get the request structure for list of entities.
 
     Args:
@@ -37,6 +43,8 @@ def list_to_request(entities: Sequence[Union[Entity, StepCollection]]) -> List[R
     Returns:
         list: A request structure for a workflow service call.
     """
+    from sagemaker.workflow.step_collections import StepCollection
+
     request_dicts = []
     for entity in entities:
         if isinstance(entity, Entity):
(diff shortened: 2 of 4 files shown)


src/sagemaker/processing.py
```diff
@@ -1195,7 +1195,9 @@ def __init__(
             source (str): The source for the output.
             destination (str): The destination of the output. If a destination
                 is not provided, one will be generated:
-                "s3://<default-bucket-name>/<job-name>/output/<output-name>".
+                "s3://<default-bucket-name>/<job-name>/output/<output-name>"
+                (Note: this does not apply when used with
+                :class:`~sagemaker.workflow.steps.ProcessingStep`).
```
jerrypeng7773, line 1200: what will be used if its pipeline session, or processing step?

src/sagemaker/workflow/utilities.py
```diff
@@ -13,30 +13,37 @@
...
 
+from sagemaker.workflow.parameters import Parameter
 from sagemaker.workflow.pipeline_context import _StepArguments
-from sagemaker.workflow.step_collections import StepCollection
 from sagemaker.workflow.entities import (
     Entity,
     RequestType,
 )
 
+if TYPE_CHECKING:
```
jerrypeng7773, line 31: interesting, so `type_checking` resolves the circular import error from importing `StepCollection` here?

src/sagemaker/workflow/utilities.py
```diff
@@ -151,3 +158,41 @@ def validate_step_args_input(
         raise TypeError(error_message)
     if step_args.caller_name not in expected_caller:
         raise ValueError(error_message)
+
+
+def override_pipeline_parameter_var(func):
+    """A temporary Decorator
```
jerrypeng7773, line 164: we might need this in the future too, but we can add it back anytime when we need it.

src/sagemaker/workflow/utilities.py
```diff
@@ -151,3 +158,41 @@ def validate_step_args_input(
...
+    This is a temporary decorator to override pipeline Parameter objects with their default value
+    and display warning information to instruct users to update their code.
+
+    This decorator can help to give a grace period for users to update their code when
+    we make changes to explicitly prevent passing any pipeline variables to a function.
+
+    We should remove this decorator after the grace period.
+    """
+    warning_info_suffix = (
+        "The default_value of this Parameter object will be used to override it. "
```
jerrypeng7773, line 175: what if the `default_value` is null?

src/sagemaker/workflow/utilities.py
```diff
@@ -151,3 +158,41 @@ def validate_step_args_input(
...
+        "Please remove this pipeline variable and use python primitives instead."
+    )
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        args = list(args)
+        for idx, arg in enumerate(args):
+            if isinstance(arg, Parameter):
+                warn_msg = (
+                    "Pipeline variable (type: %s) is not allowed here. " + warning_info_suffix
```
jerrypeng7773, line 185: can we print out the actual name of the arg that we are preventing being a pipeline param here? like `instance_type`

src/sagemaker/workflow/utilities.py
```diff
@@ -151,3 +158,41 @@ def validate_step_args_input(
...
+                )
+                logging.warning(warn_msg, type(arg))
+                args[idx] = arg.default_value
+        args = tuple(args)
+
+        for key, value in kwargs.items():
+            if isinstance(value, Parameter):
+                warn_msg = "%s should not be a pipeline variable (%s). " + warning_info_suffix
```
jerrypeng7773, line 193: nit: better have warning message having the same pattern here and above.

src/sagemaker/workflow/utilities.py
qidewenwhen: Yes

src/sagemaker/workflow/utilities.py
qidewenwhen: Then it's equivalent to pass in a None to the field, which should break the further part of retrieve code.
Note, a ParameterString without default_value could also break the retrieve code previously (before the adding explicit validation change)

src/sagemaker/workflow/utilities.py
qidewenwhen: Yea... I'm thinking about maybe we can keep this annotation definition here but removing the annotation on top of `image_uris.retrieve` once users are ready to accept the breaking change

src/sagemaker/processing.py
qidewenwhen: It should be `s3://<default-bucket-name>/<job-name created by the processor class>/output/<output-name>`

The `job-name created by the processor class` here will be popped out from the DSL and is not the actual job name (created in Pipeline backend). To avoid confusing users, I don't give this path here.

src/sagemaker/workflow/utilities.py
qidewenwhen: Updated in the new commit

src/sagemaker/workflow/utilities.py
qidewenwhen: Updated in the new commit

jerrypeng7773 (approved):

nmadan (approved):

src/sagemaker/workflow/utilities.py
```diff
@@ -151,3 +159,41 @@ def validate_step_args_input(
...
+    """
+    warning_msg_template = (
+        "%s should not be a pipeline variable (%s). "
+        "The default_value of this Parameter object will be used to override it. "
+        "Please remove this pipeline variable and use python primitives instead."
+    )
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        params = inspect.signature(func).parameters
```
mufaddal-rohawala, line 183: nit: Is there a reason for this? can't we enumerate over `args` directly?

qidewenwhen: Yes, the reason is args only can not give us the name of the argument, so that we're not able to specify the invalid argument name in the warning message.
`Args` combined with `.signature` can provide us both the arg name and the arg value.

mufaddal-rohawala: ah, got you

mufaddal-rohawala (approved):

BasilBeirouti (approved):

PR #3235, opened 2022-07-13
feature: enhance-bucket-override-support

evakravi:
*Issue #, if available:*

*Description of changes:*
This PR adds support for overriding the bucket which hosts model and script artifacts (the bucket from which metadata is retrieved is either set by the region, or a separate, pre-existing environment variable override). 

In addition, this PR provides support for reading metadata files from the local filesystem using environment variable overrides, as oppose to the default of S3.

*Testing done:*

## Merge Checklist

#### General

- [ ] I certify that the changes I am introducing will be backword compatible, and I have discussed concerns about this, if any, with the Python SDK team

#### Tests


JGuinegagne (commented):
Just want to double check that this will work with our test engine.

Right now, our test engine saves the SDK metadata with this directory [structure](https://github.com/aws/sagemaker-model-hub/tree/main/src/sagemakermodelhub/content_generation/metadata/generated_metadata/sdk):
|_`manifests`
|__`models_manifest.json`
|-`specs`
|--`community_models`
|---`<model-id>`
|----`specs_<version>.json`
|---- ...

However, the buckets have this pseudo dir structure:
|_`models_manifest.json`
|-`community_models`
|--`<model-id>`
|---`specs_<version>.json`
|--- ...

Will the SDK be able to (1) retrieve the manifests (2) retrieve the specs based with only the link to the root dir?

src/sagemaker/jumpstart/types.py
```diff
@@ -65,7 +65,7 @@ def __str__(self) -> str:
         {'content_bucket': 'bucket', 'region_name': 'us-west-2'}"
         """
 
-        att_dict = {att: getattr(self, att) for att in self.__slots__}
+        att_dict = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
```
JGuinegagne, line 68: curious: why do you need this now?

JGuinegagne (approved):

src/sagemaker/jumpstart/types.py
evakravi: this bug got exposed after i fixed a unit test

src/sagemaker/jumpstart/accessors.py
```diff
@@ -78,6 +78,22 @@ def _set_cache_and_region(region: str, cache_kwargs: dict) -> None:
...
+            ValueError: If region in `cache_kwargs` is inconsistent with `region` argument.
+
+        Args:
+            region (str): Optional. The region to use for the cache.
+        """
+        cache_kwargs = JumpStartModelsAccessor._validate_and_mutate_region_cache_kwargs(
+            JumpStartModelsAccessor._cache_kwargs, region
+        )
+        JumpStartModelsAccessor._set_cache_and_region(region, cache_kwargs)
+        return JumpStartModelsAccessor._cache.get_manifest()  # type: ignore
```
navinsoni, line 95: Is this backward compatible change. looks like you removed a variable from this function.

src/sagemaker/jumpstart/cache.py
```diff
@@ -235,7 +238,62 @@ def _get_manifest_key_from_model_id_semantic_version(
...
+        key: str,
+        filetype: JumpStartS3FileType
+    ) -> Tuple[Union[dict, list], Optional[str]]:
+        """Returns json file either from s3 or local file system.
+
+        Returns etag along with json object for s3, or just the json
+        object and None when reading from the local file system.
+        """
+        if self._is_local_metadata_mode():
+            return self._get_json_file_from_local_override(key, filetype), None
```
navinsoni, line 264: Do we always want to send second variable as none here

src/sagemaker/jumpstart/cache.py
```diff
@@ -235,7 +238,62 @@ def _get_manifest_key_from_model_id_semantic_version(
...
+        return self._get_json_file_and_etag_from_s3(key)
```
navinsoni, line 265: It's not very clear here that you are receiving 2 args and passing it here. Can you use variables here to get those and then return them

navinsoni (changes requested):

src/sagemaker/jumpstart/accessors.py
evakravi: Good call-out. This isn't backward compatible. However, this function is only invoked internally and is not referenced in any public documentation or notebooks. So I don't think this is a risk. But thank you for raising my awareness of these issues.

src/sagemaker/jumpstart/cache.py
evakravi: Yes, because when Json metadata files are retrieved locally from the file system, we don't need to obtain their etag.

src/sagemaker/jumpstart/cache.py
evakravi: yes, good suggestion

src/sagemaker/jumpstart/accessors.py
navinsoni: if it's internal then function should start with `_` as convention. Can you please change it. 
Also it's a static method so anyone from outside can use it. can you please remove static decorator from it.

src/sagemaker/jumpstart/cache.py
navinsoni: sounds good

src/sagemaker/jumpstart/accessors.py
evakravi: Sorry, this isn't internal. I will callout this is not backwards compatible in the PR description.

navinsoni: I am not comfortable merging a backward incompatible change, you can have it none by default. if it’s not passed then you can get that value.

evakravi: Got it. I have made this implementation backwards compatible.

navinsoni (approved):

PR #3268, opened 2022-07-28
feature: Added a print statement for Analysis config in clarify

dosatos:
*Issue #, if available:*

*Description of changes:*
Added a print statement for Analysis config in clarify

*Testing done:*
no testing. I simply added a print statement.

## Merge Checklist

#### General

#### Tests


larroy (commented):
let's not print. Let's make a method to return it as a string and decouple concerns.

keerthanvasist (commented):
@larroy what is the intent here? We write the analysis config to S3 already. I don't think we should be returning the analysis config (least of all as a string). If anything, I think we should return the processing job ID if we can. This can be useful for the customer when they are trying to automate it. 

Currently, customer can only know the processing job name from the log. Instead, it would be good if we can expose the name of the processing job or return it.

larroy:
I think is good to have access to the analysis config file locally, is already there and also makes the code more decoupled allowing us better testability and problem diagnosis. Agree with processing job ID we should expose that as well.

dosatos:
@keerthanvasist @larroy thank you for the feedback

The problem with returning the `analysis_config` (i.e. storing the value inside the instance) is that is is different for `run_bias`, `run_pre_training_bias`, `run_post_training_bias`, `run_explainability`. Thus, mutates each time we alternate the method. I can imaging a case when, a user runs `run_explainability`, and then `run_bias`, and then starts to debug why `run _explainability` mis-behaved by looking into `obj.analysis_config` (which is different already).

### Option 1 [not recommended]
Store analysis config for the latest executed job.

```python
@property
def analysis_config(self):
    return self.__analysis_config
```

- Advtangages: Simple to implement.
- Disadvantages: It is confusing as the config would mutate after we run different methods  (`run_bias`, `run_pre_training_bias`, `run_post_training_bias`, `run_explainability)

### Option 2:
Given this, the simplest and the most accurate is to log the `analysis_config` the `SageMakerClarifyProcessor` used to `_run` the workload. 

Example:
```
Job Name:  Clarify-Bias-2022-07-22-09-00-31-294
Inputs:  [{...}]
Outputs:  [...}]
```

- Advantages: simple to implement - solves the debugging needs. 
- Distadvantages: it polutes the std output (perhaps not a concern, as we already do inside `Processor` when print Job)

### Option 3 [recommended]
Another option is to build a factory method for the analysis configs. 
- Advantages: 
	- 1) helps debugging needs
	- 2) improves code - separation of concerns
- Disadvantages: when customer debugs, does not guarantee the person would feed the same DataConfig/Model to the Factory as to the `SageMakerClarifyProcessor` (chances are low, but possible)

PR #3271, opened 2022-08-01
feature: added _AnalysisConfigGenerator for clarify

dosatos:
*Issue #, if available:*

*Description of changes:*
- Added `_AnalysisConfigGenerator` class to simplify maintainability of the config creation. 
    - Also, with this generator, customers may reproduce the state of the config locally during debugging (i.e. without a need to opening it in the S3 bucket)
- Simplified `job_name` creation part
- Added a private method `_last_analysis_config`) to store last analysis_config used for a run (to be useful in case of debugging)
- Extended the std out with the analysis config used for a run. Currently it already prints other values, like Job Name, Inputs, and Outputs:

```
Job Name:  Clarify-Bias-2022-07-22-09-00-31-294
Inputs:  [{...}]
Outputs:  [...}]
```

*Testing done:*
Unit Tests

## Merge Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your pull request._

#### General

- [x] I have read the [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md) doc
- [x] I certify that the changes I am introducing will be backword compatible, and I have discussed concerns about this, if any, with the Python SDK team
- [x] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#committing-your-change)
- [ ] I have passed the region in to all S3 and STS clients that I've initialized as part of this change.
- [ ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/tree/master/doc) (if appropriate)

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [x] I have added unit and/or integration tests as appropriate to ensure backward compatibility of the changes
- [x] I have checked that my tests are not configured for a specific region or account (if appropriate)
- [x] I have used [`unique_name_from_base`](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/utils.py#L77) to create resource names in integ tests (if appropriate)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3282, opened 2022-08-05
Added bias_and_explainability to SageMakerClarifyProcessor

dosatos:
*Issue #, if available:*

*Description of changes:*
Added a public method to run both Bias and Explainability in a single job processing

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/clarify.py
+++ b/src/sagemaker/clarify.py
@@ -25,7 +25,7 @@ import re
 
 import tempfile
 from abc import ABC, abstractmethod
-from typing import List, Union
+from typing import List, Union, Dict
 
 from sagemaker import image_uris, s3, utils
 from sagemaker.processing import ProcessingInput, ProcessingOutput, Processor
@@ -172,7 +172,11 @@ class DataConfig:
         _set(joinsource, "joinsource_name_or_index", self.analysis_config)
         _set(facet_dataset_uri, "facet_dataset_uri", self.analysis_config)
         _set(facet_headers, "facet_headers", self.analysis_config)
-        _set(predicted_label_dataset_uri, "predicted_label_dataset_uri", self.analysis_config)
+        _set(
+            predicted_label_dataset_uri,
+            "predicted_label_dataset_uri",
+            self.analysis_config,
+        )
         _set(predicted_label_headers, "predicted_label_headers", self.analysis_config)
         _set(predicted_label, "predicted_label", self.analysis_config)
         _set(excluded_columns, "excluded_columns", self.analysis_config)
@@ -491,7 +495,10 @@ class PDPConfig(ExplainabilityConfig):
             top_k_features (int): Sets the number of top SHAP attributes used to compute
                 partial dependence plots.
         """  # noqa E501
-        self.pdp_config = {"grid_resolution": grid_resolution, "top_k_features": top_k_features}
+        self.pdp_config = {
+            "grid_resolution": grid_resolution,
+            "top_k_features": top_k_features,
+        }
         if features is not None:
             self.pdp_config["features"] = features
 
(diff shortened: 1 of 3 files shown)


src/sagemaker/clarify.py
```diff
@@ -1307,56 +1290,176 @@ def run_explainability(
...
+            data_config,
+            analysis_config,
+            wait,
+            logs,
+            job_name,
+            kms_key,
+            experiment_config,
+        )
+
+    def run_bias_and_explainability(self):
```
keerthanvasist, line 1311: Can we see if we can override the existing run() method for this purpose?

dosatos: `Signature of method 'SageMakerClarifyProcessor.run()' does not match signature of the base method in class 'Processor'`

dosatos:
I will apply formatting after the PR #3271 gets merged - it partially improves the style already

src/sagemaker/clarify.py
```diff
@@ -1370,56 +1367,348 @@ def run_explainability(
...
+
+    @classmethod
+    def bias_and_explainability(
+        cls,
+        data_config: DataConfig,
+        model_config: ModelConfig,
+        model_predicted_label_config: ModelPredictedLabelConfig,
+        explainability_config: Union[ExplainabilityConfig, List[ExplainabilityConfig]],
+        bias_config: BiasConfig,
+        pre_training_methods: Union[str, List[str]] = "all",
```
larroy, line 1534: I'm not a big fan of these Union types, I see the intention, but is it worth it to save the user typing [] ?  I would prefer if it's a plain list and we check that is not empty. Open to discuss about this and be convinced of the opposite though.

dosatos: Are you saying that, instead of making the user to pass `"all"`, we can make it `["all"]`?
Actually, I did it this way to be consistent with the current implementation - i.e. with how we expose methods yet.

src/sagemaker/clarify.py
```diff
@@ -1370,56 +1367,348 @@ def run_explainability(
...
-            data_config,
-            analysis_config,
-            wait,
-            logs,
-            job_name,
-            kms_key,
-            experiment_config,
-        )
+            return explainability_methods
+        if (
```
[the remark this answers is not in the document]
dosatos: then the last line returns it, no? `return explainability_config.get_explainability_config()`

src/sagemaker/clarify.py
larroy: I mean         explainability_config: Union[ExplainabilityConfig, List[ExplainabilityConfig]],

shreyapandit:
I have restarted PR and Local Mode tests, notebook failures are expected and I can override them. Please have 1 approval from @larroy or @keerthanvasist for us to be able to merge

src/sagemaker/clarify.py
keerthanvasist: So it was initially only `ExplainabilityConfig`, and we only had SHAPConfig. But then we added PDP, and we wanted to let customers be allowed to choose both of them in a single job. So, we made into an array but kept the original single ExplainabilityConfig as well to maintain backward compatibility.

larroy: I see this was like that and you just annotated the same type, thanks for catching up with this.

larroy (approved):

src/sagemaker/clarify.py
```diff
@@ -1368,68 +1383,198 @@ def run_explainability(
...
+            explainability_config=explainability_config,
+        )
+        analysis_config = cls._add_predictor(
+            analysis_config, model_config, model_predicted_label_config
+        )
+        return analysis_config
+
     @classmethod
     def explainability(
         cls,
```
keerthanvasist, line 1551: Just to maintain consistency, can you use `self` instead of `cls`?

keerthanvasist (approved):

dosatos:
The changes in this PR were also present in this one: https://github.com/aws/sagemaker-python-sdk/pull/3296

Thus, cancelling the PR

src/sagemaker/clarify.py
dosatos: Thanks for the comment, Keerthan.`self` implies creation of an instance, `cls` does not require an instance
This should also be driven by [pep8](https://stackoverflow.com/questions/4613000/difference-between-cls-and-self-in-python-classes).

Issue #3413, opened 2022-10-13
Regression in terms of handling the code_location attribute in SM Pipelines


navaj0:
**Describe the bug**

It worked before, but now breaks my code. It worked at least until 2.100.0 and I was kinda hoping it would be fixed without intervention, but it does not work in 2.104.0, 2.109.0 and 2.110.0.

I used to provide a string that ended in a ‘/’, like `f's3://{pipeline_session.default_bucket()}/pipeline-code/' ` But this now results in the below error message when attempting to run a Batch Transform. Please recognize the ‘//’ in the key.

```
ClientError: Failed to download model data(bucket: sagemaker-eu-west-1-426334991426, key: pipeline-code//pytorch-inference-2022-10-12-14-48-22-161/pipelines-otzuzj9grt59-lr-bow-RepackModel-0-bVEhzyl7Ih/output/model.tar.gz). NoSuchKey (HTTP StatusCode: 404)
```

aws s3 ls would still see the model tarball, but not the Batch Transform. Removing the slash from the code location “works” (as in does not create an error message), but then I have no separation between the pipeline prefix and the artifact name, e.g. `pipeline-codepytorchinference ...`

**To reproduce**

Example:

```
teacher_model = PyTorchModel(
    model_data=teacher_training_step.properties.ModelArtifacts.S3ModelArtifacts,
    **common_attributes,
    code_location=f's3://{pipeline_session.default_bucket()}/pipeline-code/',
    source_dir='.',
    entry_point='inference.py',
) 
```
In the first training job config in the pipeline json dump, you will find:
```
'OutputDataConfig': {'S3OutputPath': 's3://sagemaker-eu-west-1-426334991426/pipeline-code//pytorch-inference-2022-10-12-16-00-45-984'},
```

**Expected behavior**

**Screenshots or logs**
If applicable, add screenshots or logs to help explain your problem.

**System information**
A description of your system. Please provide:
- **SageMaker Python SDK version**:
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**:
- **Framework version**:
- **Python version**:
- **CPU or GPU**:
- **Custom Docker image (Y/N)**:

**Additional context**

marianokamp:
I tried with 2.115.0 and it works great now. Thanks a lot!

PR #3427, opened 2022-10-19
fix: adjacent slash in s3 key

nmadan:
*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/3413

*Description of changes:* Change S3 output_path to remove adjacent slashes

*Testing done:* unit testing

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/tensorflow/model.py
+++ b/src/sagemaker/tensorflow/model.py
@@ -396,8 +396,8 @@ class TensorFlowModel(sagemaker.model.FrameworkModel):
             # model is not yet there, defer repacking to later during pipeline execution
             if isinstance(self.sagemaker_session, PipelineSession):
                 self.sagemaker_session.context.need_runtime_repack.add(id(self))
-                self.sagemaker_session.context.runtime_repack_output_prefix = "s3://{}/{}".format(
-                    bucket, key_prefix
+                self.sagemaker_session.context.runtime_repack_output_prefix = s3.s3_path_join(
+                    "s3://", bucket, key_prefix
                 )
             else:
                 logging.warning(
--- a/src/sagemaker/model.py
+++ b/src/sagemaker/model.py
@@ -569,8 +569,8 @@ api/latest/reference/services/sagemaker.html#SageMaker.Client.add_tags>`_
                     )
                     return
                 self.sagemaker_session.context.need_runtime_repack.add(id(self))
-                self.sagemaker_session.context.runtime_repack_output_prefix = "s3://{}/{}".format(
-                    bucket, key_prefix
+                self.sagemaker_session.context.runtime_repack_output_prefix = s3.s3_path_join(
+                    "s3://", bucket, key_prefix
                 )
                 # Add the uploaded_code and repacked_model_data to update the container env
                 self.repacked_model_data = self.model_data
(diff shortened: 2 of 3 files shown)


src/sagemaker/tensorflow/model.py
```diff
@@ -396,8 +396,8 @@ def prepare_container_def(
             # model is not yet there, defer repacking to later during pipeline execution
             if isinstance(self.sagemaker_session, PipelineSession):
                 self.sagemaker_session.context.need_runtime_repack.add(id(self))
-                self.sagemaker_session.context.runtime_repack_output_prefix = "s3://{}/{}".format(
-                    bucket, key_prefix
+                self.sagemaker_session.context.runtime_repack_output_prefix = s3.s3_path_join(
+                    "s3://", bucket, key_prefix
                 )
```
qidewenwhen, line 401: Seems the `s3.s3_path_join` will remove the "/" in the end of the path and return "s3://my-bucket/my-key-prefix".

But given the customer input in https://github.com/aws/sagemaker-python-sdk/issues/3413
```
Removing the slash from the code location “works” (as in does not create an error message), but then I have no separation between the pipeline prefix and the artifact name, e.g. pipeline-codepytorchinference ...
```
Should we also check the estimator side on how it generates the entire code output path with the `code_location`?

Or if this is already checked in the unit test that a separation is correctly added between the prefix and the artifact name, please let me know. Then we're good

qidewenwhen: Plus, if we decide to update the model side rather than the estimator side, as this only updates the tensorflow model, same changes are needed in the model class as well:
- src/sagemaker/model.py: https://github.com/aws/sagemaker-python-sdk/pull/3257/files#diff-25e5d90876e4e075cba7fe8ae004e6c43d300faeecc2d076d1babacc5c0f6d06

nmadan: >Or if this is already checked in the unit test that a separation is correctly added between the prefix and the artifact name, please let me know. Then we're good
yep, I verified in the unit test that the s3 path will be `s3://bucket/pipeline-code/pytorchinference/...`. Basically this code change will convert the `//` to `/`
>same changes are needed in the model class as well
Yes, forgot to commit that class adding it now

qidewenwhen: After a second thought, how about we directly update the joins in `_stage_user_code_in_s3` in estimator.py (see [here](https://github.com/aws/sagemaker-python-sdk/blob/322b919c3b7eaee94adf2a1d9df6e25e756ce55b/src/sagemaker/estimator.py#L808))? The reasons are:
- even if we update the model side to remove the "/" at the end, the `code_location//some-key` would still persist in estimator code output. Though it may not cause any errors but it may not work as customer's expectation.
- If update the estimator side, we can just update in one place and no need to update the model sides.

nmadan: I am assuming that by "artifact" the user means "model-name"

qidewenwhen (approved):

src/sagemaker/tensorflow/model.py
qidewenwhen: > how about we directly update the joins in _stage_user_code_in_s3 in estimator.py

Synced up offline, this may break backward compatibility. Given that we don't see any customer complains on the estimator's code location output path, we can just simply update the model side

navaj0 (approved):

claytonparnell (commented):
/bot run slow-tests, notebook-tests

claytonparnell (commented):
/bot run slow-tests

mufaddal-rohawala (approved):

PR #3695, opened 2023-03-06
feat: Add deployment support for deployment recommendations

jinpengqi:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


jinpengqi:
> 

Good point. For model name passed in case, since we'll create model for customer so we could check the results and store in `self.instant_recommendations`. But we might also need to inform customer who is invoking IR job with modelpkgarn that `self.instant_recommendations` is null since some customers might just create modelpkg directly. 
From customer perspective, It should be extra info(optional to check)when invoking `right_size` since
1. they dont necessarily need to call `right_size()` for instant recommendation results, 
2. when customer would like to call `right_size()` it usually means customer tries to invoke default inference recommendation jobs

gwang111:
got it. So we could have the option in `right_size()` but the main intention is to do instant recommendation directly from the `deploy()` method from `recommendation_id`?

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -346,26 +340,66 @@ def _update_params_for_recommendation_id(
...
+        # Retrieve model or inference recommendation job details
+        recommendation_res, model_res = None, None
+        try:
+            recommendation_res = sage_client.describe_inference_recommendations_job(
+                JobName=recommendation_job_name
+            )
+        except sage_client.exceptions.ResourceNotFound:
+            pass
+        try:
+            model_res = sage_client.describe_model(ModelName=recommendation_job_name)
```
jbarz1, line 353: we don't need to make this call if `describe_inference_recommendations_job` works right?

May be we can create three private methods to modularize the code since this method is getting big.

```
def _get_model_recommendation_id(self):
   # get recommendation id from model

def _get_right_size_recommendation_id(self):
  # get recommendation id from right size job

def _get_recommendation_id(self):
  try:
    _get_right_size_recommendation_id(self)
  except:
    _get_model_recommendation_id(self)

def _update_params_for_recommendation_id(...):
  try:
    recommendation_id = _get_recommendation_id(self)
  except:
    raise ValueError("Inference Recommendation id is not valid")
```

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -346,26 +340,66 @@ def _update_params_for_recommendation_id(
...
+                    rec
+                    for rec in model_res["DeploymentRecommendation"][
+                        "RealTimeInferenceRecommendations"
+                    ]
+                    if rec["RecommendationId"] == inference_recommendation_id
+                ),
+                None,
+            )
+        if inference_recommendation is None and instant_recommendation is None:
+            raise ValueError("Inference Recommendation id does not exist")
```
jbarz1, line 382: id is invalid

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -346,26 +340,66 @@ def _update_params_for_recommendation_id(
...
+
+        # Update params beased on instant recommendation
+        if instant_recommendation:
```
jbarz1, line 385: could we not call it "instant"?

maybe use the names "right_size_recommendation" and "model_recommendation"? 

Because that indicates the source of the recommendation

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: 1. as we discussed before, what if customer's `JobName` and `ModelName` are the same, eg `thanos/abcd1234` and `thanos/abcd5678`, then we need to find out it's inference or model recommendation.   So go through two lists(inference_recommendations and model_recommendations) to get the recommendation is needed. So here I am thinking we need to loop through two lists anyway
2. good point, will create separate methods

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: ack, will modify

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: makes sense, will rename them

jinpengqi:
> got it. So we could have the option in `right_size()` but the main intention is to do instant recommendation directly from the `deploy()` method from `recommendation_id`?

I think we need to confirm if we want customer to use `right_size` to get the instant recommendations experience. Since today I believe the experience is obtained through `create_model`.  And `model.deploy()` should be called when customer already inspect the recommendations results no matter its inference recommendation or instant recommendation so I think we don't leverage `deploy` to do instant recommendations. 

Here is I think we could modify:
we set the fields `instant_recommendations` in  `right_size` result but only for modelName passed in case, since we will create the temp model for customer, the `instant_recommendations ` result should be good to have to be exposed. And we could instruct customers in notebook for this inspection too

@jbarz1 what do you think?

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -513,3 +525,57 @@ def _convert_to_stopping_conditions_json(
...
+        """Describe inference recommendation job and model results"""
+        recommendation_res, model_res = None, None
+        try:
+            recommendation_res = sage_client.describe_inference_recommendations_job(
+                JobName=job_or_model_name
+            )
+        except sage_client.exceptions.ResourceNotFound:
+            pass
+        try:
+            model_res = sage_client.describe_model(ModelName=job_or_model_name)
```
jbarz1, line 575: we should first check the recommendations job results. If the recommendation id is there, then do not call describe_model because it is a wasted API call.

knikure (approved):
/bot run all

knikure (commented):
/bot run all

knikure (commented):
/bot run all

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -370,30 +364,40 @@ def _update_params_for_recommendation_id(
...
-                for rec in recommendation_res["InferenceRecommendations"]
-                if rec["RecommendationId"] == inference_recommendation_id
-            ),
-            None,
-        )
+        # Update params beased on model recommendation
+        if model_recommendation:
+            if initial_instance_count is None:
+                raise ValueError(
+                    "Please specify initial_instance_count with model recommendation id"
```
jbarz1, line 386: prefer same style as messages in [model.py](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/model.py#L1162)

```
Must specify recommendation id and instance count.
```

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -370,30 +364,40 @@ def _update_params_for_recommendation_id(
...
+            self.env.update(model_recommendation["Environment"])
+            instance_type = model_recommendation["InstanceType"]
+            return (instance_type, initial_instance_count)
 
-        if not recommendation:
+        # Update params based on default inference recommendation
+        if bool(instance_type) != bool(initial_instance_count):
             raise ValueError(
-                "inference_recommendation_id does not exist in InferenceRecommendations list"
+                "Please either do not specify instance_type and initial_instance_count"
```
jbarz1, line 395: similar to [this](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/model.py#L1550)

```
instance type and intial instance count are mutually exclusive with recommendation id
```

src/sagemaker/inference_recommender/inference_recommender_mixin.py
```diff
@@ -442,8 +446,10 @@ def _update_params_for_recommendation_id(
                 self.model_data = compilation_res["ModelArtifacts"]["S3ModelArtifacts"]
                 self.image_uri = compilation_res["InferenceImage"]
 
-        instance_type = recommendation["EndpointConfiguration"]["InstanceType"]
-        initial_instance_count = recommendation["EndpointConfiguration"]["InitialInstanceCount"]
+        instance_type = right_size_recommendation["EndpointConfiguration"]["InstanceType"]
```
jbarz1, line 449: what if right_size_recommendation is also None? Won't this fail ?

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: thanks, modified in next rev

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: modified in next rev

src/sagemaker/inference_recommender/inference_recommender_mixin.py
jinpengqi: This case has been caught by [here](https://github.com/aws/sagemaker-python-sdk/pull/3695/files#diff-7dadd41a1b89ec100e812b257480c34979fc3666e68bef718ad3b0876d71bc53R537).
So `_get_recommendation` will ensure either `right_size_recommendation` or `model_recommendation` exists, nevertheless it will throw exception for `invalid recommendation id`

gwang111:
pushed commits on top of this. closing

PR #3703, opened 2023-03-08
feature: EMR step runtime role support

thbrooks22:
*Issue #, if available:* Customer issue [tracked here](https://t.corp.amazon.com/D71640585).

*Description of changes:* Allow `runtime_role` parameter to `EMRStep` step in SageMaker Pipelines. **Note**: I'd like to get this reviewed now, but we cannot merge until the backend changes are available in all new regions. I will keep this PR updated with that info, and engage SDK team as needed with updates.

*Testing done:* Two new unit tests, and one modification to an existing test. 
1. [New]: Check that the pipeline JSON contains `ExecutionRoleArn` fields where appropriate if it's included in the `EMRStep` arguments.
2. [New]: Check that SDK throws exception if `cluster_config` and `execution_role_arn` are both not None.
3. [Existing]: Check that if `ExecutionRoleArn` is not None, then it's not serialized to pipelines JSON.

## Merge Checklist

#### General

#### Tests


knikure:
@thbrooks22 Can you fix the flake8 and black check issues in your code? 
Run `tox -e black-format` and then `./.githooks/pre-push` to run all the checks.

src/sagemaker/workflow/emr_step.py
```diff
@@ -198,9 +208,16 @@ def __init__(
         if cluster_id is not None and cluster_config is not None:
             raise ValueError(ERR_STR_WITH_BOTH_CLUSTER_ID_AND_CLUSTER_CFG.format(step_name=name))
 
+        if execution_role_arn is not None and cluster_config is not None:
```
qidewenwhen, line 211: Seems `execution_role_arn` has to come along with `cluster_id` and the former is used to allow accessing the `cluster_id` right?

If it's true, then can we make the logic here more intuitive? e.g. check that execution_role_arn can be not None only if cluster_id is not None. 

And it'd be good to improve the error message to make it more informative, e.g. "execution_role_arn should not be set if ... since it's used to ...", so that it can educate the users why the proper usage looks like this.

qidewenwhen:
As the PR is public facing and is retrievable after merge, let's avoid adding any internal link here:
> Issue #, if available: Customer issue [tracked here](https://t.corp.amazon.com/D71640585).

src/sagemaker/workflow/emr_step.py
thbrooks22: I think the two ways are equally intuitive here. There are two ways for the customer to make an EMR step: by specifying ID of existing cluster to add steps to, or by specifying config of new cluster to be created with steps. EMR runtime roles are only supported on the former request (where [`AddJobFlowSteps` API](https://docs.aws.amazon.com/emr/latest/APIReference/API_AddJobFlowSteps.html) is used) and not on the latter (where [`RunJobFlow` API](https://docs.aws.amazon.com/emr/latest/APIReference/API_RunJobFlow.html) is used).

How much info should the error message include? Should it contain all the above background?

src/sagemaker/workflow/emr_step.py
```diff
@@ -155,6 +163,7 @@ def __init__(
         depends_on: Optional[List[Union[str, Step, StepCollection]]] = None,
         cache_config: CacheConfig = None,
         cluster_config: Dict[str, Any] = None,
+        execution_role_arn: str = None,
```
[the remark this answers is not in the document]
aoguo64: `step_config` is mapped to [StepConfig](https://docs.aws.amazon.com/emr/latest/APIReference/API_StepConfig.html) in EMR's [AddJobFlowSteps](https://docs.aws.amazon.com/emr/latest/APIReference/API_AddJobFlowSteps.html#EMR-AddJobFlowSteps-request-Steps), it's not a single-step configuration but applies to all steps to run in a single AddJobFlowSteps (though Pipelines does only run one job step per EMR step).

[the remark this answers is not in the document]
aoguo64: @qidewenwhen are you suggesting using kwargs intead?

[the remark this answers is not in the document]
qidewenwhen: I see, then `step_config` is not a good place to go, in terms of `execution_role_arn`.

I'm just hoping to raise discussion and caution on this public facing interface change as it's not revertible.
And I do have a concern that if EMR side keeps adding new args, our EMRStep interface will keep expanding and eventually it may be hard to use, see example of the obsoleted `RegisterModel` interface which contains ~30 args: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/workflow/step_collections.py#L57-L94

src/sagemaker/workflow/emr_step.py
qidewenwhen: I feel it'd be better to include
- The context - what EMR runtime role is used for, i.e. to add step with the existing cluster. 
   - at least we should include this in the sdk doc string as we don't expect customers to carefully explore all related docs, e.g. API docs
- What's wrong - i.e. EMR runtime role should not be set if `cluster_config` is not None. 

Given this context, that's why I feel it'd be better to group `execution_role_arn` validation with `cluster_id`. But as long as we put enough info in the doc string or error message to educate users and future developers, either way is fine.

src/sagemaker/workflow/emr_step.py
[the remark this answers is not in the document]
qidewenwhen: So **kwargs seems to be able to resolve my concern above but usually it may not be a good practice to include **kwargs.
Thus I'd suggest to have thorough consideration on this step interface and have more eyes on it.

src/sagemaker/workflow/emr_step.py
```diff
@@ -198,9 +208,16 @@ def __init__(
         if cluster_id is not None and cluster_config is not None:
             raise ValueError(ERR_STR_WITH_BOTH_CLUSTER_ID_AND_CLUSTER_CFG.format(step_name=name))
 
+        if execution_role_arn is not None and cluster_config is not None:
+            raise ValueError(ERR_STR_WITH_BOTH_EXEC_ROLE_ARN_AND_CLUSTER_CFG.format(step_name=name))
```
nmadan, line 212: Is this something we need to validate? This seems like a requirement from EMR side right? Should we start the EMR job and let it do this validation instead?

thbrooks22: We do this validation in the DSL. I used the same pattern that we are enforcing for `cluster_id` and `cluster_config`: for these, we are checking both in the SDK and in the DSL that they are not both set.

src/sagemaker/workflow/emr_step.py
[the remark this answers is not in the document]
qidewenwhen: Synced up offline: there's no better approach to avoid the call out issue. We will introduce a `step_arg` parameter to group all EMRStep argument related parameters in the future when the step interface grows too big.

qidewenwhen (approved):

src/sagemaker/workflow/emr_step.py
qidewenwhen: Out of curious, do you know is there any reason we need to do this checking in both places? 
If the DSL validation already covers these and they happen during creating pipeline time, we don't need to repeat the same validation in SDK. 
However, I'm fine to follow the same pattern that we already have for cluster_id and cluster_config in this PR.

qidewenwhen (approved):

knikure (approved):

PR #3796, opened 2023-04-18
Feat/jumpstart model estimator classes

evakravi:
*Issue #, if available:*

*Description of changes:*
Implements `JumpStartModel`, `JumpStartEstimator`, and supporting utilities. 

`JumpStartModel` and `JumpStartEstimator` are child classes of `Model` and `Estimator`.  By just supplying a model id, these classes will provide defaults for necessary fields for launching a training job or inference hosting endpoint. All the kwarg arguments are over-writable by the user, so they can customize their resources. 

Example workflow:
```
from sagemaker.jumpstart.model import JumpStartModel

model_id = "huggingface-eqa-bert-base-cased"

my_model = JumpStartModel(model_id=model_id)

predictor = my_model.deploy()

question = "What is Southern California often abbreviated as?"

response = predictor.predict(question)

print(response)

>> "SoCal"
```

important files to review:
- `src/sagemaker/jumpstart/model.py`
- `src/sagemaker/jumpstart/estimator.py`
- `src/sagemaker/jumpstart/factory/model.py`
- `src/sagemaker/jumpstart/factory/estimator.py`

*Testing done:*

Unit tests and integration tests have been added. 

## Merge Checklist

#### General

#### Tests


src/sagemaker/jumpstart/artifacts.py
```diff
@@ -571,3 +655,421 @@ def _retrieve_default_training_metric_definitions(
...
+        scope=JumpStartScriptScope.INFERENCE,
+        region=region,
+        tolerate_vulnerable_model=tolerate_vulnerable_model,
+        tolerate_deprecated_model=tolerate_deprecated_model,
+    )
+
+    return model_specs.supports_prepacked_inference()
+
+
+def _retrieve_serializer_from_content_type(
```
whittech1, line 701: All of the below serialization methods should be public functions in their own module (with unit tests please)

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -571,3 +655,421 @@ def _retrieve_default_training_metric_definitions(
...
+        tolerate_deprecated_model (bool): True if deprecated versions of model
+            specifications should be tolerated (exception not raised). If False, raises
+            an exception if the version of the model is deprecated. (Default: False).
+
+    Returns:
+        str: the default content type to use for the model.
+    """
+
+    if region is None:
+        region = JUMPSTART_DEFAULT_REGION_NAME
```
whittech1, line 929: Similar comment as above about potential consolidation

src/sagemaker/jumpstart/enums.py
```diff
@@ -75,3 +75,65 @@ class JumpStartTag(str, Enum):
...
+
+    X_IMAGE = "application/x-image"
+    LIST_TEXT = "application/list-text"
+    X_TEXT = "application/x-text"
+    JSON = "application/json"
+    CSV = "text/csv"
+    WAV = "audio/wav"
+
+    @staticmethod
+    def from_suffixed_type(mime_type_with_suffix: str) -> "MIMEType":
```
whittech1, line 136: unit tests? Also, return value in quotes?

src/sagemaker/content_types.py
```diff
@@ -10,12 +10,98 @@
...
+
+    return artifacts._retrieve_default_content_type(
+        model_id,
+        model_version,
+        region,
+        tolerate_vulnerable_model,
+        tolerate_deprecated_model,
+    )
 
-deprecations.removed_warning("The sagemaker.content_types module")
```
whittech1, line 18: If we are un-deprecating, could someone from the SDK team signoff?

whittech1: Also given that the functions are calling jumpstart utilities, should we just move to the `jumpstart` namespace?

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
...
+
+import numpy as np
+from six import with_metaclass
+
+from sagemaker.utils import DeferredError
+
+try:
+    import pandas
+except ImportError as e:
+    pandas = DeferredError(e)
```
whittech1, line 31: Could we add a comment explaining the necessity of this deferred error please?

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
...
+class BaseDeserializer(abc.ABC):
+    """Abstract base class for creation of new deserializers.
+
+    Provides a skeleton for customization requiring the overriding of the method
+    deserialize and the class attribute ACCEPT.
+    """
+
+    @abc.abstractmethod
+    def deserialize(self, stream, content_type):
+        """Deserialize data received from an inference endpoint.
```
whittech1, line 43: nit: `from an inference endpoint` is more specific than the general applicability of this class

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
...
+        Args:
+            accept (union[str, tuple[str]]): The MIME type (or tuple of allowable MIME types) that
+                is expected from the inference endpoint (default: "*/*").
+        """
+        super(SimpleBaseDeserializer, self).__init__()
+        self.accept = accept
+
+    @property
+    def ACCEPT(self):
+        """The tuple of possible content types that are expected from the inference endpoint."""
```
whittech1, line 79: nit changed docstring from base class. Also, return typings?

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
...
+    def ACCEPT(self):
+        """The content types that are expected from the inference endpoint."""
+
+
+class SimpleBaseDeserializer(with_metaclass(abc.ABCMeta, BaseDeserializer)):
+    """Abstract base class for creation of new deserializers.
+
+    This class extends the API of :class:~`sagemaker.deserializers.BaseDeserializer` with more
+    user-friendly options for setting the ACCEPT content type header, in situations where it can be
+    provided at init and freely updated.
```
whittech1, line 64: `SimpleBaseDeserializer` does not really communicate intent, and the less inheritance depth we have the better. We may want to consider if this class is really necessary as I think it is fine for each concrete subclass of `BaseDeserializer` to set `accept`.

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
...
+
+        Returns:
+            list: A list of JSON serializable objects.
+        """
+        try:
+            body = stream.read().decode("utf-8")
+            lines = body.rstrip().split("\n")
+            return [json.loads(line) for line in lines]
+        finally:
+            stream.close()
```
whittech1, line 324: Could we add basic unit tests on all of these classes please?

src/sagemaker/base_serializers.py
```diff
@@ -0,0 +1,396 @@
...
+import numpy as np
+from pandas import DataFrame
+from six import with_metaclass
+
+from sagemaker.utils import DeferredError
+
+try:
+    import scipy.sparse
+except ImportError as e:
+    scipy = DeferredError(e)
```
whittech1, line 30: Add comment please

src/sagemaker/base_serializers.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def __init__(self, content_type="application/json"):
+        """Initialize a ``SimpleBaseSerializer`` instance.
+
+        Args:
+            content_type (str): The MIME type to signal to the inference endpoint when sending
+            request data (default: "application/json").
+        """
+        super(SimpleBaseSerializer, self).__init__()
+        if not isinstance(content_type, str):
```
whittech1, line 73: We should create a method to reuse this behavior rather than inheritance

src/sagemaker/base_serializers.py
```diff
@@ -0,0 +1,396 @@
...
+            raise ValueError(
+                "content_type must be a string specifying the MIME type of the data sent in "
```
whittech1, line 75: Technically, we just check if the content type is a string so the error msg returned is misleading

src/sagemaker/base_serializers.py
```diff
@@ -0,0 +1,396 @@
...
+            try:
+                with open(data, "rb") as data_file:
+                    data_file_info = data_file.read()
+                    return data_file_info
+            except Exception as e:
+                raise ValueError(f"Could not open/read file: {data}. {e}")
+        if isinstance(data, bytes):
+            return data
+
+        raise ValueError(f"Object of type {type(data)} is not Data serializable.")
```
whittech1, line 396: Unit tests on these classes please

src/sagemaker/deserializers.py
```diff
@@ -13,312 +13,101 @@
...
-            object: The data deserialized into an object.
-        """
-
-    @property
-    @abc.abstractmethod
-    def ACCEPT(self):
-        """The content types that are expected from the inference endpoint."""
-
-
-class SimpleBaseDeserializer(with_metaclass(abc.ABCMeta, BaseDeserializer)):
```
whittech1, line 59: Now realizing that what I reviewed above may have been already in the SDK. Some of the proposed changes I made there may no longer apply then.

src/sagemaker/deserializers.py
```diff
@@ -13,312 +13,101 @@
...
-            stream.close()
-
-
-class CSVDeserializer(SimpleBaseDeserializer):
-    """Deserialize a stream of bytes into a list of lists.
-
-    Consider using :class:~`sagemaker.deserializers.NumpyDeserializer` or
-    :class:~`sagemaker.deserializers.PandasDeserializer` instead, if you'd like to convert text/csv
-    responses directly into other data types.
+    if not jumpstart_utils.is_jumpstart_model_input(model_id, model_version):
```
whittech1, line 58: Similar namespace comment, I think it is cleaner (and good for JS branding) if we keep all the utilities we add in `jumpstart` namespace.

src/sagemaker/serializers.py
```diff
@@ -13,380 +13,92 @@
 """Implements methods for serializing data for an inference endpoint."""
 from __future__ import absolute_import
 
-import abc
```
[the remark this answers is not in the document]
evakravi: The changes to `serializers.py` and `deserializers.py` look intense, but it's really just moving the file contents to `base_serializers.py` and `base_deserializers.py`, to remove circular imports.

src/sagemaker/model.py
```diff
@@ -276,6 +278,8 @@ def __init__(
                 authentication if they are provided. If they are not provided,
                 the SageMaker Python SDK attempts to use either the CodeCommit
                 credential helper or local credential storage for authentication.
+            model_id (str): JumpStart model id to be used for creating a `Predictor`.
```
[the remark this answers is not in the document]
evakravi: Interesting idea, let me look into that

src/sagemaker/jumpstart/utils.py
```diff
@@ -420,3 +420,13 @@ def verify_model_region_and_return_specs(
         )
 
     return model_specs
+
+
+def update_dict_if_key_not_present(
+    dict_to_update: dict, key_to_add: Any, value_to_add: Any
+) -> dict:
+    """If a key is not present in the dict, add the new (key, value) pair, otherwise return dict."""
```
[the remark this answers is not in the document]
evakravi: the docstring may be written poorly, but the dict does get returned

src/sagemaker/jumpstart/artifacts.py
evakravi: we expose the public variants in another module that is unit tested.

src/sagemaker/content_types.py
evakravi: we want to keep the approach of using top-level utilities for getting jumpstart arts (model_uris, script_uris, content_types, etc.)

src/sagemaker/base_deserializers.py
evakravi: this is all code that was moved from `deserializers.py`

src/sagemaker/deserializers.py
```diff
@@ -14,10 +14,14 @@
 from __future__ import absolute_import
 
 
-from typing import Any, List, Optional
+from typing import List, Optional
 
+# base_deserializers was renamed from deserializers, so this import
+# is for backwards compatibility.
```
JGuinegagne, line 20:
```py
# base_deserializers was refactored from deserializers, this import ensures backward compatibility
```

src/sagemaker/jumpstart/estimator.py
```diff
@@ -39,128 +33,78 @@ class JumpStartEstimator(Estimator):
...
+        self.instance_type = estimator_init_kwargs.instance_type
+        self.instance_count = estimator_init_kwargs.instance_count
+        self.region = estimator_init_kwargs.region
 
-        if new_hyperparameters == {}:
-            new_hyperparameters = None
+        super(JumpStartEstimator, self).__init__(**estimator_init_kwargs.to_kwargs_dict())
 
-        self.kwargs_for_base_estimator_class["hyperparameters"] = new_hyperparameters
+    def fit(self, *largs, **kwargs) -> None:
```
JGuinegagne, line 71: please use all the arguments of [fit](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/estimator.py#L1128)

At least keep `inputs` and `wait` args, but preferably also `logs` and `job_name`

src/sagemaker/jumpstart/estimator.py
```diff
@@ -39,128 +33,78 @@ class JumpStartEstimator(Estimator):
...
-            ),
+    def deploy(
+        self,
+        image_uri: Optional[str] = None,
+        source_dir: Optional[str] = None,
+        entry_point: Optional[str] = None,
+        env: Optional[Dict[str, str]] = None,
+        predictor_cls: Optional[Predictor] = None,
+        initial_instance_count: Optional[int] = None,
+        instance_type: Optional[str] = None,
```
JGuinegagne, line 92: please also expose `wait`, `kms_key`, `tags`, `model_download_timeout`, `container_startup_health_check_timeout` and `volume_size`  from the [deploy](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/estimator.py#L112) method.

Also `image_uri`, `source_dir`, `entry_point`, `env` and `predictor_cls` are inconsistent, please remove them. `deploy` seems to pass `serializer` and `deserializer`, that is what we should accept.

src/sagemaker/jumpstart/model.py
```diff
@@ -36,109 +28,54 @@ class JumpStartModel(Model):
...
         image_uri: Optional[str] = None,
-        model_uri: Optional[str] = None,
-        script_uri: Optional[str] = None,
+        model_data: Optional[str] = None,
+        source_dir: Optional[str] = None,
         entry_point: Optional[str] = None,
         env: Optional[Dict[str, str]] = None,
-        predictor_cls: Predictor = "JumpStartPredictor",
+        predictor_cls: Optional[Predictor] = None,
         **kwargs,
```
JGuinegagne, line 41: except for `model_id` and `model_version`, all other arguments should be present in the parent [constructor](https://github.com/aws/sagemaker-python-sdk/blob/409546d47eaa642f9598d0bbf966ea302e87f638/src/sagemaker/model.py#L119). Please remove all arguments that are not present (eg `instance_type`)

Please also expose at a minimum the parameters we use:
- `role`
- `sagemaker_session`
- `predictor_cls`
- `vpc_config`
- `enable_network_isolation`
- `model_kms_key`
- `code_location`

question: should `model_id` be optional so that we can support just passing a `task` in the future?

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,338 @@
...
+    instance_count: Optional[int] = None,
+    region: Optional[str] = None,
+    image_uri: Optional[str] = None,
+    model_uri: Optional[str] = None,
+    source_dir: Optional[str] = None,
+    entry_point: Optional[str] = None,
+    hyperparameters: Optional[dict] = None,
+    metric_definitions: Optional[List[dict]] = None,
+    kwargs: Optional[dict] = None,
+) -> JumpStartEstimatorInitKwargs:
```
JGuinegagne, line 56: same comment as for `Model` constructor:
- ensure you are only arguments that exists in parent [constructor](https://github.com/aws/sagemaker-python-sdk/blob/409546d47eaa642f9598d0bbf966ea302e87f638/src/sagemaker/estimator.py#L2376)
- expose at a minimum
    - `role`
    - `volume_size`
    - `volume_kms_key`
    -  `output_path`
    - `output_kms_key`
    - `sagemaker_session`
    - `tags`
    - `subnets`
    - `security_group_ids`
    - `encrypt_inter_container_traffic`
    - `enable_network_isolation`

Please also consider making `model_id` optional so that we can extend to `task`

JGuinegagne (changes requested):

src/sagemaker/jumpstart/utils.py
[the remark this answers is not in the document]
whittech1: Yea I am saying that we probably do not need to return the dict at all b/c the method is named `update_dict`. This is minor point though so very much non-blocking.

src/sagemaker/jumpstart/artifacts.py
whittech1: At a minimum, I think `_retrieve_serializer_from_content_type` and `_retrieve_deserializer_from_accept_type` should have unit tests

src/sagemaker/predictor.py
```diff
@@ -182,10 +182,14 @@ def _create_request_args(
...
-            args["ContentType"] = self.content_type
+            args["ContentType"] = (
+                self.content_type
+                if isinstance(self.content_type, str)
+                else ", ".join(self.content_type)
+            )
 
         if "Accept" not in args:
-            args["Accept"] = ", ".join(self.accept)
+            args["Accept"] = self.accept if isinstance(self.accept, str) else ", ".join(self.accept)
```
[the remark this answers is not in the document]
whittech1: Could you elaborate please? It seems that we are implicitly changing the interface of `Predictor` (and there may be a way around it)

src/sagemaker/content_types.py
whittech1: What is the rationale? At a product level, I would think that we would prefer to keep all of our utilities namespaced under `jumpstart` to increase the brand recognition.

src/sagemaker/jumpstart/estimator.py
```diff
@@ -0,0 +1,254 @@
...
+
+class JumpStartEstimator(Estimator):
+    """JumpStartEstimator class.
+
+    This class sets defaults based on the model id and version.
+    """
+
+    def __init__(
+        self,
+        model_id: str,
```
whittech1, line 46: We could consider exposing another simpler "constructor" using `@classmethod`. Something like:

```
class JumpStartEstimator(Estimator):

    def simple_create(model_id: str, model_version: str, **kwargs):
          return cls(model_id=model_id, model_version=model_version, **kwargs)
```

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,506 @@
...
+from sagemaker.instance_group import InstanceGroup
+from sagemaker.jumpstart.artifacts import _retrieve_kwargs
+from sagemaker.jumpstart.constants import (
+    JUMPSTART_DEFAULT_REGION_NAME,
+    TRAINING_ENTRY_POINT_SCRIPT_NAME,
+)
+from sagemaker.jumpstart.enums import JumpStartScriptScope, KwargUseCase
+from sagemaker.jumpstart.factory import model
+from sagemaker.jumpstart.types import (
+    JumpStartEstimatorDeployKwargs,
```
whittech1, line 43: I realize this is annoying, but could we please rename from `*Kwargs` to `*Options` or `*Settings`? I believe this alternate naming is more inline with the fundamental meaning behind the data encapsulated in these classes. It also makes those classes a more natural place to add utility methods.

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,506 @@
...
+        code_location=code_location,
+        entry_point=entry_point,
+        dependencies=dependencies,
+        instance_groups=instance_groups,
+        training_repository_access_mode=training_repository_access_mode,
+        training_repository_credentials_provider_arn=training_repository_credentials_provider_arn,
+        image_uri=image_uri,
+    )
+
+    estimator_init_kwargs = _add_model_version_to_kwargs(estimator_init_kwargs)
```
whittech1, line 156: Couldn't we set all of these defaults in the implementation of `JumpStartEstimatorInitKwargs`? I realize that that class is located in `types.py` but it seems like we end up having a lot of boilerplate code in the factory b/c of all the fields/kwargs we have to deal with.

src/sagemaker/jumpstart/factory/model.py
```diff
@@ -0,0 +1,351 @@
...
+        enable_network_isolation=enable_network_isolation,
+        model_kms_key=model_kms_key,
+        image_config=image_config,
+        code_location=code_location,
+        container_log_level=container_log_level,
+        dependencies=dependencies,
+        git_config=git_config,
+    )
+
+    model_init_kwargs = _add_model_version_to_kwargs(kwargs=model_init_kwargs)
```
whittech1, line 331: Similar comment as with Estimator factory

src/sagemaker/jumpstart/factory/estimator.py
whittech1: In fact, are we even really getting value out of the factory utilities? They make the `JumpStartEstimator` and `JumpStartModel` class end up just being boilerplate. Couldn't we do all of this logic in the implementation of those classes?

In my opinion, if we wanted to factory then we would need to identify groups of kwargs/options that make sense together (i.e. all the instance type kwargs, all the security setting kwargs, etc.). We can do this as incremental refactoring on the `JumpStartModel` / `JumpStartEstimator` classes

whittech1 (changes requested):

src/sagemaker/content_types.py
JGuinegagne: For context Chris: there has been debate in the past about the JumpStart branding vs Built-in Algorithm. 

One of the issue is that we use the common, top-level utility [image_uris.retrieve](https://github.com/aws/sagemaker-python-sdk/blob/2bac010cffa615d624578be822db6e00c2826d4c/src/sagemaker/image_uris.py#L42).

This is the precedent we've been using for all other utilities such as [script](https://github.com/aws/sagemaker-python-sdk/blob/2bac010cffa615d624578be822db6e00c2826d4c/src/sagemaker/script_uris.py#L25), [artifacts](https://github.com/aws/sagemaker-python-sdk/blob/2bac010cffa615d624578be822db6e00c2826d4c/src/sagemaker/model_uris.py#L26) or [hyperparameters](https://github.com/aws/sagemaker-python-sdk/blob/2bac010cffa615d624578be822db6e00c2826d4c/src/sagemaker/hyperparameters.py#L28).

src/sagemaker/model.py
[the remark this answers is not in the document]
JGuinegagne: agree with Chris here, remove please

src/sagemaker/model.py
```diff
@@ -611,6 +634,13 @@ def _upload_code(self, key_prefix: str, repack: bool = False) -> None:
                     s3_prefix=repacked_model_data, script_name=os.path.basename(self.entry_point)
                 )
 
+            LOGGER.info(  # pylint: disable=W1203
+                f"Repacking model artifact ({self.model_data}), script artifact "
+                f"({self.source_dir}), and dependencies ({self.dependencies}) "
+                f"into single tar.gz file located at {repacked_model_data}. "
+                "This may take some time depending on model size..."
+            )
```
JGuinegagne, line 642: please provide a way to bypass this log in the `SessionSettings`.

src/sagemaker/model.py
```diff
@@ -1307,7 +1337,19 @@ def deploy(
         )
 
         if self.predictor_cls:
-            predictor = self.predictor_cls(self.endpoint_name, self.sagemaker_session)
+            if self.predictor_cls == JumpStartPredictor:
```
JGuinegagne, line 1340: remove please: the JumpStart-specific logic shouldn't leak into the base class. Essentially, users shouldn't pass a `JumpStartPredictor` to the parent class directly.

src/sagemaker/predictor.py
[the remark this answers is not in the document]
JGuinegagne: +1 with Chris' question, I don't understand your response Evan.

tests/unit/sagemaker/jumpstart/utils.py
```diff
@@ -165,3 +165,24 @@ def patched_retrieval_function(
...
+    for key, value in dictionary_with_overwrites.items():
+
+        if key in base_dictionary:
+            base_dictionary_entry = base_dictionary[key]
+            if isinstance(base_dictionary_entry, list):
+                assert isinstance(value, list)
+                value += base_dictionary_entry
+            if isinstance(base_dictionary_entry, dict):
+                assert isinstance(value, dict)
+                value.update(base_dictionary_entry)
```
JGuinegagne, line 184: question: shouldn't this logic be recursive?

JGuinegagne (changes requested):
Please don't leak JumpStart-specific logic into `Model` class.

src/sagemaker/jumpstart/constants.py
```diff
@@ -132,3 +145,29 @@
...
+SERIALIZER_TYPE_TO_CLASS_MAP: Dict[SerializerType, BaseSerializer] = {
+    SerializerType.RAW_BYTES: DataSerializer(),
+    SerializerType.JSON: JSONSerializer(),
+    SerializerType.TEXT: IdentitySerializer(),
+    SerializerType.CSV: CSVSerializer(),
+}
+
+DESERIALIZER_TYPE_TO_CLASS_MAP: Dict[DeserializerType, BaseDeserializer] = {
+    DeserializerType.JSON: JSONDeserializer(),
+}
```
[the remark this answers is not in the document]
evakravi: Great callout, this would've caused bad memory problems. I'll change this

src/sagemaker/environment_variables.py
```diff
@@ -15,20 +15,22 @@
...
 def retrieve_default(
-    region=None,
-    model_id=None,
-    model_version=None,
+    region: Optional[str] = None,
+    model_id: Optional[str] = None,
+    model_version: Optional[str] = None,
     tolerate_vulnerable_model: bool = False,
     tolerate_deprecated_model: bool = False,
+    use_case: EnvVariableUseCase = EnvVariableUseCase.AWS_SDK,
```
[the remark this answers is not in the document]
evakravi: Not sure about this, let's discuss later, cause I agree it needs improvement

src/sagemaker/hyperparameters.py
```diff
@@ -26,10 +26,10 @@
...
 
 def retrieve_default(
-    region=None,
-    model_id=None,
-    model_version=None,
-    include_container_hyperparameters=False,
+    region: Optional[str] = None,
+    model_id: Optional[str] = None,
+    model_version: Optional[str] = None,
+    include_container_hyperparameters: bool = False,
```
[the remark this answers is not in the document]
evakravi: yes

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -386,6 +396,10 @@ def _retrieve_default_environment_variables(
         tolerate_deprecated_model (bool): True if deprecated versions of model
             specifications should be tolerated (exception not raised). If False, raises
             an exception if the version of the model is deprecated. (Default: False).
+        use_case (EnvVariableUseCase): The use case for the environment variables. The
```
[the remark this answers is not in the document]
evakravi: i think we may have env vars for training jobs or other use cases, so we wanted an enum

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -405,10 +419,80 @@ def _retrieve_default_environment_variables(
...
+
+    if region is None:
+        region = JUMPSTART_DEFAULT_REGION_NAME
+
+    if use_case in {KwargUseCase.MODEL, KwargUseCase.MODEL_DEPLOY}:
+        scope = JumpStartScriptScope.INFERENCE
+    elif use_case in {KwargUseCase.ESTIMATOR, KwargUseCase.ESTIMATOR_FIT}:
+        scope = JumpStartScriptScope.TRAINING
+    else:
+        raise ValueError(f"Unsupported kwarg use case: {use_case}")
```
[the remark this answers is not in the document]
evakravi: haha you really are allergic to kwargs 😂

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -571,3 +655,422 @@ def _retrieve_default_training_metric_definitions(
...
+) -> List[BaseDeserializer]:
+    """Retrieves the supported deserializers for the model.
+
+    Args:
+        model_id (str): JumpStart model ID of the JumpStart model for which to
+            retrieve the supported deserializers.
+        model_version (str): Version of the JumpStart model for which to retrieve the
+            supported deserializers.
+        region (Optional[str]): Region for which to retrieve default training metric
+            definitions.
```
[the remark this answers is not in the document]
evakravi: good callout

src/sagemaker/jumpstart/estimator.py
```diff
@@ -0,0 +1,278 @@
...
+        self.model_version = estimator_init_kwargs.model_version
+        self.instance_type = estimator_init_kwargs.instance_type
+        self.tolerate_deprecated_model = estimator_init_kwargs.tolerate_deprecated_model
+        self.tolerate_vulnerable_model = estimator_init_kwargs.tolerate_vulnerable_model
+        self.instance_count = estimator_init_kwargs.instance_count
+        self.region = estimator_init_kwargs.region
+
+        super(JumpStartEstimator, self).__init__(**estimator_init_kwargs.to_kwargs_dict())
+
+    def fit(
```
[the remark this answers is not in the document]
evakravi: i'd rather do that later since im unfamiliar with this

src/sagemaker/jumpstart/estimator.py
```diff
@@ -0,0 +1,278 @@
...
+            experiment_config=experiment_config,
+            tolerate_vulnerable_model=self.tolerate_vulnerable_model,
+            tolerate_deprecated_model=self.tolerate_deprecated_model,
+        )
+
+        logger.info(  # pylint: disable=W1203
+            f"Creating SageMaker Training Job for {estimator_fit_kwargs.model_id}. "
+            f"Provisioning {self.instance_type} instance in "
+            f"{estimator_fit_kwargs.region}."
+        )
```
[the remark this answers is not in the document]
evakravi: you can change the log level to ignore info logs.

src/sagemaker/jumpstart/factory/estimator.py
evakravi: I agree that we may want to rename the type, let me think about this

tests/unit/sagemaker/jumpstart/utils.py
evakravi: for the tests we run, this is good enough

JGuinegagne: Ah it's a test util, nevermind then!

src/sagemaker/model.py
JGuinegagne: Please review this message with SDK team.

tests/unit/sagemaker/jumpstart/test_model.py
```diff
@@ -0,0 +1,277 @@
...
+from sagemaker.predictor import Predictor
+
+from tests.unit.sagemaker.jumpstart.utils import get_special_model_spec, overwrite_dictionary
+
+
+class ModelTest(unittest.TestCase):
+
+    execution_role = "fake role! do not use!"
+    region = "us-west-2"
+
```
JGuinegagne, line 29: could you please add test cases to ensure that:
- that the image is retrieved with `INFERENCE` scope
- if the user does not pass an instance type, the default instance is retrieved before retrieving the image
- if the user passes an instance type, the image is retrieved with that instance
- that intelligent defaults for `vpc`, each `kms` key, `role` and `network_isolation` are correctly handled

If any of these tests are not possible at the top level, make sure you add coverage for them in the kwargs utils.

tests/unit/sagemaker/jumpstart/test_predictor.py
```diff
@@ -0,0 +1,37 @@
...
+from sagemaker.jumpstart.utils import verify_model_region_and_return_specs
+from sagemaker.serializers import IdentitySerializer
+from tests.unit.sagemaker.jumpstart.utils import (
+    get_special_model_spec,
+)
+
+
+@patch("sagemaker.jumpstart.artifacts.verify_model_region_and_return_specs")
+@patch("sagemaker.jumpstart.accessors.JumpStartModelsAccessor.get_model_specs")
+def test_list_jumpstart_scripts(
```
JGuinegagne, line 19: rename test case please

tests/unit/sagemaker/serializers/jumpstart/test_serializers.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+from mock.mock import patch
+
+from sagemaker import base_serializers, serializers
+from sagemaker.jumpstart.utils import verify_model_region_and_return_specs
+
+from tests.unit.sagemaker.jumpstart.utils import get_special_model_spec
+
+
```
JGuinegagne, line 23: please unit test both methods: `retrieve_options` and `retrieve_default`

same for `deserializers`

tests/unit/test_predictor.py
```diff
@@ -249,7 +249,7 @@ def test_predict_call_with_multiple_accept_types():
     assert kwargs == expected_request_args
```
JGuinegagne, line 250: please unit test `retrieve_default` utility

tests/unit/sagemaker/jumpstart/test_estimator.py
```diff
@@ -0,0 +1,463 @@
...
+
+from sagemaker.jumpstart.utils import get_jumpstart_content_bucket
+from tests.integ.sagemaker.jumpstart.utils import get_training_dataset_for_model_and_version
+from sagemaker.predictor import Predictor
+from sagemaker.session import Session
+from tests.unit.sagemaker.jumpstart.utils import (
+    get_special_model_spec,
+    overwrite_dictionary,
+)
+
```
JGuinegagne, line 33: Could you please add test cases to ensure that:
- that the training image is retrieved with `TRAINING` scope
- that the hosting image is retrieved with `INFERENCE` scope on `deploy`
- if the user does not pass a training instance type, the default training instance is retrieved before retrieving the training image
- if the user passes a training instance type, the training image is retrieved with that instance
- that intelligent defaults for `vpc`, each `kms` key, `role` and `network_isolation` are correctly handled
- if the user does not pass a hosting instance type on `deploy`, the default hosting instance is retrieved before retrieving the hosting image
- if the user passes a hosting instance type, the hosting image is retrieved with that instance

If any of these tests are not possible at the top level, make sure you add coverage for them in the kwargs utils.

JGuinegagne (changes requested):
Thanks for making the changes to the interface:
- let's align on how we handle intelligent defaults
- let's align on `use_case` argument
- expand unit test coverage, particularly to factory
- remove logs for instance count default
- add generic unit test that ensures that all arguments for `JumpStartModel` and `JumpStartEstimator` constructors, `deploy` and `fit` methods are passed as is to the parent class method when provided
- add generic unit test that ensures that all arguments for `Model` and `Estimator` constructors, `deploy` and `fit` methods are also arguments of `JumpStartModel` and `JumpStartEstimator`

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,570 @@
...
+
+    kwargs.model_version = kwargs.model_version or "*"
+
+    return kwargs
+
+
+def _add_role_to_kwargs(kwargs: JumpStartEstimatorInitKwargs) -> JumpStartEstimatorInitKwargs:
+    """Sets role based on default or override, returns full kwargs."""
+
+    kwargs.role = kwargs.role or get_execution_role()
```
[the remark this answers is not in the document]
evakravi: yes, good suggestion

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,570 @@
...
+            tolerate_vulnerable_model=kwargs.tolerate_vulnerable_model,
+        )
+        or []
+    )
+
+    for metric_definition in default_metric_definitions:
+        if metric_definition["Name"] not in {
+            definition["Name"] for definition in kwargs.metric_definitions
+        }:
+            kwargs.metric_definitions.append(metric_definition)
```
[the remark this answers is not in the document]
evakravi: AFAIK, adding an additional metric doesn't interfere with metrics which already exist

src/sagemaker/jumpstart/factory/estimator.py
```diff
@@ -0,0 +1,570 @@
...
+        model_version=kwargs.model_version,
+        region=kwargs.region,
+        use_case=KwargUseCase.ESTIMATOR,
+        tolerate_deprecated_model=kwargs.tolerate_deprecated_model,
+        tolerate_vulnerable_model=kwargs.tolerate_vulnerable_model,
+    )
+
+    for key, value in estimator_kwargs_to_add.items():
+        if getattr(kwargs, key) is None:
+            setattr(kwargs, key, value)
```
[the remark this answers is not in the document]
evakravi: no

src/sagemaker/jumpstart/factory/model.py
```diff
@@ -0,0 +1,460 @@
...
+    tolerate_deprecated_model: bool,
+) -> Predictor:
+    """Converts predictor returned from ``Model.deploy()`` into a JumpStart-specific one.
+
+    If the predictor is a custom predictor class, honor the user setting and do not mutate.
+    If the predictor is a base-class predictor, set model-specific defaults.
+    """
+
+    # if there's a non-default predictor, do not mutate -- return as is
+    if not isinstance(predictor, Predictor):
```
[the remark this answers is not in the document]
evakravi: interesting. i thought we would need `issubclass`

src/sagemaker/jumpstart/types.py
```diff
@@ -451,3 +511,550 @@ def __init__(
...
+    """Data class for JumpStart object kwargs."""
+
+    SERIALIZATION_EXCLUSION_SET = {}
+
+    def to_kwargs_dict(self):
+        """Serializes object to dictionary to be used for kwargs for method arguments."""
+        kwargs_dict = {}
+        for field in self.__slots__:
+            if field not in self.SERIALIZATION_EXCLUSION_SET:
+                att_value = getattr(self, field)
```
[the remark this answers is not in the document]
evakravi: that would basically be a runtime error, since the types are defined incorrectly. so that should never happen

tests/unit/sagemaker/serializers/jumpstart/test_serializers.py
evakravi: don't i currently do that? maybe i should break into multiple test cases for clarity

src/sagemaker/jumpstart/constants.py
```diff
@@ -132,3 +145,29 @@
...
+    MIMEType.CSV: SerializerType.CSV,
+    MIMEType.WAV: SerializerType.RAW_BYTES,
+}
+
+
+ACCEPT_TYPE_TO_DESERIALIZER_TYPE_MAP: Dict[MIMEType, DeserializerType] = {
+    MIMEType.JSON: DeserializerType.JSON,
+}
+
+SERIALIZER_TYPE_TO_CLASS_MAP: Dict[SerializerType, Callable[[], BaseSerializer]] = {
```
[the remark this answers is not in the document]
evakravi: technically, it's still a callable though, since it's invoked with the `__call__` method. But i'll change the typing since it's unnecessarily complicated.

src/sagemaker/jumpstart/factory/model.py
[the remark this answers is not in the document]
evakravi: ok, we need to use `type()` not `isinstance`. Just tested that out. Thanks for calling this out, it's a good python finding

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -405,10 +419,80 @@ def _retrieve_default_environment_variables(
...
+        return model_specs.model_kwargs
+
+    if use_case == KwargUseCase.MODEL_DEPLOY:
+        return model_specs.deploy_kwargs
+
+    if use_case == KwargUseCase.ESTIMATOR:
+        return model_specs.estimator_kwargs
+
+    if use_case == KwargUseCase.ESTIMATOR_FIT:
+        return model_specs.fit_kwargs
```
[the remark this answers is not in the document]
evakravi: This utility is private. I made `KwargUseCase` -> `_KwargUseCase`, so it's also denoted as private. The implementation and testing is easier IMO with the enum and single function.

src/sagemaker/jumpstart/artifacts.py
```diff
@@ -571,3 +655,421 @@ def _retrieve_default_training_metric_definitions(
...
+    model_specs = verify_model_region_and_return_specs(
+        model_id=model_id,
+        version=model_version,
+        scope=JumpStartScriptScope.INFERENCE,
+        region=region,
+        tolerate_vulnerable_model=tolerate_vulnerable_model,
+        tolerate_deprecated_model=tolerate_deprecated_model,
+    )
+
+    return model_specs.supports_prepacked_inference()
```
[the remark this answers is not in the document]
evakravi: In the pattern of the other utilities, we want to create a wrapper in `artifacts.py`. It would be an anti-pattern to call the low level utility, IMO.

src/sagemaker/accept_types.py
```diff
@@ -0,0 +1,103 @@
...
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated
+            (exception not raised). False if these models should raise an exception.
+            (Default: False).
+    Returns:
+        list: The supported accept types to use for the model.
+
+    Raises:
+        ValueError: If the combination of arguments specified is not supported.
+    """
+    if not jumpstart_utils.is_jumpstart_model_input(model_id, model_version):
```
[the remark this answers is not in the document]
evakravi: this is a standard line we put in our `retrieve` functions. The idea is that these functions could theoretically be used for non-JumpStart purposes.

src/sagemaker/predictor.py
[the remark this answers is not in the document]
evakravi: Basically, someone can make the accept type field a `str` and `list` type, both work, even though the typings indicate only `str` work. However, they both must be serialized to string type.

src/sagemaker/predictor.py
```diff
@@ -182,10 +182,14 @@ def _create_request_args(
             args["EndpointName"] = self.endpoint_name
 
         if "ContentType" not in args:
-            args["ContentType"] = self.content_type
+            args["ContentType"] = (
+                self.content_type
+                if isinstance(self.content_type, str)
```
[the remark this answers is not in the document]
evakravi: See comment for `Accept`. The typings are incorrect. It can be a string or array.

src/sagemaker/jumpstart/estimator.py
```diff
@@ -0,0 +1,947 @@
...
+logger = logging.getLogger(__name__)
+
+
+class JumpStartEstimator(Estimator):
+    """JumpStartEstimator class.
+
+    This class sets defaults based on the model ID and version.
+    """
+
+    def __init__(
```
akozd, line 49: This `__init__` is insane, I don't think I've ever seen this many params

JGuinegagne: Isn't it? We are trying to provide intellisense for developers by exposing all the parameters of the parent class [Estimator](https://github.com/aws/sagemaker-python-sdk/blob/68885dcba8975f06f1e041dc7974c939e231e573/src/sagemaker/estimator.py#L2376). This forces us to expose all the parameters of the parent class constructor.

If we do not, and provide a catch all kwargs for some of these params, we would deteriorate the devX.

src/sagemaker/jumpstart/artifacts/predictors.py
```diff
@@ -0,0 +1,422 @@
...
+    return _retrieve_deserializer_from_accept_type(MIMEType.from_suffixed_type(default_accept_type))
+
+
+def _retrieve_default_serializer(
+    model_id: str,
+    model_version: str,
+    region: Optional[str],
+    tolerate_vulnerable_model: bool = False,
+    tolerate_deprecated_model: bool = False,
+) -> BaseSerializer:
```
JGuinegagne, line 114: sanity-check: Does it return an (a) instance of a child class of `BaseSerializer`, or (b) the child class itself?

I _think_ it is (a), in which case your typing is correct, just want to double check.

src/sagemaker/jumpstart/artifacts/environment_variables.py
```diff
@@ -0,0 +1,80 @@
...
+        region=region,
+        tolerate_vulnerable_model=tolerate_vulnerable_model,
+        tolerate_deprecated_model=tolerate_deprecated_model,
+    )
+
+    default_environment_variables: Dict[str, str] = {}
+    for environment_variable in model_specs.inference_environment_variables:
+        if include_aws_sdk_env_vars or (
+            not include_aws_sdk_env_vars and environment_variable.required_for_model_class is True
+        ):
```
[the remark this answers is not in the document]
evakravi: Lol yup. I think I replaced the enum with a boolean, forgot to simplify the logic

src/sagemaker/jumpstart/artifacts/predictors.py
evakravi: It is (a)

tests/unit/sagemaker/deserializers/jumpstart/test_deserializers.py
```diff
@@ -0,0 +1,65 @@
...
+
+from sagemaker import base_deserializers, deserializers
+from sagemaker.jumpstart.utils import verify_model_region_and_return_specs
+
+from tests.unit.sagemaker.jumpstart.utils import get_special_model_spec
+
+
+@patch("sagemaker.jumpstart.artifacts.predictors.verify_model_region_and_return_specs")
+@patch("sagemaker.jumpstart.accessors.JumpStartModelsAccessor.get_model_specs")
+def test_jumpstart_deserializers(
```
JGuinegagne, line 26: could you break into two test cases please?

tests/unit/sagemaker/environment_variables/jumpstart/test_default.py
```diff
@@ -90,3 +92,76 @@ def test_jumpstart_default_environment_variables(patched_get_model_specs):
         environment_variables.retrieve_default(
             model_id=model_id,
         )
+
+
+@patch("sagemaker.jumpstart.accessors.JumpStartModelsAccessor.get_model_specs")
+def test_jumpstart_sdk_environment_variables(patched_get_model_specs):
```
JGuinegagne, line 98: break into different test cases please

src/sagemaker/jumpstart/cache.py
```diff
@@ -218,8 +219,7 @@ def _get_manifest_key_from_model_id_semantic_version(
 
         error_msg = f"Unable to find model manifest for '{model_id}' with version '{version}'. "
         error_msg += (
-            "Visit https://sagemaker.readthedocs.io/en/stable/doc_utils/pretrainedmodels.html"
-            " for updated list of models. "
+            f"Visit {MODEL_ID_LIST_WEB_URL} for updated list of models. "
```
JGuinegagne, line 222: Could you add a line explaining that customers should upgrade to the latest version of the SDK to access all the models?

src/sagemaker/jumpstart/notebook_utils.py
```diff
@@ -419,3 +419,16 @@ def get_model_url(
         region=region, model_id=model_id, version=model_version
     )
     return model_specs.url
+
+
+def is_valid_model_id(
+    model_id: str, script: JumpStartScriptScope = JumpStartScriptScope.INFERENCE
+) -> bool:
+    """Returns True if the model id is supported for the given script."""
```
JGuinegagne, line 427: nit^2: `Return True if the model ID [...]`

src/sagemaker/jumpstart/notebook_utils.py
```diff
@@ -419,3 +419,16 @@ def get_model_url(
...
+    if model_id in {None, ""}:
+        return False
+    if not isinstance(model_id, str):
```
JGuinegagne, line 430: nit: weird for a primitive, `if not type(model_id) == str`

tests/unit/sagemaker/jumpstart/test_utils.py
```diff
@@ -906,3 +914,16 @@ def test_mime_type_enum_from_str():
         for suffix in suffixes:
             mime_type_with_suffix = mime_type + suffix
             assert MIMEType.from_suffixed_type(mime_type_with_suffix) == mime_type
+
+
+def test_stringify_object():
+    class MyTestClass:
+        def __init__(self):
+            self.blah = "blah"
+            self.wtafigo = "eiifccreeeiuclkftdvttufbkhirtvvbhrieclghjiru"
```
JGuinegagne, line 923: although the string is long, it's functionally the same test as line 922.

Could you vary the value type, for example could you try `int`, `float`, `List[str]`, `Dict[str, str]` and `List[Dict[str, str]]` ?

src/sagemaker/jumpstart/notebook_utils.py
evakravi: Agh, i missed this one

src/sagemaker/jumpstart/notebook_utils.py
JGuinegagne: nevermind then!

src/sagemaker/base_deserializers.py
```diff
@@ -0,0 +1,324 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mufaddal-rohawala, line 1: When moving the contents of modules around should be a seperate PR, it's very hard to check on what code is new and what already existed.

src/sagemaker/base_predictor.py
```diff
@@ -0,0 +1,576 @@
...
+    def content_type(self) -> str:
+        """The MIME type of the data sent to the inference server."""
+
+    @property
+    @abc.abstractmethod
+    def accept(self) -> Tuple[str]:
+        """The content type(s) that are expected from the inference server."""
+
+
+class Predictor(PredictorBase):
```
mufaddal-rohawala, line 79: This movement of code changes accross modules are backwards incompatible, and might code like - https://github.com/aws/amazon-sagemaker-examples/blob/692c23949ddddc8abdeca7e58ddde85535f35ffd/advanced_functionality/autogluon-sagemaker-pipeline/pipelines/ag_model.py#L7
I can't allow this.

mufaddal-rohawala (changes requested):

src/sagemaker/base_predictor.py
evakravi: Hey, on the surface it looks backwards incompatible. But if you look at the top of predictors.py, we import everything from base_predictors.py. You can see that the unit/integ tests pass which means this is backwards compatible.

src/sagemaker/predictor.py
```diff
@@ -13,548 +13,67 @@
...
+from sagemaker.jumpstart.factory.model import get_default_predictor
+from sagemaker.jumpstart.utils import is_jumpstart_model_input
+
+from sagemaker.session import Session
+
+from sagemaker.base_predictor import Predictor
+
+# base_predictor was refactored from predictor.
+# this import ensures backward compatibility.
+from sagemaker.base_predictor import *  # noqa: F403, F401 # pylint: disable=W0614,W0401
```
mufaddal-rohawala, line 27: Though this ensures backwards compatibility, it would be hard for us/users to trace the specific objects/functions, thats why `import *` is usually not recommended, is it possible to add import specifying each objects/functions to import. Or any other alternative?

tests/unit/test_predictor.py
mufaddal-rohawala: +1

tests/integ/sagemaker/jumpstart/estimator/test_jumpstart_estimator.py
```diff
@@ -0,0 +1,77 @@
...
+    model_id, model_version = "huggingface-spc-bert-base-cased", "*"
+
+    estimator = JumpStartEstimator(
+        model_id=model_id,
+        role=get_sm_session().get_caller_identity_arn(),
+        sagemaker_session=get_sm_session(),
+        tags=[{"Key": JUMPSTART_TAG, "Value": os.environ[ENV_VAR_JUMPSTART_SDK_TEST_SUITE_ID]}],
+    )
+
+    estimator.fit(
```
mufaddal-rohawala, line 45: which instance will this use for training and inference below? concern is do we need limit increases for CI accounts.

src/sagemaker/accept_types.py
```diff
@@ -0,0 +1,103 @@
...
+        raise ValueError(
+            "Must specify `model_id` and `model_version` when retrieving accept types."
```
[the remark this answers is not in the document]
evakravi: That's a good point. Let me update it

src/sagemaker/predictor.py
evakravi: let me importing each individually, see if that breaks anything

tests/integ/sagemaker/jumpstart/estimator/test_jumpstart_estimator.py
evakravi: Good question, this one is ml.p3.2xlarge

src/sagemaker/jumpstart/estimator.py
```diff
@@ -897,6 +911,16 @@ def deploy(
                 (Default: None).
         """
 
+        self.orig_predictor_cls = predictor_cls
+
+        sagemaker_session = sagemaker_session or self.sagemaker_session
+        role = resolve_model_intelligent_default_field(
+            field_name="role",
+            field_val=role,
```
JGuinegagne, line 919: question: what is `role` here? I don't see this variable initialized...

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -80,15 +81,17 @@ def __init__(
                 self.message = (
                     f"Version '{version}' of JumpStart model '{model_id}' "  # type: ignore
                     "has at least 1 vulnerable dependency in the inference script. "
-                    "Please try targetting a higher version of the model. "
-                    f"List of vulnerabilities: {', '.join(vulnerabilities)}"  # type: ignore
+                    "Please try targetting a higher version of the model or using a "
```
JGuinegagne, line 84: question: shouldn't this read `targeting` ?

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -80,15 +81,17 @@ def __init__(
...
+                    "different model. List of vulnerabilities: "
+                    f"{', '.join(vulnerabilities)}"  # type: ignore
                 )
             elif scope == JumpStartScriptScope.TRAINING:
                 self.message = (
                     f"Version '{version}' of JumpStart model '{model_id}' "  # type: ignore
                     "has at least 1 vulnerable dependency in the training script. "
-                    "Please try targetting a higher version of the model. "
-                    f"List of vulnerabilities: {', '.join(vulnerabilities)}"  # type: ignore
+                    "Please try targetting a higher version of the model or using a "
```
JGuinegagne, line 92: same question here.

src/sagemaker/jumpstart/exceptions.py
```diff
@@ -122,7 +125,8 @@ def __init__(
                 raise RuntimeError("Must specify `model_id` and `version` arguments.")
             self.message = (
                 f"Version '{version}' of JumpStart model '{model_id}' is deprecated. "
-                "Please try targetting a higher version of the model."
+                "Please try targetting a higher version of the model or using a "
```
JGuinegagne, line 128: and here.

src/sagemaker/jumpstart/utils.py
```diff
@@ -523,5 +532,36 @@ def resolve_estimator_intelligent_default_field(
...
+    return f"{type(obj).__name__}: {str(non_none_atts)}"
+
+
+def is_valid_model_id(
+    model_id: Optional[str],
+    region: Optional[str] = None,
+    model_version: Optional[str] = None,
+    script: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+) -> bool:
+    """Returns True if the model ID is supported for the given script."""
```
JGuinegagne, line 545: could you add a `Raises` section please?

tests/unit/sagemaker/jumpstart/estimator/test_estimator.py
```diff
@@ -447,7 +475,9 @@ def evaluate_estimator_workflow_with_kwargs(
             f"{get_training_dataset_for_model_and_version(model_id, model_version)}",
         }
 
-        expected_fit_kwargs = overwrite_dictionary({"inputs": channels, "wait": True}, fit_kwargs)
+        expected_fit_kwargs = overwrite_dictionary(
+            {"inputs": channels, "wait": True, "job_name": "none of your business"}, fit_kwargs
```
JGuinegagne, line 479: lol, except that I don't think you can have spaces in resource names.

Does that mean the validator is not catching that the name isn't valid?

evakravi: the validator is in the base class, and we mock the call to the base class, thus why the validator is not working.

evakravi: Anyways, i'll remove the spaces so to not raise any eyebrows

JGuinegagne (approved):

src/sagemaker/jumpstart/estimator.py
JGuinegagne: Got it, that's an arg, thanks for confirming!

src/sagemaker/jumpstart/artifacts/predictors.py
```diff
@@ -0,0 +1,422 @@
...
+    deserializers_with_duplicates: List[BaseDeserializer] = [
+        _retrieve_deserializer_from_accept_type(MIMEType.from_suffixed_type(accept_type))
+        for accept_type in supported_accept_types
+    ]
+
+    deserializers: List[BaseDeserializer] = []
+
+    for deserializer in deserializers_with_duplicates:
+        if type(deserializer) not in seen_classes:
+            seen_classes.add(type(deserializer))
```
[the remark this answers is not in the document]
evakravi: it's not a duplicate, the set keeps track of classes, the list keeps track of instances of the classes

src/sagemaker/jumpstart/model.py
```diff
@@ -0,0 +1,154 @@
...
+        kms_key: Optional[str] = None,
+        wait: Optional[bool] = None,
+        data_capture_config: Optional[DataCaptureConfig] = None,
+        async_inference_config: Optional[AsyncInferenceConfig] = None,
+        serverless_inference_config: Optional[ServerlessInferenceConfig] = None,
+        volume_size: Optional[int] = None,
+        model_data_download_timeout: Optional[int] = None,
+        container_startup_health_check_timeout: Optional[int] = None,
+        inference_recommendation_id: Optional[str] = None,
+        explainer_config: Optional[ExplainerConfig] = None,
```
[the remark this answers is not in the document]
evakravi: no kwargs. this was making the implementation too complicated. no functionality has been lost

mufaddal-rohawala (commented):
/bot run pr

mufaddal-rohawala (commented):
/bot run slow-tests

mufaddal-rohawala (approved):

PR #3821, opened 2023-04-27
fix: double Run create on load_run

### LINKED ISSUE #3673: list_runs doesn't work twice.  Raises a ValueError about the name length of the run.
**Describe the bug**
I have a dummy Sagemaker notebook where I use an experiment name `BO-POC-synthetic-data`.
When I `list_runs` for that experiment once, it works. 
When I do `list_runs` for that experiment a second time, it makes 
> `ValueError: The run_name (length: 73) must have length less than or equal to 59`

**To reproduce**
```
from smexperiments.experiment import Experiment
from sagemaker.experiments.run import Run, load_run, list_runs

EXPERIMENT_NAME = "BO-POC-synthetic-data"

with Run(experiment_name=EXPERIMENT_NAME) as run:
    
    run.log_parameter("my_param", "AP") 
    run.log_metric("my_metric", 0.8)

    print(f"experiment name: {run.experiment_name}")
    print(run.experiment_config)

runs = list_runs(EXPERIMENT_NAME)      # <---- WORKS ONCE
runs = list_runs(EXPERIMENT_NAME)      # <---- FAILS SECOND TIME
![Screenshot 2023-02-21 at 2 00 35 PM](https://user-images.githubusercontent.com/406350/220468454-83f2533d-f0fb-48c6-86af-ab555b0f5f39.png)

```
Note that there are two trial components generated.  
Conceptually, I'd only expect one trial component; why is there one with the experiment name repeated twice? 
I suspect the longer one is causing the exception.
Screenshot 2023-02-21 at 2 00 35 PM

**Expected behavior**
I'd expect to be able to list-runs twice. 

**Screenshots or logs**
Screenshot 2023-02-21 at 1 57 48 PM

Also, a zipped version of my Studio notebook: 
[JM-2023-02-21_bug_report_Sagemaker_Experiments_bug_report.html.zip](https://github.com/aws/sagemaker-python-sdk/files/10798119/JM-2023-02-21_bug_report_Sagemaker_Experiments_bug_report.html.zip)

**System information**
- **SageMaker Python SDK version**: 2.133.0
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: None
- **Framework version**:
- **Python version**:  Python 3.9.11
- **CPU or GPU**: CPU (sagemaker Studio)
- **Custom Docker image (Y/N)**: N 

**Additional context**

danabens:
fixes #3673

*Issue #, if available:* #3673

*Description of changes:* Fixes a bug in `load_run` which would result in a double run created

*Testing done:* 

`tox -e py39 -- tests/unit/sagemaker/experiments`

## Merge Checklist

#### General

#### Tests


src/sagemaker/experiments/run.py
```diff
@@ -633,7 +633,8 @@ def _extract_run_name_from_tc_name(trial_component_name: str, experiment_name: s
         Returns:
             str: The name of the Run object supplied by a user.
         """
-        return trial_component_name.replace("{}{}".format(experiment_name, DELIMITER), "", 1)
+        # TODO: we should revert the lower casting once backend fix reaches prod
+        return trial_component_name.replace("{}{}".format(experiment_name.lower(), DELIMITER), "", 1)
```
qidewenwhen: So I assume this line would fix the issue of second run created? And will fix the list_runs issue https://github.com/aws/sagemaker-python-sdk/issues/3673?

Can we add a unit test which does list_runs twice in case to validate in case any future changes break it again?

src/sagemaker/experiments/run.py
```diff
@@ -843,7 +844,6 @@ def list_runs(
     created_after: Optional[datetime.datetime] = None,
     sagemaker_session: Optional["Session"] = None,
     max_results: Optional[int] = None,
-    next_token: Optional[str] = None,
```
qidewenwhen, line 846: To remove a field from a public method can be a breaking change. Is it causing any issues so we need to remove it?

Or are we sure the `list_runs` is not widely used? If so it's fine to remove it

src/sagemaker/experiments/run.py
```diff
@@ -861,14 +861,15 @@ def list_runs(
...
-        next_token (str): Token for next page of results (default: None).
         sort_by (SortByType): The property to sort results by. One of NAME, CREATION_TIME
             (default: CREATION_TIME).
         sort_order (SortOrderType): One of ASCENDING, or DESCENDING (default: DESCENDING).
 
     Returns:
         list: A list of ``Run`` objects.
     """
+
+    # all trial components retrieved by default
```
qidewenwhen, line 874: You mean the `_TrialComponent.list` can retrieve all trial components?
https://github.com/aws/sagemaker-python-sdk/blob/8d2c16bb912f7a86f8598b1fa799fe6981e79dba/src/sagemaker/experiments/trial_component.py#L248-L262

Or the list trial component API can retrieve all TCs in one call?

If the next_token is not causing any issues, I'd suggest to keep it to avoid a breaking change

src/sagemaker/experiments/run.py
```diff
@@ -633,7 +633,10 @@ def _extract_run_name_from_tc_name(trial_component_name: str, experiment_name: s
         Returns:
             str: The name of the Run object supplied by a user.
         """
-        return trial_component_name.replace("{}{}".format(experiment_name, DELIMITER), "", 1)
+        # TODO: we should revert the lower casting once backend fix reaches prod
+        return trial_component_name.replace(
+            "{}{}".format(experiment_name.lower(), DELIMITER), "", 1
+        )
```
qidewenwhen, line 639: Good catch! So I assume this line would fix the issue of second run created? And will fix the list_runs issue https://github.com/aws/sagemaker-python-sdk/issues/3673?

Can we add a unit test which does list_runs twice in case to validate in case any future changes break it again?

src/sagemaker/experiments/run.py
danabens: the argument is unusable as it currently is, so no one is currently using it

src/sagemaker/experiments/run.py
danabens: all trial components are retrieved by default:

https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/apiutils/_base_types.py#L118-L147

The problem with this parameter is where is the customer going to get the next_token to supply to method in the first place? The only way to get next_token is to call `ListTrialComponents` in which case they have no reason to use this method. 

Also this method doesn't return a `NextToken`.

src/sagemaker/experiments/run.py
danabens: sure, will do

src/sagemaker/experiments/run.py
danabens: I prefer to remove since it is not providing useful behavior as it currently is. In the future we could re-add it but will need to return a NextToken so the customer can supply it in their own page-loop behavior.

src/sagemaker/experiments/run.py
leoroy: +1 on @qidewenwhen 's comment.

Can we keep this parameter in public method but instead doing nothing with it?

danabens: ill separate the next_token change from the bug fix and we can discuss what to do with it separately

danabens:
Updates:
* remove removal of `next_token` and created backlog item
* updated `list_runs` test to include a mixed case experiment name argument to cover root cause of issue

leoroy (approved):

leoroy (approved):

PR #3870, opened 2023-05-18
fix: intelligent defaults for volume size, JS Estimator image uri region, Predictor str method

evakravi:
*Description of changes:*
1. Fixes intelligent defaults for when a KMS key used with an unsupported instance type
2. Fixes bug with `JumpStartEstimator` where the region wasn't being used with the image_uri
3. Add `__str__` method to base `Predictor` class
4. Fix `volume_size_supported` for local mode
5. Add `ignore_intelligent_defaults` to `SessionSettings` 

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/sagemaker/base_predictor.py
```diff
@@ -75,6 +75,10 @@ def content_type(self) -> str:
     def accept(self) -> Tuple[str]:
         """The content type(s) that are expected from the inference server."""
 
+    def __str__(self) -> str:
+        """Overriding str(*) method to make more human-readable."""
+        return stringify_object(self)
```
JGuinegagne, line 80: this looks to me like a good improvement, but please double check with SDK team if that's considered backward incompatible please. The alternative to be to write a different serialization function.

src/sagemaker/estimator.py
```diff
@@ -599,10 +601,30 @@ def __init__(
...
+            )
+            or instance_groups is not None
+            and all(
+                [
+                    instance_supports_kms(instance_group.instance_type)
+                    for instance_group in instance_groups
+                ]
+            )
+            else volume_kms_key
         )
```
JGuinegagne, line 628: nit: I find ternary hard to read in this context, could you do:
```py
if <condition>:
  self.volume_kms_key = resolve_value_from_config(...)
else:
  self.volume_kms_key = volume_kms_key
```

More importantly, from a security/reasonable expection POV: should we let this code throw if any of the instance in the group _does_ support KMS, but some do not?

If we silently remove it as you do here, the result is that there will be instances with EBS volumes out there that aren't encrypted with the default KMS key provided by admins. I don't think that's desirable.

src/sagemaker/session.py
```diff
@@ -3850,7 +3861,16 @@ def create_endpoint_config_from_existing(
 
         if new_kms_key is not None or existing_endpoint_config_desc.get("KmsKeyId") is not None:
             request["KmsKeyId"] = new_kms_key or existing_endpoint_config_desc.get("KmsKeyId")
-        if KMS_KEY_ID not in request:
+
+        supports_kms = all(
```
JGuinegagne: same comment here please, I would vote for letting it fail in that case.

src/sagemaker/session.py
```diff
@@ -4471,15 +4491,28 @@ def endpoint_from_production_variants(
         Returns:
             str: The name of the created ``Endpoint``.
         """
+
+        supports_kms = all(
```
JGuinegagne: same comment here please

tests/unit/test_estimator.py
```diff
@@ -453,6 +453,50 @@ def test_estimator_initialization_with_sagemaker_config_injection(sagemaker_sess
...
+    assert estimator.role == expected_role_arn
+    assert estimator.enable_network_isolation() == expected_enable_network_isolation
+    assert (
+        estimator.encrypt_inter_container_traffic
+        == expected_enable_inter_container_traffic_encryption
+    )
+    assert estimator.output_kms_key == expected_kms_key_id
+    assert estimator.volume_kms_key is None
+    assert estimator.security_group_ids == expected_security_groups
+    assert estimator.subnets == expected_subnets
```
JGuinegagne, line 497: please add a test case where the instance group has types that support attaching a volume and types that do not support it.

tests/unit/test_session.py
```diff
@@ -1633,6 +1633,102 @@ def test_train_with_sagemaker_config_injection(sagemaker_session):
...
+    )
+    assert actual_train_args["RoleArn"] == expected_role_arn
+    assert actual_train_args["ResourceConfig"] == {
+        "InstanceCount": INSTANCE_COUNT,
+        "InstanceType": "ml.g5.12xlarge",
+    }
+    assert actual_train_args["OutputDataConfig"] == {
+        "S3OutputPath": S3_OUTPUT,
+        "KmsKeyId": expected_kms_key_id,
+    }
```
JGuinegagne, line 1729: same comment here please

JGuinegagne (changes requested):

tests/unit/test_session.py
```diff
@@ -2932,6 +3047,127 @@ def test_endpoint_from_production_variants_with_sagemaker_config_injection(
...
+
+    expected_async_inference_config_dict = AsyncInferenceConfig()._to_request_dict()
+    expected_async_inference_config_dict["OutputConfig"]["KmsKeyId"] = expected_inference_kms_key_id
+    expected_pvs = [
+        sagemaker.production_variant("A", "ml.g5.xlarge"),
+        sagemaker.production_variant("B", "ml.p2.xlarge"),
+        sagemaker.production_variant("C", "ml.p2.xlarge"),
+    ]
+    # Add DestinationS3Uri, KmsKeyId to only one production variant
+    expected_production_variant_0_kms_key_id = SAGEMAKER_CONFIG_ENDPOINT_CONFIG["SageMaker"][
```
JGuinegagne, line 3090: same comment please

JGuinegagne (approved):

src/sagemaker/estimator.py
```diff
@@ -599,10 +601,33 @@ def __init__(
...
+        )
+
+        self.volume_kms_key = (
+            resolve_value_from_config(
+                volume_kms_key,
+                TRAINING_JOB_VOLUME_KMS_KEY_ID_PATH,
+                sagemaker_session=self.sagemaker_session,
+            )
+            if use_volume_kms_config
+            else volume_kms_key
```
rubanh, line 630: What are the alternate options to skipping volume kms key default configs that the admin set? If an admin set volume kms keys in the config, is it reasonable for them to expect that it will only be plugged in for only some instance types? Or would they prefer that their users only use instance types that support kms keys?

I think its worth looping in Prateek to discuss whether this is is the right approach

tests/unit/test_session.py
```diff
@@ -2798,6 +3047,127 @@ def test_endpoint_from_production_variants_with_sagemaker_config_injection(
     )
 
 
+def test_endpoint_from_production_variants_with_sagemaker_config_injection_partial_kms_support(
```
rubanh, line 3050: Thanks for adding these tests. Optional but I would prefer one or more comments for each test that make it clear what is being verified. Based on the title alone I can imagine it being difficult for someone reading through later to find whats different in the contents of one test vs another. 

Also, do we need to verify all the configs for these kms_support tests? I wonder if we could get rid of all the asserts that dont relate to the thing being tested?

rubanh (changes requested):

src/sagemaker/session.py
```diff
@@ -855,6 +856,12 @@ def train(  # noqa: C901
         inferred_resource_config = update_nested_dictionary_with_values_from_config(
             resource_config, TRAINING_JOB_RESOURCE_CONFIG_PATH, sagemaker_session=self
         )
+        if (
+            "InstanceType" in inferred_resource_config
+            and not instance_supports_kms(inferred_resource_config["InstanceType"])
+            and "VolumeKmsKeyId" in inferred_resource_config
+        ):
+            del inferred_resource_config["VolumeKmsKeyId"]
```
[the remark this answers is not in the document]
evakravi: These instances don't have any notion of KMS, so there's no security problem, AFAIK. If *any* instance though supports KMS, we always apply the KMS key.

src/sagemaker/estimator.py
```diff
@@ -599,10 +601,33 @@ def __init__(
...
-        self.volume_kms_key = resolve_value_from_config(
-            volume_kms_key,
-            TRAINING_JOB_VOLUME_KMS_KEY_ID_PATH,
-            sagemaker_session=self.sagemaker_session,
+        if instance_type is None or isinstance(instance_type, str):
+            instance_type_for_volume_kms = instance_type
+        elif isinstance(instance_type, ParameterString):
+            instance_type_for_volume_kms = instance_type.default_value
+        else:
+            raise ValueError(f"Bad value for instance type: '{instance_type}'")
```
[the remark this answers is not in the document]
evakravi: I'm pretty sure. The unit tests make it seem like it's compatible. There shouldn't be a different python type for the instance type.

src/sagemaker/session.py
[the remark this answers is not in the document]
evakravi: The admin sets the KMS config with the expectation it is used whenever KMS is supported. But if someone uses a instance type like `local` mode or a g5 instance, the KMS settings makes no sense, and shouldn't be applied.

src/sagemaker/utils.py
```diff
@@ -1408,3 +1416,47 @@ def update_nested_dictionary_with_values_from_config(
...
+
+        # Any instance type with a "d" in the instance family (i.e. c5d, p4d, etc) + g5
+        # does not support attaching an EBS volume.
+        family = parts[0]
+        return "d" not in family and not family.startswith("g5")
+    except Exception as e:
+        raise ValueError(f"Failed to parse instance type '{instance_type}': {str(e)}")
+
+
+def instance_supports_kms(instance_type: str) -> bool:
```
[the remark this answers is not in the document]
evakravi: The rules for supporting KMS are equal to whether the instance supports volume size. Unfortunately, there's no other source of truth, so this will have to be updated manually. However, this function is already used in other parts of the code (JumpStart).

src/sagemaker/session.py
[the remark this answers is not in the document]
JGuinegagne: Let me add some more context here.

The only thing you can encrypt with a custom KMS key is an EBS volume. The issue here is that you cannot attach an EBS volume to these types of instances. See [documentation](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateEndpointConfig.html#sagemaker-CreateEndpointConfig-request-KmsKeyId).

As a result, passing any KMS key with such instance is nonsensical, the only admissible value is `None`. The issue is, if the user attempts to pass `None`, the intelligent default can substitute a value, causing the service call to fail and the user has no recourse.

src/sagemaker/base_predictor.py
mufaddal-rohawala: Agree this would be a good improvement, I dont see any negative implications of this.

src/sagemaker/estimator.py
mufaddal-rohawala: I think it's fine to have this conditional logic as we dont want to limit users to use only those instances which support volume kms key.
cc - @prateekmehrotra

src/sagemaker/estimator.py
mufaddal-rohawala: I think this is a valid concern with respect to instance_groups, lets involve the instance_groups feature team to weigh in on this.

src/sagemaker/session.py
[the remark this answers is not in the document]
mufaddal-rohawala: >As a result, passing any KMS key with such instance is nonsensical, the only admissible value is None. The issue is, if the user attempts to pass None, the intelligent default can substitute a value, causing the service call to fail and the user has no recourse.

AFAIK, intelligent defaults will take the user supplied value (None) over the ones set in Inteliigent defaults. @rubanh correct me if Im wrong. But with the flexibility we want to provide with PySDK, the user need not figure out that the instance doesn't support kms volume and explicitly pass None for such a case.

src/sagemaker/utils.py
```diff
@@ -1362,3 +1371,47 @@ def update_nested_dictionary_with_values_from_config(
...
+
+    try:
+
+        # local mode does not support volume size
+        if instance_type.startswith("local"):
+            return False
+
+        parts: List[str] = instance_type.split(".")
+
+        if len(parts) == 3 and parts[0] == "ml":
```
mufaddal-rohawala, line 1397: nit: can use some regex magic to check for instancetype input - https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/image_uris.py#LL266C10-L266C10

evakravi: Discussed offline that the other utility has different specifications and would be large blast radius to merge utilities in this PR.

tests/unit/test_session.py
mufaddal-rohawala: +1

src/sagemaker/utils.py
```diff
@@ -1066,6 +1065,15 @@ def resolve_value_from_config(
     Returns:
         The value that should be used by the caller
     """
+
+    if (
+        sagemaker_session is not None
+        and sagemaker_session.settings is not None
+        and sagemaker_session.settings.ignore_intelligent_defaults
```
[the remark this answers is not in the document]
rubanh: So Im wondering why we dont want to do something like the following (ignore at initialization rather than when its time to use it). Seems like this could limit potential for confusion. But curious if there's usecases you have in mind that make this approach unideal.

```
class Session(object):
...
    def _initialize(...):
       ...
        if self.settings and sagemaker_session.settings.ignore_sagemaker_config: # new
            self.sagemaker_config = {}  # new
        elif sagemaker_config:
            validate_sagemaker_config(sagemaker_config)
            self.sagemaker_config = sagemaker_config
        elif not (self.settings and self.settings.ignore_sagemaker_config):
            self.sagemaker_config = load_sagemaker_config(s3_resource=self.s3_resource)       
```

src/sagemaker/session.py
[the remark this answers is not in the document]
rubanh: So actually @JGuinegagne is right about this case @mufaddal-rohawala , if the user supplies None, the Defaults Config understands that as "the user didnt supply a value" (because it doesnt know if the None was a explicitly or implicitly provided)

And then to followup on my original comment, thanks for linking to documentation.

So lets break this down into separate questions.
1. Should the VolumeKMS key be skipped if all instances in a request have local storage when...
     1. ... a user passes in a non-None VolumeKMS key explicitly?
Right now, seems this PR in most cases says NO to this question. But in this one case above it says YES
     1. ... the VolumeKMS key was explicitly set in the DefaultConfig?
The goal of this PR is YES. If we are going with this I think at minimum we need to make sure Admins understand this behavior when they set a VolumeKMS key in the Config.
1. Should the VolumeKMS key be skipped if even one instance in a request supports KMS? NO.

src/sagemaker/estimator.py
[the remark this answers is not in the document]
rubanh: Today if a user passes in a different python type, they wont get an error right? (My understanding is that the "Optional[Union[str, PipelineVariable]]" is not strictly enforced by Python). Let me know if im mistaken

If Im not wrong about the above, this would be changing behavior of the Estimator class (and I havent seen type assertions like these in other places anyway). Seems safer to not change this behavior

src/sagemaker/utils.py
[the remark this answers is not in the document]
rubanh: I'll defer to @mufaddal-rohawala 's judgement on this, but I would say that this function needs to be explicit and clear about (1) why is kms support dependent on volume size and (2) how to check whether the answers given by this function are up to date. Those things would help with maintainability, especially if this is something that needs to be updated manually

src/sagemaker/estimator.py
rubanh: And I suppose the whole purpose of the Config is to improve user experience. If the customer wants to enforce that KMS keys be used always, the right place for them to strictly enforce that would be through IAM Policies, SCPs, etc. There is no guarantee that populating the Config means those values will be used no matter what.

src/sagemaker/utils.py
```diff
@@ -1097,6 +1097,15 @@ def get_sagemaker_config_value(sagemaker_session, key):
     """
     if not sagemaker_session:
         return None
+
+    if (
+        sagemaker_session is not None
+        and sagemaker_session.settings is not None
+        and sagemaker_session.settings.ignore_sagemaker_config
+    ):
+        logger.info("Ignoring defaults config.")
```
rubanh, line 1106: This should say something like the following no? (Otherwise its not clear why its being ignored)
"Ignoring defaults config because ignore_sagemaker_config was set in Session settings"

Also, now that the logging commit for the DefaultConfig is out, this should be logged to the logger from get_sagemaker_config_logger(), instead of to the sagemaker logger.

src/sagemaker/utils.py
```diff
@@ -1097,6 +1097,15 @@ def get_sagemaker_config_value(sagemaker_session, key):
     """
     if not sagemaker_session:
         return None
+
+    if (
+        sagemaker_session is not None
```
rubanh, line 1102: nit: this can be removed because of the check above

src/sagemaker/estimator.py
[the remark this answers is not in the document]
evakravi: Ok, I can skip throwing an exception. However, if a variable that isn't a string or `PipelineVariable` is used, there will probably still be an exception thrown somewhere

rubanh (approved):

mufaddal-rohawala (commented):
/bot run slow-tests, notebook-tests

src/sagemaker/estimator.py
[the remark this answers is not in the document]
pviolette3: https://github.com/aws/sagemaker-python-sdk/issues/3917 - let me know how I can help

src/sagemaker/utils.py
```diff
@@ -1362,3 +1363,47 @@ def update_nested_dictionary_with_values_from_config(
...
+    """Returns True if SageMaker allows volume_size to be used for the instance type.
+
+    Raises:
+        ValueError: If the instance type is improperly formatted.
+    """
+
+    try:
+
+        # local mode does not support volume size
+        if instance_type.startswith("local"):
```
jerrypeng7773, line 1384: will need to add another check `or is_pipeline_variable(instance_type)`, otherwise, this breaks pipeline cx.

src/sagemaker/estimator.py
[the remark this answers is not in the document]
qidewenwhen: It turns out this is not backward compatible, see the issue https://github.com/aws/sagemaker-python-sdk/issues/3993

Have we made the fix to skip PipelineVariable type?

PR #3965, opened 2023-06-28
feat: Enable customizing artifact output path

### LINKED ISSUE #3609: Customizing the artifact output path.
Before Experiments were integrated with the base sagemaker library, I could customize the path where the training artifacts are stored with Tracker, since the load method allowed for specifying the artifact bucket and the artifact prefix:

`Tracker.load(artifact_bucket='xxx',  artifact_prefix='yyy')`

The new, integrated sagemaker.experiments no longer allows it in load_run
https://github.com/aws/sagemaker-python-sdk/blob/5cb00a4d9042f96814582006161639527ebc1e15/src/sagemaker/experiments/run.py#L713-L717

I would like to be able to customize the path. At the moment the bucket and prefix sit in the _helper.py in the _ArtifactUploader class, the prefix seems to be hardcoded https://github.com/aws/sagemaker-python-sdk/blob/5cb00a4d9042f96814582006161639527ebc1e15/src/sagemaker/experiments/_helper.py#L31-L40

There does not seem to be a way to customize the output path for artifacts anymore, and that is something that has come extremely handy while using the old SDK.

Being able to define the bucket AND the prefix where the output artifacts would be stored.

The only alternative I have found is using the old SDK - if someone has another idea, please let me know.

rd-pong:
*Issue #, if available:*
#3609

*Description of changes:*
- Allow customers to define the bucket and the prefix where the output artifacts would be stored.
- Add `artifact_bucket` and `artifact_prefix` to `Run` constructor
- Add `artifact_bucket` and `artifact_prefix` to `load_run`

*Testing done:*
- Existing unit tests pass
- New unit test added to cover this case 
  - `tests/unit/sagemaker/experiments/test_run.py`
  - `tests/unit/sagemaker/experiments/helpers.py`

## Merge Checklist

#### General

#### Tests


knikure (commented):
/bot run all

src/sagemaker/experiments/run.py
```diff
@@ -29,6 +29,7 @@
 from sagemaker.experiments._helper import (
     _ArtifactUploader,
     _LineageArtifactTracker,
+    _DEFAULT_ARTIFACT_PREFIX,
```
jiahaoli97, line 32: Why we are using UpperCase, should we keep consistent with exist parameter?

rd-pong: `_DEFAULT_ARTIFACT_PREFIX` is not a new constant introduced in this commit. It is an existing constant defined in `src/sagemaker/experiments/_helper.py`

jiahaoli97 (approved):

src/sagemaker/experiments/run.py
danabens, line 32: nit the name of this sounds like it will prefix the actual name of the artifact but its really a prefix for a s3 path where the associated files will be stored

src/sagemaker/experiments/run.py
```diff
@@ -197,6 +205,10 @@ def __init__(
         self._artifact_uploader = _ArtifactUploader(
             trial_component_name=self._trial_component.trial_component_name,
             sagemaker_session=sagemaker_session,
+            artifact_bucket=artifact_bucket,
+            artifact_prefix=_DEFAULT_ARTIFACT_PREFIX
+            if artifact_prefix is None
+            else artifact_prefix,
```
danabens, line 211: nit this indentation is unintuitive

danabens: nm looks like there is not many better alternatives https://stackoverflow.com/questions/28897010/how-to-make-a-line-break-on-the-python-ternary-operator

src/sagemaker/experiments/run.py
rd-pong: Yes, it's a bit confusing to use this name here. But `_DEFAULT_ARTIFACT_PREFIX` is not introduced in this PR, it's already defined in [_helper.py](https://github.com/aws/sagemaker-python-sdk/blob/9caf00f0d266a1bac9275d67fada28b5c849310a/src/sagemaker/experiments/_helper.py#L28). There is a variable called `artifact_prefix`, so I'm assuming that is why we called the default variable to be `_DEFAULT_ARTIFACT_PREFIX`.

A workaround could be change the following method defined in `_ArtifactUploader`.
```
class _ArtifactUploader(object):
    """Artifact uploader"""
    def __init__(
        self,
        trial_component_name,
        sagemaker_session,
        artifact_bucket=None,
        artifact_prefix=None, # HERE!!!
    ):
        self.sagemaker_session = sagemaker_session
        self.trial_component_name = trial_component_name
        self.artifact_bucket = artifact_bucket
        self.artifact_prefix = (
            _DEFAULT_ARTIFACT_PREFIX if artifact_prefix is None else artifact_prefix # HERE!!!
        )
        self._s3_client = self.sagemaker_session.boto_session.client("s3")
```
In this way, we don't need to import `_DEFAULT_ARTIFACT_PREFIX` to run.py

danabens (approved):

knikure (commented):
/bot run all

knikure (approved):

knikure (approved):

Issue #1200, opened 2019-12-26
Unable to use this library in AWS Lambda due to package size exceeded max limit


nemalipuri:
Please fill out the form below.

### System Information
- **AWS Lambda**:
- **Python v3.6**:
- **Sagemaker Python SDK 1.49.0**:

### Describe the problem
I'm trying to use Sagemaker Python SDK in Lambda to trigger train and deploy steps. Packaged the dependencies along with function code and when trying to create Lambda function it is throwing error 'Unzipped size must be smaller than 262144000 bytes'

**Sorry, though this issue is related to Lambda service limit I want to check is there anyway I can reduce the size of the dependencies?**

I have tried removing boto3 and botocare from function zip file since Lambda provides these libraries but it lead to different issue 'expecting python-dateutil<2.8.1,>=2.1'

### Minimal repro / logs
AWS Lambda error 'Unzipped size must be smaller than 262144000 bytes'

- **Exact command to reproduce**:
mkdir python
cd python
pip install sagemaker --target .
chmod 777 python
zip python directory
Upload Zip file to S3
Error when creating AWS Layer 'Failed to create layer version: Unzipped size must be smaller than 262144000 bytes'

Similarly, instead of Laye when packaged code with dependencies and uploading the zip file into Lambda function I received error 'Unzipped size must be smaller than 262144000 bytes'

Appreciate your help.

knakad:
Hi @nemalipuri !

Unfortunately, running sagemaker-python-sdk in AWS Lambda is not currently supported. This is a pain point that we're aware of and for which we are working on prioritizing a solution.

I would normally recommend pinning python-dateutil to 2.8.0 to resolve the conflict, but I actually experimented locally and found that, even without boto3, the zip (55MB) is still over the 50MB zipped limit for Lambda.

An alternative is to remove numpy and scipy dependencies entirely for specific sagemaker installations, as they account for ~73% of the installation size.
In order for me to gauge the solution's viability, can you tell me if you will need numpy/scipy functionality when running sagemaker-python-sdk in AWS Lambda?
Similarly, what are your sagemaker-python-sdk AWS Lambda use-cases?

Thanks!

nemalipuri:
@knakad Thanks for looking into this.

Almost a year back I've used Sagemaker Python SDK in Lambda without any issues, the version it was 1.18.0 and size of the package was smaller.

[...]

PR #4222, opened 2023-10-23
change: makes pandas optional

trungleduc:
*Issue #, if available:*

- closes https://github.com/aws/sagemaker-python-sdk/issues/1200
- The SDK still installs `pandas` by default but won't raise any error without it. Users can install the SDK without pandas by following the guide in the documentation  https://sagemaker--4222.org.readthedocs.build/en/4222/installation.html

*Description of changes:*
- All pandas import calls are guarded behind a `try-except` block with `DeferredError`
- `DataFrame` is imported under `TYPE_CHECKING` check
- New helper function `is_likely_a_pandas_df` to heuristically check if the variable is a `DataFrame` 
- New "Getting started" section in the documentation.

*Testing done:*

- unit test added

## Merge Checklist

#### General

#### Tests


requirements/extras/pandas_requirements.txt
```diff
@@ -0,0 +1 @@
+pandas==2.1.1
```
jjerphan, line 1: We could use [compatible releases](https://peps.python.org/pep-0440/#compatible-release) to be less strict.
```suggestion
pandas~=2.1.1
```

src/sagemaker/utils.py
```diff
@@ -1449,3 +1449,12 @@ def get_instance_type_family(instance_type: str) -> str:
...
     return instance_type_family
+
+
+def is_likely_a_pandas_df(data: Any) -> bool:
+    """Return True if data is likely a pandas dataframe."""
+    return (
+        data.__class__.__name__ == "DataFrame"
+        and hasattr(data, "columns")
+        and hasattr(data, "iloc")
+    )
```
jjerphan, line 1460: I think we must test this utility function.

We could use [`pytest.importorskip`](https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-on-a-missing-import-dependency) in the test:

```python
pd = pytest.importorskip("pandas")
```

requirements/extras/pandas_requirements.txt
jjerphan: Or we could use the same version range as the one provided in
`requirements/extras/test_requirements.txt` for consistency:

https://github.com/aws/sagemaker-python-sdk/blob/4befd9320ac7768b8d64c371cf9f1bd356baebb8/requirements/extras/test_requirements.txt#L23

src/sagemaker/tuner.py
jjerphan, line 1: Nitpick: we could omit this change.

src/sagemaker/feature_store/dataset_builder.py
```diff
@@ -19,14 +19,27 @@
...
 from sagemaker.feature_store.feature_group import FeatureDefinition, FeatureGroup, FeatureTypeEnum
 
+from sagemaker.utils import DeferredError, is_likely_a_pandas_df
+
+logger = logging.getLogger(__name__)
+
+try:
+    import pandas as pd
+except ImportError as e:
+    logger.warning("pandas failed to import. Feature store features will be impaired or broken.")
```
jjerphan, line 37: Would being explicit help in this case?

```suggestion
    logger.warning("pandas is not installed in your environment. Feature store features will be impaired or broken.")
```

This suggestion also applies to other warning messages.

jjerphan (commented):
Thank you, @trungleduc.

Let's see if some configurations tested on the CI break with this change.

In the meantime, here is a first review.

src/sagemaker/feature_store/dataset_builder.py
trungleduc: I want to keep it consistent with other messages of the SDK and also this message reflects correctly the import issue

setup.py
```diff
@@ -56,8 +56,8 @@ def read_requirements(filename):
     "smdebug_rulesconfig==1.0.1",
     "importlib-metadata>=1.4.0,<7.0",
     "packaging>=20.0",
-    "pandas",
```
mufaddal-rohawala, line 59: if we are adding pandas dependency as an extra, we might need to add documentation around using the extras install for use cases which involve using `pandas`.

src/sagemaker/feature_store/dataset_builder.py
mufaddal-rohawala: Can we also add an instruction to install from extras, would it help for users to unblock when hit with this issue.

setup.py
mufaddal-rohawala: This I think will be a breaking change, customers who auto consume newest pysdk release with this fix would now need to also install pandas (in case they use it). But I do still think this is a valuable change considering the use-cases of supporting lambda.

setup.py
```diff
@@ -56,8 +56,8 @@ def read_requirements(filename):
...
     "pathos",
+    "pytz",
```
mufaddal-rohawala, line 60: do we need this new dependency, how much of a file size addition is this? if needed can we rely on existing dependencies (/some dependency from python standard library)

setup.py
mufaddal-rohawala: Can we try this approach where we can provide documentation for minimal installation like:
```
pip install --no-deps sagemaker
pip install -r minimal_requirements.txt // this has the minimal requirements without pandas.
```
we can put this in a shell script and call it lambda-install.sh or something and put it in sdk,

trungleduc: ok. i'm on it

setup.py
trungleduc: https://github.com/aws/sagemaker-python-sdk/blob/a4ba730b906f1b15504da471157ac6af513ebde4/src/sagemaker/feature_store/feature_processor/feature_scheduler.py#L21

it's used in the code base but is installed implicitly through `pandas`. I wanted to address it in a separate PR. But can also do it here

setup.py
trungleduc: There is no easy way to provide the `minimal_requirements.txt` file to users, since installing the SDK with pip only copies the Python files to the install path, not the requirement files.

trungleduc: Do you want to include requirement files in the Pypi package?

setup.py
mufaddal-rohawala: Yeah we should really really careful while adding new dependency and only do it if absolutely necessary, otherwise use extras [added dependency]

setup.py
mufaddal-rohawala: is there a way to do it? the only way to solve the issue in a backwards compatible way is to provide documentation to users for AWS Lambda support with no-deps instructions above.

mufaddal-rohawala: if its not possible to include files in pip dist, lets just provide adequate documentation to users.

setup.py
jjerphan: Pandas < 2.0 necessitates `pytz` but [Pandas>=2.0 now uses `datetime.timezone` by default](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#utc-and-fixed-offset-timezones-default-to-standard-library-tzinfo-objects).

Probably `pytz` was introduced when installing Pandas < 2.0.

If we depends on Pandas 2.0, we can safely remove `pytz` as a constraint.

setup.py
trungleduc: Yes, it's possible

trungleduc: The requirement files are not stored in the source folder of the SDK, so shipping them with the SDK is not recommended. I went with the documentation route.

trungleduc: https://sagemaker--4222.org.readthedocs.build/en/4222/installation.html

PR #4566, opened 2024-04-08
feat: JumpStart alternative config parsing

Captainia:
*Issue #, if available:*

*Description of changes:*

Initial round of reading and parsing the JumpStart alternative configs.

Class structures that goes from bottom to top:

                           JumpStartMetadataBaseFields 
                 (inherits) /                       \(inherits)
           JumpStartModelSpecs             JumpStartConfigComponent   
              (includes)/                            \ (consists)
       JumpStartMetadataConfigs                    JumpStartMetadataConfig
                                                             \ (consists)
                                                    JumpStartMetadataConfigs
*Testing done:*

## Merge Checklist

#### General

#### Tests


src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
         return alias_value
 
 
+class JumpStartBenchmarkStat(JumpStartDataHolderType):
+    """Data class JumpStart benchmark stats."""
+
+    __slots__ = ["benchmark_name", "instance_type", "latency_ms"]
```
evakravi, line 742: can you base the benchmark stats off the field in this table? https://quip-amazon.com/NamhAEZUkUFz/QuickSilver-Python-SDK-Demo#temp:C:ZWOc84db6cd2c4e43c8b5661cb88 i updated the mock metadata

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+        self.instance_type: str = json_obj["instance_type"]
+        self.latency_ms: int = json_obj["latency_ms"]
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartBenchmarkStat object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+
+class JumpStartInferenceComponent(JumpStartDataHolderType):
```
evakravi, line 768: can we rename this `JumpStartMetadataComponent`, and have 1 type for training and inference?

`InferenceComponent` could get confused with goldfinch

evakravi: the idea with these classes is to be able to override _every_ field of the spec, with the exception of some (hard code the exception list). And ideally this parity would be maintained automatically.  Can yo refactor `JumpStartModelSpecs` to inherit another class/interface `JumpStartRawMetadataFields` that has all the fields, and then have `class JumpStartMetadataComponent(JumpStartRawMetadataFields)` inherit the class?

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+            "supported_inference_configs"
+        )
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartTrainingComponent object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+
+class JumpStartAlternativeConfig(JumpStartDataHolderType):
```
evakravi, line 873: Can we rename `JumpStartDeploymentConfig`?

src/sagemaker/jumpstart/types.py
Captainia: Makes a lot of sense, I was under the impression that we are only overriding a subset of the specs as currently in this component. I'll refactor

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+        self.config_name = json_obj["config_name"]
+        self.component_names = json_obj["component_names"]
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartAlternativeConfig object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+    @property
+    def resolved_config(self) -> Dict[str, Any]:
```
evakravi, line 910: this logic won't work for dicts inside dicts. can you instead do:
```

def deep_override_dict(dict1: dict, dict2: dict) -> dict:
    """Overrides any overlapping contents of dict1 with the contents of dict2."""
    flattened_dict = flatten(dict1)
    flattened_dict.update(flatten(dict2))
    return unflatten(flattened_dict) if flattened_dict else {}
```

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+        self.description: str = json_obj["description"]
+        self.ranking: List[str] = json_obj["ranking"]
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartConfigRanking object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+
+class JumpStartInferenceAlternativeConfigs(JumpStartDataHolderType):
```
evakravi, line 959: same comment here about 1 class for inference and training

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+            if "overall" in self.config_ranking
+            else self.config_ranking.values(0)
+        )  # TODO: choose the best ranking
+
+        for config_name in ranking.ranking:
+            config = self.configs[config_name]
+            if instance_type in config.resolved_config["training_instance_type"]:
+                return config
+        raise RuntimeError(
+            f"No training alternative config found for instance type {instance_type}"
```
evakravi, line 1058: there's a lot of repeated code here, can we make a common parent class?

src/sagemaker/jumpstart/types.py
evakravi: no worries!

src/sagemaker/jumpstart/types.py
```diff
@@ -736,6 +736,334 @@ def _get_regional_property(
...
+
+    def from_json(self, json_obj: Dict[str, Any]) -> None:
+        """Sets fields in object based on json.
+
+        Args:
+            json_obj (Dict[str, Any]): Dictionary representation of benchmark stats.
+        """
+        self.benchmark_name: str = json_obj["benchmark_name"]
+        self.instance_type: str = json_obj["instance_type"]
+        self.latency_ms: int = json_obj["latency_ms"]
```
whittech1, line 760: IIUC, this could be an ever-expanding set of benchmark stats (i.e. latency, cpu utilization, gpu utilization, etc.), we should confirm the expected rate of change for Quicksilver to ensure enumerating every field is going to be a scalable approach.

src/sagemaker/jumpstart/types.py
whittech1: imo `DeploymentConfig` is an overloaded term because "deployment" often means Hosting. I'd lobby for `JumpStartPresetConfig`

src/sagemaker/jumpstart/types.py
```diff
@@ -894,6 +1229,43 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
             if json_obj.get("hosting_instance_type_variants")
             else None
         )
+        self.benchmarh_stats: Optional[Dict[str, JumpStartBenchmarkStat]] = (
```
whittech1, line 1232: nit: `benchmark`

src/sagemaker/jumpstart/types.py
whittech1: @evakravi @Captainia let's please not use "raw metadata" to mean anything outside of the contributor-side config. Otherwise, the terminology becomes very confusing. `JumpstartMetadataFields` should be fine.

We should also align on the overall approach. Here are my thoughts:

1. Each "config_name" from the design is called a `Preset`
2. `JumpStartModelSpecs` contains a set of `Presets`, but uses the default one for its initialization of fields
3.  `Presets` are constructed from an array of `ConfigSets` (i.e. `component` -> `ConfigSet`)
4. No distinction between `InferenceConfigSet` or `TrainingConfigSet`, there is just `ConfigSet`
5. `ConfigSet` contains all the JumpStart fields

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
                     json_obj[att] = cur_val
         return json_obj
 
+
+class JumpStartPresetComponent(JumpStartMetadataBaseFields):
```
evakravi, line 1007: later we should refactor this file, it's wayyy too big

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
...
+    """Data class of JumpStart config component."""
+
+    slots = ["component_name", "component_override_fields"]
```
evakravi, line 1010: what is `component_override_fields`?

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
...
+        Deep merge two dictionaries recursively, with override overriding source.
+        """
+        for key, value in source.items():
+            if (
+                key not in JumpStartPresetComponent.OVERRIDING_DENY_LIST
+                and key in override
+                and isinstance(value, dict)
+                and isinstance(override[key], dict)
+            ):
+                JumpStartPresetComponent._override_merge(value, override[key])
```
evakravi, line 1050: generally not a fan of recursion, you can overflow the stack. what was wrong with the other solution i commented?

evakravi: i guess this is a nit, since our metadata isn't nested so much it'd cause the process to crash

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
...
+        """
+        super().from_json(spec)
+        self.component_name: str = component["component_name"]
+        self.component_overrides: JumpStartMetadataBaseFields = JumpStartMetadataBaseFields(
+            component
+        )
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartInferenceComponent object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
```
evakravi, line 1035: is this not already a method of a parent class?

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
...
+        return json_obj
+
+    @staticmethod
+    def _override_merge(source: Dict[str, Any], override: Dict[str, Any]):
```
evakravi, line 1039: this implementation will mutate `source`. can we make a `deepcopy` first?

src/sagemaker/jumpstart/types.py
```diff
@@ -944,6 +1003,273 @@ def to_json(self) -> Dict[str, Any]:
...
+            benchmark_metrics (Dict[str, JumpStartBenchmarkStat]):
+                The dictionary of benchmark metrics with name being the key.
+        """
+        self.preset_components: Dict[str, JumpStartPresetComponent] = preset_components
+        self.benchmark_metrics: Dict[str, JumpStartBenchmarkStat] = benchmark_metrics
+        self.resolved_preset_config: Optional[JumpStartPresetConfig] = None
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartInferencePresetConfig object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
```
evakravi, line 1091: same comment here about parent class method

src/sagemaker/jumpstart/types.py
evakravi: i guess this is fine, gives us flexibility in the future

src/sagemaker/jumpstart/types.py
Captainia: I remembered I used something similar before to this but right now I don't see a native Python supported method `flatten`.  There's something similar but doesn't seem it trivially support `unflatten`: https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html

src/sagemaker/jumpstart/types.py
Captainia: Yes this is not needed, I'll remove from the inherited classes

src/sagemaker/jumpstart/types.py
Captainia: yeah, I can move some types to a separate file

src/sagemaker/jumpstart/types.py
Captainia: It is the fields that this particular component will override to the `JumpStartMetadataBaseFields`. Each `JumpStartPresetComponent` inherits the base class so contains all the base fields, but since we only want to override to these specific fields so I kept a separate list only for them.

src/sagemaker/jumpstart/types.py
Captainia: Moved to utils with flatten dict method

src/sagemaker/jumpstart/types.py
```diff
@@ -1007,81 +1009,74 @@ def to_json(self) -> Dict[str, Any]:
...
-            component
-        )
+        self.component_name = component_name
+        self.from_json(component)
 
-    def to_json(self) -> Dict[str, Any]:
-        """Returns json representation of JumpStartInferenceComponent object."""
-        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
-        return json_obj
+    def from_json(self, json_obj: Dict[str, Any]) -> None:
```
evakravi, line 1045: can we not define this in a parent class?

src/sagemaker/jumpstart/types.py
```diff
@@ -1091,31 +1086,38 @@ def to_json(self) -> Dict[str, Any]:
         json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
         return json_obj
 
-    def resolve(self) -> JumpStartPresetComponent:
-        """Returns the final config that is resolved from the list of components."""
+    @property
```
evakravi, line 1089: does `@property` ensure this logic is only performed once and then cached?

src/sagemaker/jumpstart/types.py
```diff
@@ -1137,21 +1142,43 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
         json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
         return json_obj
 
-    def get_resolved_config(self) -> JumpStartPresetConfig:
-        """Resolves the preset config."""
-        # TODO: return the best preset config based on config rankings
-        return self.preset_configs[0] if self.preset_configs else None
+    def get_top_config_from_ranking(
+        self,
+        ranking_name: str = "overall",
```
evakravi, line 1160: maybe define `overall` as a constant somewhere?

src/sagemaker/jumpstart/types.py
```diff
@@ -1137,21 +1142,43 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
+        instance_type: Optional[str] = None,
+    ) -> JumpStartPresetConfig:
+        """Gets the best the preset config based on config ranking."""
+        if not self.preset_config_rankings or not self.preset_config_rankings.get(ranking_name):
```
evakravi, line 1164: can we log something? or maybe even throw an exception. this would only happen for bad metadata

src/sagemaker/jumpstart/types.py
```diff
@@ -1137,21 +1142,43 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
+            return list(self.preset_configs.values())[0] if self.preset_configs else None
+
+        instance_type_attribute = (
```
evakravi, line 1167: maybe throw an exception if the scope is not recognized

src/sagemaker/jumpstart/types.py
```diff
@@ -1137,21 +1142,43 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
+        rankings = self.preset_config_rankings.get(ranking_name)
+        for config_name in rankings.rankings:
+            resolved_config = self.preset_configs[config_name].resolved_config
+            if instance_type and instance_type not in getattr(
+                resolved_config, instance_type_attribute
+            ):
+                continue
+            return self.preset_configs[config_name]
+
+        return None
```
evakravi, line 1181: will this lead to an exception being thrown later?

src/sagemaker/jumpstart/types.py
```diff
@@ -1170,11 +1197,14 @@ class JumpStartModelSpecs(JumpStartMetadataBaseFields):
...
     def __init__(self, spec: Dict[str, Any]):
         """Initializes a JumpStartModelSpecs object from its json representation.
+        By default only apply the inference preset config.
 
         Args:
             spec (Dict[str, Any]): Dictionary representation of spec.
         """
         self.from_json(spec)
+        if self.inference_presets and self.inference_presets.get_top_config_from_ranking():
+            super().from_json(self.inference_presets.get_top_config_from_ranking().resolved_config)
```
evakravi, line 1207: im a bit confused, can you explain what's going on here?

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,99 @@ def get_region_fallback(
         return constants.JUMPSTART_DEFAULT_REGION_NAME
 
     return list(combined_regions)[0]
+
+
+def get_preset_names(
```
evakravi, line 885: we should rename preset -> config later

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,99 @@ def get_region_fallback(
...
+
+
+def get_benchmark_stats(
+    region: str,
+    model_id: str,
+    model_version: str,
+    config_names: Optional[List[str]] = None,
+    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
+    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
```
evakravi, line 919: what happens for proprietary models?

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,99 @@ def get_region_fallback(
...
+        presets = model_specs.training_presets
+
+    if not config_names:
+        config_names = presets.preset_configs.keys() if presets else []
+
+    benchmark_stats = {}
+    for config_name in config_names:
+        if config_name not in presets.preset_configs:
+            raise ValueError(f"Unknown preset config name: '{config_name}'")
+        benchmark_stats[config_name] = presets.preset_configs.get(config_name).benchmark_metrics
```
evakravi, line 943: i think you need to do `get(config_name, {})`, or else you'll get an error if the key isn't there

src/sagemaker/jumpstart/types.py
Captainia: I don't think it [gets cached with this tag automatically](https://stackoverflow.com/questions/17330160/how-does-the-property-decorator-work-in-python), that's why I stored it in a variable  `self.resolved_preset_config`

tests/unit/sagemaker/jumpstart/test_utils.py
```diff
@@ -1495,3 +1504,174 @@ def test_get_region_fallback_success(s3_bucket_name, s3_client, sagemaker_sessio
...
+class TestPresetConfigs:
+    @patch("sagemaker.jumpstart.accessors.JumpStartModelsAccessor.get_model_specs")
+    def test_get_jumpstart_presets_empty(
+        self,
+        patched_get_model_specs,
+    ):
+
+        patched_get_model_specs.side_effect = get_special_model_spec
+
+        assert utils.get_preset_names("mock-region", "gemma-model", "mock-model-version") == []
```
evakravi, line 1518: would we not want to raise an exception here?

src/sagemaker/jumpstart/types.py
Captainia: One `JumpStartModelSpec` has two set of `PresetConfigs` inside of it - one set for inference and one set for training. When initializing the class, we choose to apply the default inference `PresetConfig` from the overall ranking to override these fields. (Open to discussion though if we want to apply the best training config as well)

Alternatively, we can just remove this two lines of logic, so that the fields in JumpStartModelSpec stays with default base fields, until someone calls the `set` method with a config they want to apply.

src/sagemaker/jumpstart/utils.py
Captainia: For proprietary model it returns empty dict. This will gets passed to the `verify_model_region_and_return_specs`. I'm open to raise an exception for proprietary models actually

src/sagemaker/jumpstart/types.py
Captainia: Sorry I'm not following, there's no other classes inheriting this class. I did removed this in other parent classes.

tests/unit/sagemaker/jumpstart/test_utils.py
Captainia: I'm open to it either way, say a model doesn't have any configs in its metadata, we can raise an exception saying the model has empty config, or we treat empty list as the expected result.
For now I'll keep as is, let me know if you feel strongly we should raise an exception.

src/sagemaker/jumpstart/types.py
Captainia: No, it will not apply anything to the spec if there are no configs for this model. We should not fail for models without the alt config metadata.

src/sagemaker/jumpstart/types.py
Captainia: There's a circular dependency to put it in `constants.py`, so I created a enum to store it

makungaj1 (approved):

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        self.benchmark_metrics: Dict[str, JumpStartBenchmarkStat] = benchmark_metrics
+        self.resolved_metadata_config: Optional[Dict[str, Any]] = None
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartMetadataConfig object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+    @property
+    def resolved_config(self) -> None:
```
qiyunz: why is the return type None?

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+                The dictionary of benchmark metrics with name being the key.
+        """
+        self.base_fields = base_fields
+        self.config_components: Dict[str, JumpStartConfigComponent] = config_components
+        self.benchmark_metrics: Dict[str, JumpStartBenchmarkStat] = benchmark_metrics
+        self.resolved_metadata_config: Optional[Dict[str, Any]] = None
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartMetadataConfig object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
```
qiyunz, line 1102: do we need a separate interface like `resolved_config_to_json()`?

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+
+def get_jumpstart_configs(
+    region: str,
+    model_id: str,
+    model_version: str,
+    config_names: Optional[List[str]] = None,
+    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
+    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
+) -> Dict[str, List[JumpStartMetadataConfig]]:
```
qiyunz, line 959: what's the key of the returned dict? If key is the config name, why does each config name map to multiple JumpStartMetadataConfig?

src/sagemaker/jumpstart/types.py
Captainia: Thanks I just fixed this return typing in the latest commit, this should be a `Dict[str, Any]`

src/sagemaker/jumpstart/utils.py
Captainia: The key is config name, I can update this to return only the list of configs. Sorry this typing is confusing, it originally should be `Dict[str, JumpStartMetadataConfig]`.

src/sagemaker/jumpstart/types.py
Captainia: Should not need a separate interface, resolved config is a json object because we override and merge to the base fields.

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        rankings = self.config_rankings.get(ranking_name)
+        for config_name in rankings.rankings:
+            resolved_config = self.configs[config_name].resolved_config
+            if instance_type and instance_type not in getattr(
+                resolved_config, instance_type_attribute
+            ):
+                continue
+            return self.configs[config_name]
+
+        return None
```
whittech1, line 1184: The return type should be `Optional` right?

whittech1 (approved):

src/sagemaker/jumpstart/types.py
Captainia: Yes, I'll address this in my next PR, thanks for catching

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+    """Returns metadata configs for the given model ID and region."""
+    model_specs = verify_model_region_and_return_specs(
```
akozd, line 961: nit: this block of code appears 3 times in `get_jumpstart_configs`, `get_benchmark_stats`, and `get_config_names`

akozd (approved):

src/sagemaker/jumpstart/utils.py
Captainia: Yea I think it's required because we use `verify_model_region_and_return_specs` as the entry point to get specs in all these utils, and each function can be called standalone. But I think we can probably choose a better naming of  `verify_model_region_and_return_specs` and move outside of utils as it's used heavily everywhere.

src/sagemaker/jumpstart/types.py
```diff
@@ -1001,12 +1004,19 @@ def to_json(self) -> Dict[str, Any]:
                             json_obj[att].append(obj.to_json())
                         else:
                             json_obj[att].append(obj)
+                elif isinstance(cur_val, dict):
+                    json_obj[att] = {}
+                    for key, val in cur_val.items():
+                        if issubclass(type(val), JumpStartDataHolderType):
+                            json_obj[att][key] = val.to_json()
+                        else:
+                            json_obj[att][key] = val
```
evakravi, line 1013: is this not defined in some parent class?

src/sagemaker/jumpstart/types.py
```diff
@@ -1001,12 +1004,19 @@ def to_json(self) -> Dict[str, Any]:
                             json_obj[att].append(obj.to_json())
                         else:
                             json_obj[att].append(obj)
+                elif isinstance(cur_val, dict):
+                    json_obj[att] = {}
+                    for key, val in cur_val.items():
+                        if issubclass(type(val), JumpStartDataHolderType):
```
evakravi, line 1010: can we consolidate with this: https://github.com/aws/sagemaker-python-sdk/blob/2f449759b2835eae8552b8d67ef9ff5c55f191e4/src/sagemaker/jumpstart/types.py#L934

src/sagemaker/jumpstart/types.py
```diff
@@ -1028,22 +1038,23 @@ class JumpStartPresetComponent(JumpStartMetadataBaseFields):
 
     __slots__ = slots + JumpStartMetadataBaseFields.__slots__
 
-    def __init__(
+    def __init__(  # pylint: disable=super-init-not-called
```
evakravi, line 1041: why aren't we calling the superclass constructor? This isn't good OOP design IMO, it violates [Liskov Substution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle)

Captainia: I'll update, I left the component to only have its specific fields but not construct all the parent class fields which are optional, but it should work with superclass constructor. Good callout

src/sagemaker/jumpstart/types.py
Captainia: Sounds good, I'll address in my next PR.

src/sagemaker/jumpstart/types.py
```diff
@@ -736,8 +743,65 @@ def _get_regional_property(
...
 
-class JumpStartModelSpecs(JumpStartDataHolderType):
-    """Data class JumpStart model specs."""
+class JumpStartBenchmarkStat(JumpStartDataHolderType):
+    """Data class JumpStart benchmark stats."""
+
+    __slots__ = ["name", "value", "unit"]
+
+    def __init__(self, spec: Dict[str, Any]):
+        """Initializes a JumpStartBenchmarkStat object
```
JGuinegagne, line 752: nit: missing period `.`

src/sagemaker/jumpstart/types.py
```diff
@@ -794,26 +858,26 @@ class JumpStartModelSpecs(JumpStartDataHolderType):
...
-        """Initializes a JumpStartModelSpecs object from its json representation.
+    def __init__(self, fields: Optional[Dict[str, Any]]):
+        """Initializes a JumpStartMetadataFields object.
 
         Args:
-            spec (Dict[str, Any]): Dictionary representation of spec.
+            fields (Dict[str, Any]): Dictionary representation of metadata fields.
         """
-        self.from_json(spec)
+        self.from_json(fields)
```
JGuinegagne, line 867: how does this work, I don't `mypy` should accept this signature given that `from_json` accepts a `Dict[str, Any]` and get passed an `Optional[Dict[str, Any]]`. This could cause null pointer errors.

src/sagemaker/jumpstart/types.py
```diff
@@ -794,26 +858,26 @@ class JumpStartModelSpecs(JumpStartDataHolderType):
...
-        self.version: str = json_obj["version"]
-        self.min_sdk_version: str = json_obj["min_sdk_version"]
+        self.model_id: str = json_obj.get("model_id")
+        self.url: str = json_obj.get("url")
+        self.version: str = json_obj.get("version")
+        self.min_sdk_version: str = json_obj.get("min_sdk_version")
         self.incremental_training_supported: bool = bool(
-            json_obj.get("incremental_training_supported", False)
+            json_obj.get("incremental_training_supported")
         )
```
JGuinegagne, line 881: all these types are de facto `Optional[*]` now?

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        """
+        self.component_name = component_name
+        self.from_json(component)
+
+    def from_json(self, json_obj: Dict[str, Any]) -> None:
+        """Initializes a JumpStartConfigComponent object from its json representation.
+
+        Args:
+            json_obj (Dict[str, Any]):
+                Dictionary representation of the config component.
```
JGuinegagne, line 1061: add a `Raises` section to the docstring please

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        self.resolved_metadata_config: Optional[Dict[str, Any]] = None
+
+    def to_json(self) -> Dict[str, Any]:
+        """Returns json representation of JumpStartMetadataConfig object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+    @property
+    def resolved_config(self) -> Dict[str, Any]:
+        """Returns the final config that is resolved from the list of components.
```
JGuinegagne, line 1107: The components isn't a list, `[...] from the components map.` or `[...] from the dict of components`.

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+
+        Construct the final config by applying the list of configs from list index,
+        and apply to the base default fields in the current model specs.
+        """
+        if self.resolved_metadata_config:
+            return self.resolved_metadata_config
+
+        resolved_config = JumpStartMetadataBaseFields(self.base_fields)
+        for component in self.config_components.values():
```
JGuinegagne, line 1116: how does the override sequence get preserved if you use a dict?

Doesn't it call to be stored as list in order to preserve priority of overidding?

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        self,
+        configs: Optional[Dict[str, JumpStartMetadataConfig]],
+        config_rankings: Optional[Dict[str, JumpStartConfigRanking]],
+        scope: JumpStartScriptScope = JumpStartScriptScope.INFERENCE,
+    ):
+        """Initializes a JumpStartMetadataConfigs object.
+
+        Args:
+            configs (Dict[str, JumpStartMetadataConfig]):
+                List of configs that the current model has.
```
JGuinegagne, line 1142: fix docstring please, this isn't a list.

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        """Returns json representation of JumpStartMetadataConfigs object."""
+        json_obj = {att: getattr(self, att) for att in self.__slots__ if hasattr(self, att)}
+        return json_obj
+
+    def get_top_config_from_ranking(
+        self,
+        ranking_name: str = JumpStartConfigRankingName.DEFAULT,
+        instance_type: Optional[str] = None,
+    ) -> JumpStartMetadataConfig:
+        """Gets the best the config based on config ranking."""
```
JGuinegagne, line 1162: add a `Raises` section please

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        if self.configs and (
+            not self.config_rankings or not self.config_rankings.get(ranking_name)
+        ):
+            raise ValueError("Config exists but missing config ranking.")
```
JGuinegagne, line 1166: optional: how about embedding the `ranking_name` and `model_id` in the error message?

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+
+        if self.scope == JumpStartScriptScope.INFERENCE:
+            instance_type_attribute = "supported_inference_instance_types"
+        elif self.scope == JumpStartScriptScope.TRAINING:
+            instance_type_attribute = "supported_training_instance_types"
+        else:
+            raise ValueError(f"Unknown script scope {self.scope}")
```
JGuinegagne, line 1173: Maintain code consistency as a first priority, but here we could consider `NotImplementedError`.

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+
+        rankings = self.config_rankings.get(ranking_name)
+        for config_name in rankings.rankings:
+            resolved_config = self.configs[config_name].resolved_config
+            if instance_type and instance_type not in getattr(
+                resolved_config, instance_type_attribute
+            ):
```
JGuinegagne, line 1180: sanity-check: does this handle instance family? If not, how do you handle instance families?

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+        "inference_config_components",
+        "inference_config_rankings",
+        "training_configs",
+        "training_config_components",
+        "training_config_rankings",
+    ]
+
+    __slots__ = JumpStartMetadataBaseFields.__slots__ + slots
+
+    def __init__(self, spec: Dict[str, Any]):  # pylint: disable=super-init-not-called
```
JGuinegagne, line 1201: ack please, call the parent constructor in every code paths, otherwise the code is counter-intuitive.

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+            else None
+        )
+        self.inference_config_rankings: Optional[Dict[str, JumpStartConfigRanking]] = (
+            {
+                alias: JumpStartConfigRanking(ranking)
+                for alias, ranking in json_obj["inference_config_rankings"].items()
+            }
+            if json_obj.get("inference_config_rankings")
+            else None
+        )
```
JGuinegagne, line 1233: optional: inlining dict comprehension with ternary operator makes the code a bit hard to read IMHO, consider:

```py
if json_obj.get("inference_config_rankings"):
  self.inference_config_rankings = { ... }
else:
  self.inference_config_rankings = None
```

src/sagemaker/jumpstart/types.py
```diff
@@ -940,10 +1004,340 @@ def to_json(self) -> Dict[str, Any]:
...
+
+    def set_config(
+        self, config_name: str, scope: JumpStartScriptScope = JumpStartScriptScope.INFERENCE
+    ) -> None:
+        """Apply the seleted config and resolve to the current model spec.
+
+        Args:
+            config_name (str): Name of the config.
+            scope (JumpStartScriptScope, optional):
+                Scope of the config. Defaults to JumpStartScriptScope.INFERENCE.
```
JGuinegagne, line 1332: add `Raises` section please

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+
+def get_config_names(
+    region: str,
+    model_id: str,
+    model_version: str,
+    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
+    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
+) -> List[str]:
+    """Returns a list of config names for the given model ID and region."""
```
JGuinegagne, line 893: add `Raises` section please

src/sagemaker/jumpstart/utils.py
JGuinegagne: Then fix the docstring return type.

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+def get_benchmark_stats(
+    region: str,
+    model_id: str,
+    model_version: str,
+    config_names: Optional[List[str]] = None,
+    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
+    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
+) -> Dict[str, List[JumpStartBenchmarkStat]]:
+    """Returns benchmark stats for the given model ID and region."""
```
JGuinegagne, line 922: add a `Raises` section please

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+        scope=scope,
+        model_type=model_type,
+    )
+
+    if scope == enums.JumpStartScriptScope.INFERENCE:
+        metadata_configs = model_specs.inference_configs
+    elif scope == enums.JumpStartScriptScope.TRAINING:
+        metadata_configs = model_specs.training_configs
+    else:
+        raise ValueError(f"Unknown script scope {scope}.")
```
JGuinegagne, line 937: inconsistent error message: `f"Unknown script scope: {scope}"`

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+
+    if not config_names:
+        config_names = metadata_configs.configs.keys() if metadata_configs else []
+
+    benchmark_stats = {}
+    for config_name in config_names:
+        if config_name not in metadata_configs.configs:
+            raise ValueError(f"Unknown config name: '{config_name}'")
```
JGuinegagne, line 945: same comment: `f"Unknown config name: {config_name}"` (no `'`)

src/sagemaker/jumpstart/utils.py
```diff
@@ -878,3 +880,105 @@ def get_region_fallback(
...
+def get_jumpstart_configs(
+    region: str,
+    model_id: str,
+    model_version: str,
+    config_names: Optional[List[str]] = None,
+    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
+    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
+    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
+) -> Dict[str, List[JumpStartMetadataConfig]]:
+    """Returns metadata configs for the given model ID and region."""
```
JGuinegagne, line 960: `Raises` section please

src/sagemaker/jumpstart/types.py
Captainia: I might be missing something, the ordering of dict `key` and `value` should be preserved as the ordering when we inserted them: https://stackoverflow.com/questions/835092/python-dictionary-are-keys-and-values-always-the-same-order. This should be the same as a list

src/sagemaker/jumpstart/types.py
Captainia: Sorry this is confusing typing, the `fields` object is `Dict[str, Any]` but not optional. I think `mypy` is not checking these typings, updated in my next PR

src/sagemaker/jumpstart/types.py
Captainia: At a first iteration we are not considering instance type families, the config is chosen if the `supported_instance_types` is listed in the config with a exact match.

PR #4625, opened 2024-04-26
chore: require config name and instance type in set_deployment_config

Captainia:
*Issue #, if available:*

*Description of changes:*

This change made the `set_deployment_config` interface to require two mandatory input: config_name and instance_type, since these two combination will be deterministic for a config. This change also removes the possibility to unset a config, since now both config name and instance_type is required field.

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/jumpstart/factory/model.py
+++ b/src/sagemaker/jumpstart/factory/model.py
@@ -569,13 +569,18 @@ def _add_config_name_to_kwargs(kwargs: JumpStartModelInitKwargs) -> JumpStartMod
             kwargs.config_name or specs.inference_configs.get_top_config_from_ranking().config_name
         )
 
-        if kwargs.config_name:
-            resolved_config = specs.inference_configs.configs[kwargs.config_name].resolved_config
-            supported_instance_types = resolved_config.get("supported_inference_instance_types", [])
-            if kwargs.instance_type not in supported_instance_types:
-                raise ValueError(
-                    f"Instance type {kwargs.instance_type} is not supported for config {kwargs.config_name}."
-                )
+        if not kwargs.config_name:
+            return kwargs
+
+        if kwargs.config_name not in set(specs.inference_configs.configs.keys()):
+            raise ValueError(f"Config {kwargs.config_name} is not supported for model {kwargs.model_id}.")
+        
+        resolved_config = specs.inference_configs.configs[kwargs.config_name].resolved_config
+        supported_instance_types = resolved_config.get("supported_inference_instance_types", [])
+        if kwargs.instance_type not in supported_instance_types:
+            raise ValueError(
+                f"Instance type {kwargs.instance_type} is not supported for config {kwargs.config_name}."
+            )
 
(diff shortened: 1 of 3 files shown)


qiyunz (approved):

src/sagemaker/jumpstart/model.py
```diff
@@ -429,16 +429,22 @@ def retrieve_example_payload(self) -> JumpStartSerializablePayload:
             sagemaker_session=self.sagemaker_session,
         )
 
-    def set_deployment_config(self, config_name: Optional[str]) -> None:
+    def set_deployment_config(self, config_name: str, instance_type: str) -> None:
```
jiapinw, line 432: do we need validation on checking whether config_name + instance_type combination exists? Or customers can just choose whatever instance types they want?

Captainia: Good point, this instance_type override happens to the entire model, essentially equilvant to re-initialize the model with the instance_type. I will add a logic to check if the instance_type is supported in the config.

Captainia: Updated

src/sagemaker/jumpstart/factory/model.py
```diff
@@ -565,6 +569,14 @@ def _add_config_name_to_kwargs(kwargs: JumpStartModelInitKwargs) -> JumpStartMod
             kwargs.config_name or specs.inference_configs.get_top_config_from_ranking().config_name
         )
 
+        if kwargs.config_name:
```
jiapinw, line 572: Are we initializing config_name as None? if not we should use getattr/hasattr to avoid attribute not found exception

src/sagemaker/jumpstart/factory/model.py
```diff
@@ -565,6 +569,14 @@ def _add_config_name_to_kwargs(kwargs: JumpStartModelInitKwargs) -> JumpStartMod
...
+            resolved_config = specs.inference_configs.configs[kwargs.config_name].resolved_config
```
jiapinw, line 573: is it ever possible to have key error here?

jiapinw (approved):

src/sagemaker/jumpstart/factory/model.py
Captainia: We are initializing config_name as None similar to other fields across the kwargs class, this should be fine

src/sagemaker/jumpstart/factory/model.py
Captainia: I can add a more indicative error message when the config name is invalid (not in the keys)

jiapinw: I'm ok with this now since it's private beta, but let's make sure these are accounted for in GA

Captainia: This is already included in the `types.py`, added a test case though

zhaoqizqwang (approved):

PR #4729, opened 2024-06-12
change: Enable telemetry logging for Remote function

knikure:
*Issue #, if available:*

*Description of changes:*
Enable telemetry logging for Remote function 
* We need to retrieve the `sagemaker_session` from the [kwargs](https://github.com/knikure/sagemaker-python-sdk/blob/382fde11a534b9848176994c16b15c1872acc92a/src/sagemaker/remote_function/client.py#L80) of remote function since it is not a class function.

*Testing done:*
Tested on the Example notebook [here](https://github.com/aws/amazon-sagemaker-examples/blob/main/sagemaker-remote-function/quick_start/quick_start.ipynb).

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/remote_function/client.py
+++ b/src/sagemaker/remote_function/client.py
@@ -40,6 +40,8 @@ from sagemaker.remote_function import logging_config
 from sagemaker.utils import name_from_base, base_from_name
 from sagemaker.remote_function.spark_config import SparkConfig
 from sagemaker.remote_function.custom_file_filter import CustomFileFilter
+from sagemaker.telemetry.telemetry_logging import _telemetry_emitter
+from sagemaker.telemetry.constants import Feature
 
 _API_CALL_LIMIT = {
     "SubmittingIntervalInSecs": 1,
@@ -57,6 +59,7 @@ _FINISHED = "FINISHED"
 logger = logging_config.get_logger()
 
 
+@_telemetry_emitter(feature=Feature.REMOTE_FUNCTION, func_name="remote_function.remote")
 def remote(
     _func=None,
     *,
--- a/src/sagemaker/telemetry/telemetry_logging.py
+++ b/src/sagemaker/telemetry/telemetry_logging.py
@@ -17,7 +17,10 @@ import platform
 import sys
 from time import perf_counter
 from typing import List
+import functools
+import requests
 
+from sagemaker.session import Session
 from sagemaker.utils import resolve_value_from_config
 from sagemaker.config.config_schema import TELEMETRY_OPT_OUT_PATH
 from sagemaker.telemetry.constants import (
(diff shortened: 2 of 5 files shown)


src/sagemaker/remote_function/client.py
```diff
@@ -57,6 +59,7 @@
 logger = logging_config.get_logger()
 
 
+@_telemetry_emitter(feature=Feature.REMOTE_FUNCTION, func_name="remote_function.remote")
```
qidewenwhen, line 62: The full path seems to be sagemaker.remote_function.client.remote, given the import. Wondering if we should use the full function name here, e.g. `remote_function.client.remote`.

If it does not matter for the telemetry, then no need to update. 

```
from sagemaker.remote_function.client import (
    remote,
...
)
```

src/sagemaker/telemetry/telemetry_logging.py
```diff
@@ -59,86 +64,103 @@ def _telemetry_emitter(feature: str, func_name: str):
...
+                            str(e),
+                            e.__class__.__name__,
+                            extra,
+                        )
+                    caught_ex = e
+                finally:
+                    if caught_ex:
+                        raise caught_ex
+                    return response  # pylint: disable=W0150
+            else:
```
qidewenwhen, line 151: Nit: we can move the else branch on top to reduce to clean the code a little bit. For example:
```
if not sagemaker:
   logger.debug(...)
    return ...

logger.debug("sagemaker_session found, preparing to emit telemetry...")
...

File: tests/unit/sagemaker/remote_function/test_client.py [hunk 3]
```diff
@@ -1498,7 +1499,6 @@ def test_consistency_between_remote_and_step_decorator():
     from sagemaker.workflow.function_step import step
 
     remote_args_to_ignore = [
-        "_remote",
```
qidewenwhen, line 1501: This is removed because we replace `remote.__code__.co_varnames` with `inspect.signature(remote).parameters.keys()`, and the latter retrieves the function parameters precisely?

qidewenwhen (approved):

src/sagemaker/telemetry/telemetry_logging.py
knikure: ack, Will do this in the follow-up PR for telemetry.

tests/unit/sagemaker/remote_function/test_client.py
knikure: Yes, right. The change was also needed since now that we are decorating `remote()` decorator with `_telemetry_emitter()` decorator, the `func.__code__.co_varnames` doesn't seem to return correct list of args. So I fixed the unit-test.

PR #4813, opened 2024-07-30
fix: ModelReference deployment for Alt Configs models

malav-shastri:
*Issue #, if available:*
(Please check the internal ticket for the latest updates)

*Description of changes:*
We have identified that currently Model Ref deployment in CuratedHub is broken which looks like in general incompatibility of new QS code with CuratedHub Code path + some Schema parsing issues. 

- Parser issue: where is_hub_content flag is not set for the alt config parsing, resulting in failure when parsing the Alt config schema

- Parser issue: inference_configs_dict is not being fetched correctly from the json object which results in AttributeError: 'JumpStartScriptScope' object has no attribute 'get'

- hub_arn has not been passed for get_jumpstart_configs() resulting in empty hub_arn and code path for curatedHub is not getting executed.

- when get_top_config_from_ranking() returns None there's no handling of None response in the _add_config_name_to_init_kwargs() function which results in key error.

*Testing done:*
- Notebook testing, tried to deploy the ModelReference using the local changes

## Merge Checklist

#### General

#### Tests


src/sagemaker/jumpstart/types.py
```diff
@@ -1732,32 +1732,15 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
-                        if config and config.get("component_names")
-                        else None
-                    ),
-                )
-                for alias, config in json_obj["inference_configs"].items()
-            }
-            if json_obj.get("inference_configs")
-            else None
-        )
+        inference_configs_dict: Optional[Dict[str, JumpStartMetadataConfig]] = json_obj[
```
Captainia, line 1741: I'm not sure if this works, now that the `inference_configs_dict` becomes a json object, how would we construct the `JumpStartMetadataConfig` class without parsing the json?

bencrabtree: +1, also curious if we should do `.get` here also to mitigate future errors

src/sagemaker/jumpstart/types.py
```diff
@@ -1732,32 +1736,50 @@ def from_json(self, json_obj: Dict[str, Any]) -> None:
...
-                        else None
-                    ),
-                )
-                for alias, config in json_obj["inference_configs"].items()
-            }
-            if json_obj.get("inference_configs")
-            else None
-        )
+
+        if self._is_hub_content:
```
Captainia, line 1746: non-blocking: since we are here, it would be good if we could update line 1792 for training configs, but doesn't have to be in this PR.

Captainia (approved):

src/sagemaker/jumpstart/types.py
malav-shastri: thanks for the callout, I added those changed as well

malav-shastri:
> Please add unittest coverage that in types.py, we should catch issues like this in the future

sure I'll have a followup PR for that right after this one, I believe we should be unblocking the customers here first

bencrabtree (approved):

nileshvd (approved):
I dont have context about this change. But since Captainia@ has already approved the change, and since this is for an active sev2 I am approving this change.

nargokul (approved):

PR #4992, opened 2025-01-13
feat: use jumpstart deployment config image as default optimization image

gwang111:
*Issue #, if available:*

*Description of changes:*
- Fetch the image uri default for `.optimize()` from the JumpStart deployment config in vLLM usecases
- Update unit and integ tests to verify that the default is being propagated to `.optimize()`

*Testing done:*
Set
```
"ModelShardingConfig": {
    "Image": "dkr:jumpstart-deployment-config-image-uri"
}
or 
"ModelQuantizationConfig": {
    "Image": "dkr:jumpstart-deployment-config-image-uri"
}
```
In the following instances:
1. When the customer wants to quantize their model
2. When the customer wants to optimize with model sharding

Skip:
1. For speculative decoding, `.optimize()` does not need to be called. No defaulting
2. Skip configurations not supported by vLLM
3. Skip for compilation as vLLM does not support compilation

## Merge Checklist

#### General

#### Tests


src/sagemaker/serve/builder/jumpstart_builder.py
```diff
@@ -986,3 +986,24 @@ def _get_neuron_model_env_vars(
...
+        """Defaults the optimization image to the JumpStart deployment config default
+
+        Args:
+            create_optimization_job_args (Dict[str, Any]): create optimization job request
+
+        Returns:
+            Dict[str, Any]: create optimization job request with image uri default
+        """
+        for optimization_config in create_optimization_job_args.get("OptimizationConfigs"):
+            model_compilation_config = optimization_config.get("ModelCompilationConfig")
```
Lokiiiiii, line 1002: This logic should only be applied for vLLM optimizations - https://github.com/aws/sagemaker-python-sdk/blob/a58654ecbd5997aa59e4fcd8389715fe04c71f5d/src/sagemaker/serve/validations/optimization.py#L29. 

since JumpStart only provides the vLLM container in it's metadata, it would not work for Neuron or TRT-LLM

Lokiiiiii: It is also not specific to ModelCompilationConfig. It should be applied to optimizations supported on vLLM like Quantization and ModelSharding.

Lokiiiiii (changes requested):

src/sagemaker/serve/builder/jumpstart_builder.py
gwang111: Updated to only default for vLLM usecases.

To my understanding:
- Compilation/Compilation + Quantization is no longer valid as vLLM does not support any form of compilation

cj-zhang (approved):

src/sagemaker/serve/builder/jumpstart_builder.py
```diff
@@ -986,3 +992,103 @@ def _get_neuron_model_env_vars(
...
+        """
+        dlc_name, _ = image.split(":")
+        major_version_number, _, _ = self._parse_lmi_version(image)
+
+        if int(major_version_number) < 13:
+            minimum_version_default = f"{dlc_name}:0.31.0-lmi13.0.0-cu124"
+            return minimum_version_default
+        return image
+
+    def _compare_lmi_versions(self, version: str, version_to_compare: str) -> bool:
```
Lokiiiiii, line 1052: Can we make this more readable ? If the name can reflect the nature of the comparison that would be better. Something like - `get_latest_lmi_version_from_list`

src/sagemaker/serve/builder/jumpstart_builder.py
```diff
@@ -986,3 +992,103 @@ def _get_neuron_model_env_vars(
...
+            image (str): JumpStart provided default image
+
+        Returns:
+            str: minimum working image version
+        """
+        dlc_name, _ = image.split(":")
+        major_version_number, _, _ = self._parse_lmi_version(image)
+
+        if int(major_version_number) < 13:
+            minimum_version_default = f"{dlc_name}:0.31.0-lmi13.0.0-cu124"
```
Lokiiiiii, line 1048: Define as a constant please outside the function scope please. This is something we will need to update in the future.

src/sagemaker/serve/builder/jumpstart_builder.py
```diff
@@ -986,3 +992,103 @@ def _get_neuron_model_env_vars(
...
+                model_sharding_config = optimization_config.get("ModelShardingConfig")
+                provided_image = model_sharding_config.get("Image")
+                if provided_image and self._compare_lmi_versions(default_image, provided_image):
+                    default_image = provided_image
+
+        # default to latest vLLM version
+        for optimization_config in create_optimization_job_args.get("OptimizationConfigs"):
+            if optimization_config.get("ModelQuantizationConfig") is not None:
+                optimization_config.get("ModelQuantizationConfig")["Image"] = default_image
+            elif optimization_config.get("ModelShardingConfig") is not None:
```
Lokiiiiii, line 1026: In the future, we will allow stacking quantization + sharding. I think we should convert this from `elif` -> `if` to keep the code scalable.

tests/unit/sagemaker/serve/builder/test_js_builder.py
```diff
@@ -75,7 +75,7 @@
     "-inference:2.1.1-tgi1.4.0-gpu-py310-cu121-ubuntu20.04"
 )
 mock_djl_image_uri = (
-    "123456789712.dkr.ecr.us-west-2.amazonaws.com/djl-inference:0.24.0-neuronx-sdk2.14.1"
+    "763104351884.dkr.ecr.us-west-2.amazonaws.com/djl-inference:0.31.0-lmi13.0.0-cu124"
```
Lokiiiiii, line 78: We should add a test to ensure our original intent is satisfied - 

if pysdk_model.init_kwargs has an image more recent than the hardcoded default, then that should be the image used for optimization.

Lokiiiiii (changes requested):

src/sagemaker/serve/builder/jumpstart_builder.py
```diff
@@ -986,3 +993,105 @@ def _get_neuron_model_env_vars(
...
+        # find the latest vLLM image version
+        for optimization_config in create_optimization_job_args.get("OptimizationConfigs"):
+            if optimization_config.get("ModelQuantizationConfig"):
+                model_quantization_config = optimization_config.get("ModelQuantizationConfig")
+                provided_image = model_quantization_config.get("Image")
+                if provided_image and self._get_latest_lmi_version_from_list(
+                    default_image, provided_image
+                ):
+                    default_image = provided_image
+            if optimization_config.get("ModelShardingConfig"):
```
Lokiiiiii, line 1019: For the future case where both QuantizationConfig and ShardingConfig can be provided at the same time, we should ensure we use the larger of default, user input in QuantizationConfig and user input in ShardingConfig. Right now based on order of execution ShardingConfig.Image will always take precedence over QuantizationConfig.Image

Lokiiiiii (changes requested):

src/sagemaker/serve/builder/jumpstart_builder.py
gwang111: I see.

Maybe I can clean the logic up a bit here but default_image should always be the largest image version. Regardless of the order they're provided in.

Let sharding image version be v21.0.0, quantization image version be v20.0.0

1. default_image starts at 13.0.0
2. quantization v20.0.0 > v13.0.0 (default_image) so now default_image = v20.0.0
3. sharding v21.0.0 > v20.0.0 (default_image) so now default_image = v21.0.0

So now all images will be set to v21.0.0

Lokiiiiii (approved):

gwang111:
Integ tests r failing due to `botocore.errorfactory.ResourceLimitExceeded` exceptions

benieric (approved):

Issue #5099, opened 2025-03-24
Job Name length limit


EnyMan:
**Describe the bug**
When using the PipelineDefinitionConfig i get the error 
```
ClientError: An error occurred (ValidationException) when calling the CreatePipeline operation: Unable to
parse pipeline definition. Model Validation failed: string 
'my-custom-name-SHA256-hash' with 
length=91 cannot be greater than max=63.0 defined for ProcessingJobName.
```
The issue here is that the hash itself is 64 characters. Also the function that generates the `job_name` only limits the length of the generated name to 1024 characters.

**To reproduce**
Any basic pipeline with `PipelineDefinitionConfig(use_custom_job_prefix=True)`

```Python
from sagemaker import ScriptProcessor
from sagemaker.workflow.pipeline import Pipeline
from sagemaker.workflow.pipeline_definition_config import PipelineDefinitionConfig
from sagemaker.workflow.steps import ProcessingStep

ROLE_ARN = "ADD_YOUR_ROLE_HERE"
IMAGE_ARN = "ADD_YOUR_IMAGE_HERE"

definition_config = PipelineDefinitionConfig(use_custom_job_prefix=True)
steps = [
    ProcessingStep(
        name=f"my-first-step",
        display_name="My First Step",
        processor=ScriptProcessor(
            command=["python3"],
            role=ROLE_ARN,
            instance_count=1,
            image_uri=IMAGE_ARN,
            instance_type="ml.m5.xlarge",
        ),
        code=f"example.py",
    ),
]
pipeline = Pipeline(
    name="test-pipeline",
    steps=steps,
    pipeline_definition_config=definition_config,
)

pipeline.upsert(
    role_arn=ROLE_ARN,
)
```

**Expected behavior**
The name should be truncated, so it fits in the 63 character limit

**System information**
A description of your system. Please provide:
- **SageMaker Python SDK version**: 2.242.0
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: Any
- **Framework version**: Any
- **Python version**: 3.12.9
- **CPU or GPU**: CPU
- **Custom Docker image (Y/N)**: Any

lifo9:
I can confirm the issue.

deneriz-veridas:
[...]

PR #5118, opened 2025-04-03
fix: remove historical job_name caching which causes long job name

jkasiraj:
*Issue #, if available:* #5099 

*Description of changes:* 
* Remove historical job_name caching which causes job name to exceed 63 characters with the addition of the 64 char sha256 hash
* Both TrainingJobName and ProcessingJobName have character limits of 63:
  * https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html#sagemaker-CreateProcessingJob-request-ProcessingJobName
  * https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html#sagemaker-CreateTrainingJob-request-TrainingJobName
* local artifact hash is now used instead of timestamp in the s3 path to support caching usecases

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/sagemaker/workflow/steps.py
+++ b/src/sagemaker/workflow/steps.py
@@ -464,6 +464,7 @@ class TrainingStep(ConfigurableRetryStep):
         self.step_args = step_args
         self.estimator = estimator
         self.inputs = inputs
+        self.job_name = None
 
         self._properties = Properties(
             step_name=name, step=self, shape_name="DescribeTrainingJobResponse"
--- a/src/sagemaker/workflow/steps.py
+++ b/src/sagemaker/workflow/steps.py
@@ -493,19 +493,6 @@ class TrainingStep(ConfigurableRetryStep):
                 DeprecationWarning,
             )
 
-        self.job_name = None
-        if estimator and (estimator.source_dir or estimator.entry_point):
-            # By default, `Estimator` will upload the local code to an S3 path
-            # containing a timestamp. This causes cache misses whenever a
-            # pipeline is updated, even if the underlying script hasn't changed.
-            # To avoid this, hash the contents of the training script and include it
-            # in the `job_name` passed to the `Estimator`, which will be used
-            # instead of the timestamped path.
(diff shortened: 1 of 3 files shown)


src/sagemaker/workflow/steps.py
```diff
@@ -493,19 +493,6 @@ def __init__(
                 DeprecationWarning,
             )
 
-        self.job_name = None
```
benieric, line 496: Can we generate the name using the name of the image? That is what other interfaces do. Does removing these code blocks break anything? Maybe need a closer look but is job_name = None we may need to generate a name some other way

benieric: https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/pipeline.py#L244
https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/modules/train/model_trainer.py#L452

jkasiraj: Sure I think we can set it to `self.name`, since the removed code was setting it to `f"{self.name}-{code_hash}"[:1024]`

I believe right now the behavior is that if an estimator is set in a particular way (logic from lines 497-506) it leads to a job name that's too long + errors or otherwise it defaults to None, which is why I considered leaving it as None

I'm a bit hesitant to add new behavior like trying to parse the image (which technically isn't required by the api)
* https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html

jkasiraj: Actually, I looked into it a little further and the job name will eventually get generated [here](https://github.com/aws/sagemaker-python-sdk/blob/228310246557dd36e2b439b7e11a10344faf2f8b/src/sagemaker/estimator.py#L913) if it doesn't exist so I don't think it needs to be set in this function

benieric: Gotchu sounds good 👍

benieric (dismissed):

benieric (approved):

PR #5146, opened 2025-04-25
Add Owner ID check for bucket with path when prefix is provided

nargokul:
**Description**

Previously we called the head_bucket call to ensure the owner ID check, but this doesnt take into consideration cases where the s3 path is provided through the prefix.

This change makes sure that director level permissions are supported.

**Testing Done**
Tested through unit tests, integ tests and manual testing through the installation file.

#### General

#### Tests


src/sagemaker/session.py
```diff
@@ -682,7 +688,14 @@ def general_bucket_check_if_user_has_permission(
             bucket_creation_date_none (bool):Indicating whether S3 bucket already exists or not
         """
         try:
-            s3.meta.client.head_bucket(Bucket=bucket_name)
+            if self.default_bucket_prefix:
+                s3.meta.client.list_objects_v2(
+                    Bucket=bucket_name,
+                    Prefix=self.default_bucket_prefix,
+                    ExpectedBucketOwner=expected_bucket_owner_id
```
benieric, line 695: I believe this check is a bit different in this method. Looks like it just checks for general permissions rather than checking for `ExpectedBucketOwner`. For this case, probably need to remove the `ExpectedBucketOwner=expected_bucket_owner_id` to match the behavior of the second block which just does a regular head bucket check - `s3.meta.client.head_bucket(Bucket=bucket_name)`

nargokul: good callout. Updated the PR.

src/sagemaker/session.py
```diff
@@ -668,7 +678,7 @@ def expected_bucket_owner_id_bucket_check(self, bucket_name, s3, expected_bucket
                 raise
 
     def general_bucket_check_if_user_has_permission(
-        self, bucket_name, s3, bucket, region, bucket_creation_date_none
+        self, bucket_name, s3, bucket, region, bucket_creation_date_none, expected_bucket_owner_id
```
benieric, line 681: nit: this extra parameter is unused

benieric (dismissed):

benieric (approved):

PR #5228, opened 2025-07-01
HF changes for TGI and TEI

pagezyhf:
https://github.com/aws/deep-learning-containers/releases/tag/v2.0-hf-tgi-3.3.4-pt-2.7.0-inf-gpu-py311 for TGI images.
and 
https://github.com/awslabs/llm-hosting-container/blob/main/releases.json for latest TEI release


pagezyhf:
@varunmoris @malav-shastri I think like this it should accomodate the change in the tei tagging methodology

PR #5396, opened 2025-12-09
Test pull, please ignore

kellertk:


Mattral (commented):
Nice try but I think this PR would benefit from a clearer description of the changes and their motivation, or at least trying to find minor issues

PR #5397, opened 2025-12-09
Update PR check workflow file

zhaoqizqwang:
*Description of changes:*

Github changes the workflow behavior yesterday. Now default branch controls all the workflows in other branchs.


zhaoqizqwang:
> Is the pull request target okay with master*? I think that should work, but if we have a chance to change it to make it a bit more specific that might be nice!

I will update it to either master or master-v2. Now it's still in testing stage

PR #5398, opened 2025-12-10
Fix pr check action

zhaoqizqwang:
*Issue #, if available:*

*Description of changes:*


zhaoqizqwang:
> Just confirming we don't need any of the other workflow files anymore? I'm okay with removing but just want to double check

yes other workflow files like ci-health will only run if they are in master branch

PR #5418, opened 2025-12-16
Add input validation and resource management improvements V3

aviruthen:
*Issue #, if available:*

*Description of changes:*
Adds additional input validation and resource management improvements to prevent accidental customer misuse.

### DIFF
--- a/sagemaker-core/src/sagemaker/__init__.py
+++ b/sagemaker-core/src/sagemaker/__init__.py
@@ -1,0 +1,2 @@
+"""Namespace package for SageMaker."""
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
--- a/sagemaker-serve/src/sagemaker/__init__.py
+++ b/sagemaker-serve/src/sagemaker/__init__.py
@@ -1,0 +1,2 @@
+"""Namespace package for SageMaker."""
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
--- a/sagemaker-train/README.rst
+++ b/sagemaker-train/README.rst
@@ -47,7 +47,7 @@ Table of Contents
 
 
 Installing the SageMaker Python SDK Train
------------------------------------
+-----------------------------------------
 
 You can install from source by cloning this repository and running a pip install command in the root directory of the repository:
 
--- a/sagemaker-train/src/sagemaker/__init__.py
+++ b/sagemaker-train/src/sagemaker/__init__.py
@@ -1,0 +1,2 @@
+"""Namespace package for SageMaker."""
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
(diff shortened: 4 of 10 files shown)


sagemaker-core/src/sagemaker/core/common_utils.py
```diff
@@ -607,6 +607,95 @@ def _save_model(repacked_model_uri, tmp_model_path, sagemaker_session, kms_key):
...
+    Raises:
+        ValueError: If the path is not allowed.
+    """
+    if not dependency:
+        return
+
+    abs_dependency = abspath(dependency)
+
+    # Blocklist of sensitive directories that should not be accessible
+    sensitive_paths = [
```
jam-jee: Is this `sensitive_paths` duplicate of _validate_source_directory's sensitive path?  if so, can we refactor to use a common source of `sensitive_paths`

I can also see another list `RESTRICTED_PATH` defined in sagemaker-core/src/sagemaker/core/local/data.py, that seem duplicate. 
If these are duplicate , can we fix redundancy and use a common list of sensitive/restricted paths ?

sagemaker-core/src/sagemaker/core/common_utils.py
```diff
@@ -607,6 +607,95 @@ def _save_model(repacked_model_uri, tmp_model_path, sagemaker_session, kms_key):
...
+    for sensitive_path in sensitive_paths:
+        if abs_source.startswith(sensitive_path):
+            raise ValueError(
+                f"source_directory cannot access sensitive system paths. "
+                f"Got: {source_directory} (resolved to {abs_source})"
+            )
+
+    # Check for symlinks to prevent symlink-based escapes
+    if os.path.islink(abs_source):
+        raise ValueError(f"source_directory cannot be a symlink: {source_directory}")
```
jam-jee: just curious: Are symlinks not supported for source path ?

sagemaker-core/src/sagemaker/core/iterators.py
```diff
@@ -114,6 +114,9 @@ def __next__(self):
 class LineIterator(BaseIterator):
     """A helper class for parsing the byte Event Stream input to provide Line iteration."""
 
+    # Maximum buffer size to prevent unbounded memory consumption (10 MB)
+    MAX_BUFFER_SIZE = 10 * 1024 * 1024
```
jam-jee: Move this to some constants file.

sagemaker-core/src/sagemaker/core/common_utils.py
aviruthen: They should be supported... and this is a good point! Now that I think about it, we don't have to completely prevent sym-links, we just need to resolve the sym-link with os.path.realpath() before validation. This prevents the bug with race conditions. I will make that change

sagemaker-core/src/sagemaker/core/iterators.py
aviruthen: Good point! Adding it to common_utils.py

jam-jee (approved):

PR #5422, opened 2025-12-16
Fix _run_shell_cmd() to use list input

pintaoz-aws:
*Issue #, if available:*

*Description of changes:*
Fix _run_shell_cmd() to use list input


rsareddy0329 (approved):
* Can we also add the test details given in the ticket and the solution is successful.

Issue #5391, opened 2025-12-09
ModelBuilder cannot call `register()` method Without Setting `framework` and `framework_version` Attributes


plockxng:
## Issue

Hello SageMaker Python SDK team,

It seems like based on this [notebook](https://github.com/aws/sagemaker-python-sdk/blob/master/v3-examples/ml-ops-examples/v3-model-registry-example/v3-model-registry-example.ipynb), I can register a model directly without having to "build" it:

<img width="916" height="569" alt="Image" src="https://github.com/user-attachments/assets/2a92ad85-77dc-442e-b786-2a831fb2eac3" />

What I'm observing is that it's missing a `framework` and `framework_version` attribute:

error 1: `AttributeError: 'ModelBuilder' object has no attribute 'framework'`
error 2: `AttributeError: 'ModelBuilder' object has no attribute 'framework_version'`

This can be circumvented by setting these explicitly like so before calling the `model_builder.register()` method, but is there any way to remove this dependency, it seems unnecessary:

```python
model_builder.framework = None
model_builder.framework_version = None
```

## Resolution

Don't want to set these attributes explicitly, and most likely shouldn't be set by an end user.

mohamedzeidan2021:
Fixed. `ModelBuilder` now defaults `framework` and `framework_version` to `None` at construction (in `__post_init__`), so `register()` can be called directly without building first — no more `AttributeError`, and the manual `model_builder.framework = None` workaround is no longer needed. Landed via #5558 and present in the current SDK. Closing as resolved.

PR #5424, opened 2025-12-17
Initialize framework and framework version in post_init

aviruthen:
*Issue #, if available:*
https://github.com/aws/sagemaker-python-sdk/issues/5391

*Description of changes:*
Fix a bug to initialize framework and framework version during post_init to None by default. Update model registry notebook.


sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -489,6 +489,11 @@ def _initialize_defaults(self) -> None:
         self.container_config = "host"
         self.inference_recommender_job_results = None
         self.container_log_level = logging.INFO
+        
+        if not hasattr(self, 'framework') or self.framework is None:
```
zhaoqizqwang: Do we need the `or self.framework is None` here?

aviruthen: Oh yes I guess not... it doesn't change the logic but for simplicity I can make this change!

rsareddy0329 (approved):

mollyheamazon (approved):

PR #5437, opened 2025-12-23
fix: Update telemetry constants to track v2 feature seperately

rsareddy0329:
*Issue #, if available:*

*Description of changes:*
* Update telemetry constants to track v2 features seperately

### DIFF
--- a/src/sagemaker/telemetry/telemetry_logging.py
+++ b/src/sagemaker/telemetry/telemetry_logging.py
@@ -50,15 +50,14 @@ TELEMETRY_OPT_OUT_MESSAGING = (
 )
 
 FEATURE_TO_CODE = {
-    str(Feature.SDK_DEFAULTS): 1,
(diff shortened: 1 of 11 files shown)


src/sagemaker/telemetry/telemetry_logging.py
aviruthen, line 1: Would entries be inconsistent for V2 with this setup? For new users this makes sense, but for users that are already using V2 and don't pip install again, they will emit telemetry for these features at their original codes (1-9) and only new users will emit telemetry in the range 11-17. Should we consider setting the V3 telemetry codes to be 11-17 and keeping V2 in the range of 1-9? I don't feel strongly either way since there are trade-offs with either approach

aviruthen (approved):

aviruthen (approved):

jam-jee (approved):

PR #5438, opened 2025-12-23
fix: Update telemetry constants to include MLOPS as well

rsareddy0329:
*Issue #, if available:*

*Description of changes:*
* fix: Update telemetry constants to include MLOPS as well to have parity with v2


sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
aviruthen, line 1: Do you want MLOPS to also be equal to 8 like model customization? Should this be 9?

aviruthen (approved):

PR #5447, opened 2025-12-31
feat: Add Lake Formation governance support for FeatureGroup

BassemHalim:
Enabling lakeformation governance for a Feature Group following this [blog](https://aws.amazon.com/blogs/machine-learning/control-access-to-amazon-sagemaker-feature-store-offline-using-aws-lake-formation/) requires multiple steps. This PR is to simplify this process.

*Description of changes:*
- Add `enable_lake_formation()` method to enable lf for an existing Feature Group
- Add `enable_lake_formation` flag to the `.create()` method to enable lf for a feature group being created.
- Add helper methods for all the phases required to enable lf:
 1. register the s3 location as a Lakeformation managed resource
 2. grant the offline role permissions to write to the glue table
 3. revoke IAMAllowedPrincipals

### DIFF
--- a/sagemaker-core/src/sagemaker/core/resources.py
+++ b/sagemaker-core/src/sagemaker/core/resources.py
@@ -11818,6 +11818,7 @@ class EndpointConfig(Base):
             list_method_kwargs=operation_input_args,
         )
 
+
 '''
 class EndpointConfigInternal(Base):
     """
@@ -12510,6 +12511,7 @@ class EvaluationJob(Base):
         )
 '''
 
+
 class Experiment(Base):
     """
     Class representing resource Experiment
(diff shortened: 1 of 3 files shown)


sagemaker-core/src/sagemaker/core/resources.py
```diff
@@ -13036,6 +13036,335 @@ def get_name(self) -> str:
...
+        Returns:
+            Dict with status of each Lake Formation operation:
+            - s3_registration: bool
+            - iam_principal_revoked: bool
+            - permissions_granted: bool
+
+        Raises:
+            ValueError: If the Feature Group has no offline store configured,
+                if role_arn is not set on the Feature Group, or if use_service_linked_role
+                is False but registration_role_arn is not provided.
```
romiik: This check seems to be missing

sagemaker-core/src/sagemaker/core/resources.py
```diff
@@ -13036,6 +13036,335 @@ def get_name(self) -> str:
...
+        use_service_linked_role: bool = True,
+        registration_role_arn: Optional[str] = None,
+    ) -> dict:
+        """
+        Enable Lake Formation governance for this Feature Group's offline store.
+
+        This method:
+        1. Registers the offline store S3 location with Lake Formation
+        2. Grants the execution role permissions on the Glue table
+        3. Revokes IAMAllowedPrincipal permissions from the Glue table
```
romiik: What about the s3 deny policy?

sagemaker-core/src/sagemaker/core/resources.py
BassemHalim: missed that thanks

PR #5459, opened 2026-01-06
fix: Remove tags from ProcessingJob creation in Processor

nargokul:
The ProcessingJob resource doesn't accept tags parameter during initialization. Remove tags from the transformed dict before passing to ProcessingJob constructor to prevent errors.

The tags are passed on to the actual create_processing_job call in the submit() function . 

https://tiny.amazon.com/1h1h2lk2x


jam-jee (approved):

aviruthen (approved):
I'm good with making this change. I just want to note that the ProcessingJob object won't have a tags attribute even though the user can set it via the SDK and can view their tag on console. Perhaps this is fine since I'm not sure what the use case is for customers retrieving tags from a ProcessingJob python object directly, but just wanted to point that out here!

PR #5500, opened 2026-01-21
Fix: Don't apply default experiment config for pipelines in non-Eurek…

wutimot:
…a GA regions

*Issue #, if available:*
https://t.corp.amazon.com/V1777657621/communication

*Description of changes:*
Only add the default experiment config if the region is GA with Eureka, using a static list since Eureka isn't being built anymore

Ran Unit Tests locally and passed

### DIFF
--- a/sagemaker-mlops/src/sagemaker/mlops/workflow/_utils.py
+++ b/sagemaker-mlops/src/sagemaker/mlops/workflow/_utils.py
@@ -58,6 +58,36 @@ python _repack_model.py \
 --source_dir "${var_source_dir}"
 """
 
+# Static list of regions where Experiments (Eureka) is Generally Available.
+# Note: Experiments is not expanding to new regions, so this list is static.
+EUREKA_GA_REGIONS = frozenset([
+    "us-east-1",       # iad (N. Virginia)
+    "us-east-2",       # cmh (Ohio)
+    "us-west-1",       # sfo (N. California)
+    "us-west-2",       # pdx (Oregon)
+    "ca-central-1",    # yul (Montreal)
+    "eu-west-1",       # dub (Dublin)
+    "eu-west-2",       # lhr (London)
+    "eu-west-3",       # cdg (Paris)
+    "eu-central-1",    # fra (Frankfurt)
+    "eu-north-1",      # arn (Stockholm)
+    "eu-south-1",      # mxp (Milan)
+    "eu-south-2",      # zaz (Spain)
+    "ap-northeast-1",  # nrt (Tokyo)
+    "ap-northeast-2",  # icn (Seoul)
+    "ap-northeast-3",  # kix (Osaka)
+    "ap-southeast-1",  # sin (Singapore)
+    "ap-southeast-2",  # syd (Sydney)
+    "ap-southeast-3",  # cgk (Jakarta)
+    "ap-south-1",      # bom (Mumbai)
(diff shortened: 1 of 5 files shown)


sagemaker-mlops/src/sagemaker/mlops/workflow/_utils.py
```diff
@@ -58,6 +58,38 @@
 --source_dir "${var_source_dir}"
 """
 
+# ======================= ADD THIS SECTION =======================
```
jkasiraj: nit: remove this line?

sagemaker-mlops/src/sagemaker/mlops/workflow/pipeline.py
```diff
@@ -146,6 +149,15 @@ def __init__(
         self._event_bridge_scheduler_helper = EventBridgeSchedulerHelper(
             self.sagemaker_session.boto_session.client("scheduler"),
         )
+        # Apply default experiment config only in Eureka GA regions
+        if pipeline_experiment_config is _DEFAULT_EXPERIMENT_CFG:
```
jkasiraj: This is a little confusing since `is` should be doing an in memory comparison. Can we set the default to None in the constructor? Then this check will become is not None and it will be populated with the default if appropriate

bmouryakr: Defaulting to None in the constructor and checking here could be an issue if the value is set to None explicitly while instantiating the pipeline. Instead we can leave this as before and add a method to return either None or 
```
PipelineExperimentConfig(
    ExecutionVariables.PIPELINE_NAME, ExecutionVariables.PIPELINE_EXECUTION_ID
)
```
based on the region and call that to set the value for `_DEFAULT_EXPERIMENT_CFG`

wutimot: Yea i was running into issues where the code couldn't tell between default (no parameter passed) and None (explicitly passed None)

jkasiraj: Ah makes sense, I like the method check idea

Alternatively I prefer

```
if pipeline_experiment_config is not None:
  region = self.sagemaker_session.boto_region_name
  if region is not in EUREKA_GA_REGIONS:
    self.pipeline_experiment_config = None
  else
    self.pipeline_experiment_config = pipeline_experiment_config
```

wutimot: this method would override the user's custom config right? I thought we wanted to fail it on the service level instead of prematurely setting it to None

bmouryakr: > Defaulting to None in the constructor and checking here could be an issue if the value is set to None explicitly while instantiating the pipeline

I think right now we are still overriding it even if it is explicitly set to None and setting it to Default value. I believe we can do what @jkasiraj suggested. We can set the constructor default to None and then override it by checking if it is None. 

```
if pipeline_experiment_config is None:  
     region =  self.sagemaker_session.boto_region_name
     if region is in EUREKA_GA_REGIONS:
          self.pipeline_experiment_config = _DEFAULT_EXPERIMENT_CFG
```

This way we will only override it if it is set to None or not set at all, which is the current behavior. If the user sets the value to something other than None, we need to keep that and add Validations in the Service to fail it.

Issue #5518, opened 2026-01-28
Pipeline TypeError: can only concatenate list (not "NoneType") to list Using SourceCode


CongaJAMM:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [X] PySDK V3 (3.3.1)

**Describe the bug**
Forces a parameter for an optional argument in a SourceCode object when using the Pipeline class upsert method.

**To reproduce**
A clear, step-by-step set of instructions to reproduce the bug.
The provided code need to be **complete** and **runnable**, if additional data is needed, please include them in the issue.
Conda environment with this requirements.yaml file content:
name: aws_env
channels:
  - pytorch
  - nvidia
  - conda-forge
  - defaults
dependencies:
  - python=3.12
  - pip=25.3
  - docker=20.10.9
  - imbalanced-learn=0.14.0
  - ipykernel=7.1.0
  - ipywidgets=8.1.8
  - numpy=2.4.0
  - pandas=2.3.2
  - pillow=10.4.0
  - pyarrow=21.0.0
  - pytorch=2.5.1
  - pytorch-cuda=12.4
  - s3fs=2025.10.0
  - scikit-learn=1.7.2
  - seaborn=0.13.2
  - pip:
    - sagemaker==3.3.1

Code -
def main():

src = SourceCode(
        source_dir=str(
            SOURCE_DIR
        ),  # Everything inside the `source_dir` will be copied to `/opt/ml/input/data/code/**`
        entry_script="weather_pred/training/train.py",  # Relative to the source_dir
        # command="python train.py epochs=1",
        ignore_patterns=[
            ".env",
            ".git",
            "__pycache__",
            ".DS_Store",
            ".cache",
            ".ipynb_checkpoints",
            "*.egg-info",
        ],
    )
    # ADD THIS LINE to satisfy the internal hashing utility
    compute = Compute(
        instance_type=TRAIN_INSTANCE_TYPE,
        instance_count=TRAIN_INSTANCE_COUNT,
        volume_size_in_gb=TRAIN_VOLUME_SIZE_GB,
    )
    my_output_config = OutputDataConfig(
        s3_output_path=TRAIN_S3_OUTPUT_PATH,  # Configure path to output artifacts
        compression_type="GZIP",  # Options: 'GZIP' (default) or 'NONE'
        # kms_key_id="your-kms-key-arn", # Optional: for encryption at rest
    )

    trainer = ModelTrainer(
[...]

Issue #5443, opened 2025-12-30
get_code_hash fails when trying to create a TrainStep with a ModelTrainer defined with a source_code


mikhailvorobiev1-max:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [X] PySDK V3 (3.x)

**Describe the bug**

When creating a pipeline that contains a TrainStep using a ModelTrainer with a custom source_code defined, running pipeline.definition() or pipeline.upsert() fails. 

The error comes from the [utilities](https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/workflow/utilities.py#L175) trying to get the code hash of the executed code and accessing an attribute that does not exist.

**To reproduce**
A clear, step-by-step set of instructions to reproduce the bug.
The provided code need to be **complete** and **runnable**, if additional data is needed, please include them in the issue.

1. Create a ModelTrainer with the source_code defined
2. Create a TrainStep using this ModelTrainer.train()
3. Add the TrainStep to a Pipeline
4. Upsert pipeline

**Expected behavior**

Pipeline is upserted.

**Screenshots or logs**
If applicable, add screenshots or logs to help explain your problem.

<img width="535" height="23" alt="Image" src="https://github.com/user-attachments/assets/c95e8c77-9abf-4d49-9e21-edb1c3d52e4f" />

**System information**
A description of your system. Please provide:
- **SageMaker Python SDK version**: 3.3.0
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**:
- **Framework version**:
- **Python version**:
- **CPU or GPU**:
- **Custom Docker image (Y/N)**:

**Additional context**

mohamedzeidan2021:
Both the original error and the follow-up validation failure are fixed.

- The original `get_code_hash` crash (accessing a nonexistent attribute on `SourceCode`) was fixed in #5467, shipped in 3.3.1 — the "partial fix" you observed.
- The remaining failure you hit after 3.3.1 was in `get_training_code_hash`, where the requirements/dependencies value (a single path string, or `None`) was being concatenated to a list. That was fixed in #5618 and shipped in sagemaker-core v2.5.1 (Mar 2026).

A `TrainStep` built from a `ModelTrainer` with `source_code` now hashes correctly through `pipeline.definition()` / `pipeline.upsert()`, covered by unit tests. Please upgrade to the current release. Closing as resolved — reopen with a repro if you still see a failure on the latest version.

PR #5527, opened 2026-01-30
Fix: Catch case where no `requirements` argument is sent to create S…

CongaJAMM:
Noticed SourceCode.requirements is an optional argument and causes an error when left to default `None` value in a TrainingStep  as part of a Pipeline.

Catches case where no `requirements` argument is sent to create SourceCode object and `get_code_hash` fails to provide a `List` to the `get_training_code_hash` function.

*Issue #, if available:* 
* [5443](https://github.com/aws/sagemaker-python-sdk/issues/5443)
* [5518](https://github.com/aws/sagemaker-python-sdk/issues/5518)

*Description of changes:*
From the `get_code_hash` function created a conditional statement to get the right DataType to pass to `get_training_code_hash` and supply a default value.

### DIFF
--- a/sagemaker-core/src/sagemaker/core/workflow/utilities.py
+++ b/sagemaker-core/src/sagemaker/core/workflow/utilities.py
@@ -172,7 +172,7 @@ def get_code_hash(step: Entity) -> str:
         source_code = model_trainer.source_code
         if source_code:
             source_dir = source_code.source_dir
-            requirements = source_code.requirements
+            requirements = [source_code.requirements] if source_code.requirements else []
             entry_point = source_code.entry_script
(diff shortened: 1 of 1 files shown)


sagemaker-core/src/sagemaker/core/workflow/utilities.py
```diff
@@ -172,7 +172,7 @@ def get_code_hash(step: Entity) -> str:
         source_code = model_trainer.source_code
         if source_code:
             source_dir = source_code.source_dir
-            requirements = source_code.requirements
+            requirements = [source_code.requirements] if source_code.requirements else []
```
mollyheamazon, line 175: Change to 
`requirements = source_code.requirements or None`

PR #5599, opened 2026-03-04
Feature store lakeformation

BassemHalim:
### Description                                                                                                                        
                                                                                                                                         
  This PR adds Lake Formation integration to SageMaker Feature Store, enabling customers to govern access to their offline store data through AWS Lake Formation instead of relying solely on IAM policies.                                                                  
                                                                                                                                         
  This simplifies the manual process described in this blog                                                                              
  https://aws.amazon.com/blogs/machine-learning/control-access-to-amazon-sagemaker-feature-store-offline-using-aws-lake-formation/       
                                                                                                                                         
  #### New Features                                                                                                                      
                                                                                                                                         
  **`LakeFormationConfig`** — declarative configuration for Lake Formation governance:                                                   
                                                                                                                                         
  ```python                                                                                                                              
  class LakeFormationConfig(Base):                                                                                                       
      enabled: bool = False                                                                                                              
      use_service_linked_role: bool = True                                                                                               
      registration_role_arn: Optional[str] = None                                                                                        
      disable_hybrid_access_mode: bool          # Required — no default                                                                  
      acknowledge_risk: bool                    # Required — no default                                      
```                                                                                                                                      
  `FeatureGroupManager.create()` — added lake_formation_config parameter                                                                 
                                                                                                                                         
  - Enables Lake Formation governance at Feature Group creation time                                                                     
  - Automatically waits for Feature Group to reach "Created" status before configuring Lake Formation                                    
                                                                                                                                         
  `FeatureGroupManager.enable_lake_formation()` — new method                                                                             
                                                                                                                                         
  - Enables Lake Formation on existing Feature Groups                                                                                    
  - Three-phase setup:                                                                                                                   
    1. Registers S3 location with Lake Formation                                                                                         
    2. Grants permissions to the execution role on the Glue table                                                                        
    3. Optionally revokes IAMAllowedPrincipal permissions from the Glue table (controlled by disable_hybrid_access_mode)                 
                                                                                                                                         
  - Fail-fast behavior with clear error reporting at each phase                                                                          
  - Interactive confirmation prompts warn users about risks (controllable via acknowledge_risk)                                          
  - Logs recommended S3 deny policy as a warning                                                                                         
                                                                                                                                         
  Usage                                                                                                                                  
                                                                                                                                         
  Enable at creation:                                                                                                                    
```python                                                                                                                                         
  from sagemaker.mlops.feature_store import FeatureGroupManager, LakeFormationConfig                                                     
                                                                                                                                         
  lf_config = LakeFormationConfig(                                                                                                       
      enabled=True,                                                                                                                      
      disable_hybrid_access_mode=True,                                                                                                   
      acknowledge_risk=True,  # skip interactive prompt                                                                                  
  )                                                                                                                                      
                                                                                                                                         
  fg = FeatureGroupManager.create(                                                                                                       
      feature_group_name="my-feature-group",                                                                                             
      # ... other params ...                                                                                                             
      lake_formation_config=lf_config,                                                                                                   
  )                                                                                                                                      
                                                                                                                                         
  #Enable on existing Feature Group:                                                                                                      
                                                                                                                                         
  fg = FeatureGroupManager.get("my-feature-group")                                                                                       
  result = fg.enable_lake_formation(                                                                                                     
      disable_hybrid_access_mode=True,                                                                                                   
      acknowledge_risk=True,                                                                                                             
  )                                                                                                                                      
```
                                                                                                                                         
  Testing                                                                                                                                
                                                                                                                                         
  - Unit tests: comprehensive coverage of all new methods, validation logic, and acknowledge_risk behavior                               
  - Integration tests: end-to-end tests for both creation workflows and negative scenarios                                               
                                                                                                                                         
  Notes                                                                                                                                  
                                                                                                                                         
  - S3 deny policy is logged as a recommendation (not applied automatically) to avoid breaking existing workflows                        
  - disable_hybrid_access_mode is a required field with no default — users must explicitly choose whether to revoke IAMAllowedPrincipal permissions                                                                                                                            
  - acknowledge_risk controls interactive confirmation: None (default) prompts via input(), True skips the prompt, False aborts with RuntimeError                                                                                                                           
                                                                                                                                         

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group.py
```diff
@@ -0,0 +1,746 @@
...
+    """
+
+    enabled: bool = False
+    use_service_linked_role: bool = True
+    registration_role_arn: Optional[str] = None
+    show_s3_policy: bool = False
+    disable_hybrid_access_mode: bool = True
+
+
+class FeatureGroup(CoreFeatureGroup):
```
nargokul: This would be confusing to the users to choose between `sagemaker.core.FeatureGroup` and `sagemaker.mlops.FeatureGroup`

Can this be utils or called something else ?

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group_manager.py
```diff
@@ -0,0 +1,746 @@
...
+            Lake Formation service-linked role ARN in format:
+            arn:{partition}:iam::{account}:role/aws-service-role/lakeformation.amazonaws.com/
+            AWSServiceRoleForLakeFormationDataAccess
+        """
+        partition = aws_partition(region) if region else "aws"
+        return (
+            f"arn:{partition}:iam::{account_id}:role/aws-service-role/"
+            f"lakeformation.amazonaws.com/AWSServiceRoleForLakeFormationDataAccess"
+        )
+
```
nargokul, line 192: Is this the string that gets created by default ? 

Can this functionality be gotten from an AWS library instead of hardcoding ?

If this changes , these changes would break

nargokul (commented):
Please fix integ tests

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group.py
BassemHalim: The v3 sagemaker.mlops version already re-exports the FeatureGroup from sagemaker-core [ref](https://github.com/aws/sagemaker-python-sdk/blob/d3dbc28ebfa2d705df8c47c23d949c1b165cdfe2/sagemaker-mlops/src/sagemaker/mlops/feature_store/MIGRATION_GUIDE.md#L49-L50) 

This class is meant as an extension to add an extra method `enable_lake_formation` and update the `create` method to enable lakeformation during creation

nargokul: re-exporting is fine. 
The class is a sagemaker-core resource class which does basic crud operations exactly mimicing boto. 
If we introfuce this change , then there is a difference in the behavior between core and mlops .

Why does this class need to be called FeatureGroup ?

BassemHalim: The idea is that it should be a drop-in replacement for the one in sagemaker-core just with extra functionality so that the user can enable lakeformation during feature group creation instead of creating then enabling. I was thinking this might be better UX. What do you suggest renaming the class or moving that to a utility function? I'm open to both but not sure what it can be renamed to

nargokul: I think we can go with something like `GovernedFeatureGroup` or `ManagedFeatureGroup`. 

We can discuss internally .

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group_manager.py
BassemHalim: Yes it gets created by default by Lakeformation [ref](https://docs.aws.amazon.com/lake-formation/latest/dg/service-linked-roles.html#:~:text=linked%20role%20named-,AWSServiceRoleForLakeFormationDataAccess,-.%20This%20role%20provides) 
I didn't find any functionality in the botocore sdk to get that role instead of hardcoding

I'm thinking it's provided in the lakeformation public docs so it's unlikely to change.

nargokul (dismissed):

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group_manager.py
```diff
@@ -0,0 +1,834 @@
...
+                f"IAMAllowedPrincipal permissions remain in effect, which means IAM-based access "
+                f"to the table is still allowed alongside Lake Formation permissions. "
+                f"For more info: https://docs.aws.amazon.com/lake-formation/latest/dg/hybrid-access-mode.html"
+            )
+            if acknowledge_risk is None:
+                proceed = input(
+                    "Hybrid access mode is not disabled. IAM-based access to the Glue table will "
+                    "still be allowed. Do you want to proceed without revoking IAMAllowedPrincipal "
+                    "permissions? (y/n): "
+                ).strip().lower() == "y"
```
mollyheamazon: `input()` is not safe for non-interactive environments (CI, pipelines, SageMaker jobs). 
Please require `acknowledge_risk` to be explicitly set to `True` or `False` and remove the `None` interactive path.

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group_manager.py
```diff
@@ -0,0 +1,834 @@
...
+                f"This will revoke IAMAllowedPrincipal permissions, which may break existing jobs "
+                f"(e.g., training, processing, ETL) that access this table via IAM-based permissions. "
+                f"After this change, all principals must be granted access through Lake Formation. "
+                f"For more info: https://docs.aws.amazon.com/lake-formation/latest/dg/hybrid-access-mode.html"
+            )
+            if acknowledge_risk is None:
+                proceed = input(
+                    "This will revoke IAMAllowedPrincipal permissions and may break existing jobs "
+                    "that rely on IAM-based access. Do you want to proceed? (y/n): "
+                ).strip().lower() == "y"
```
mollyheamazon: Same comment as above:
`input()` is not safe for non-interactive environments (CI, pipelines, SageMaker jobs). 
Please require `acknowledge_risk` to be explicitly set to `True` or `False` and remove the `None` interactive path.

mollyheamazon (dismissed):

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_group_manager.py
BassemHalim: Done, just made it a required field and removed `input()`

nargokul (dismissed):

mollyheamazon (dismissed):

mollyheamazon (approved):

nargokul (approved):

PR #5618, opened 2026-03-10
Fix handling of training step dependencies to allow successful pipeline creation

bobbywlindsey:
*Issue #, if available:*

*Description of changes:* Fix handling of dependencies in `get_training_code_hash` workflow utility so that training steps without a `requirements` param specified will still create a pipeline.

Originally, `dependencies` param of function `get_training_code_hash` assumed a list of strings. But for the training step, this is actually referring to `requirements` which can be a string, pipeline variable, or None - not a list. When this function is called, it tries to return `hash_files_or_dirs([source_dir] + dependencies)` but throws a list concatenation error because `dependencies` will always be either a string, pipeline variable, or None.

### DIFF
--- a/sagemaker-core/tests/unit/workflow/test_utilities.py
+++ b/sagemaker-core/tests/unit/workflow/test_utilities.py
@@ -269,27 +269,44 @@ class TestWorkflowUtilities:
         with tempfile.TemporaryDirectory() as temp_dir:
             entry_file = Path(temp_dir, "train.py")
             entry_file.write_text("print('training')")
+            requirements_file = Path(temp_dir, "requirements.txt")
+            requirements_file.write_text("numpy==1.21.0")
 
-            result = get_training_code_hash(
-                entry_point=str(entry_file), source_dir=temp_dir, dependencies=[]
+            result_no_deps = get_training_code_hash(
+                entry_point=str(entry_file), source_dir=temp_dir, dependencies=None
+            )
+            result_with_deps = get_training_code_hash(
+                entry_point=str(entry_file), source_dir=temp_dir, dependencies=str(requirements_file)
             )
 
-            assert result is not None
-            assert len(result) == 64
+            assert result_no_deps is not None
+            assert result_with_deps is not None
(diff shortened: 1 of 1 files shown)


sagemaker-core/src/sagemaker/core/workflow/utilities.py
```diff
@@ -248,11 +248,17 @@ def get_training_code_hash(entry_point: str, source_dir: str, dependencies: List
...
+                else:
+                    return hash_files_or_dirs([source_dir])
         elif entry_point:
             entry_point_url = urlparse(entry_point)
             if entry_point_url.scheme == "" or entry_point_url.scheme == "file":
-                return hash_files_or_dirs([entry_point] + dependencies)
+                if dependencies:
+                    return hash_files_or_dirs([entry_point] + [dependencies])
+                else:
+                    return hash_files_or_dirs([source_dir])
```
mollyheamazon: This should be `return hash_files_or_dirs([entry_point])`

sagemaker-core/src/sagemaker/core/workflow/utilities.py
mollyheamazon: Update to `def get_training_code_hash(entry_point: str, source_dir: str, dependencies: Optional[str] = None) -> str:`

sagemaker-core/src/sagemaker/core/workflow/utilities.py
```diff
@@ -236,9 +236,9 @@ def get_training_code_hash(entry_point: str, source_dir: str, dependencies: List
                 training
         source_dir (str): Path to a directory with any other training source
                 code dependencies aside from the entry point file
-        dependencies (str): A list of paths to directories (absolute
-                or relative) with any additional libraries that will be exported
-                to the container
+        dependencies (str): The relative path within ``source_dir`` to a
```
mollyheamazon: nit: Update to `dependencies Optional[str]`

mollyheamazon (commented):
Please add unit test to your fix

PR #5625, opened 2026-03-13
fixes for model builder

joshuatowner:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -985,11 +985,111 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
                     )
                     return
 
+        # Fallback: Nova recipes don't have hosting configs in the hub document
+        if self._is_nova_model():
+            nova_config = self._get_nova_hosting_config(instance_type=self.instance_type)
+            if not self.image_uri:
+                self.image_uri = nova_config["image_uri"]
+            if not self.env_vars:
+                self.env_vars = nova_config["env_vars"]
+            if not self.instance_type:
+                self.instance_type = nova_config["instance_type"]
+            return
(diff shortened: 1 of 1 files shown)


sagemaker-core/src/sagemaker/core/shapes/shapes.py
```diff
@@ -8577,7 +8577,7 @@ class InferenceComponentComputeResourceRequirements(Base):
     max_memory_required_in_mb: The maximum MB of memory to allocate to run a model that you assign to an inference component.
     """
 
-    min_memory_required_in_mb: int
+    min_memory_required_in_mb: Optional[int] = Unassigned()
```
nargokul, line 8580: Is this required for this to work ?
My only concern is that this is auto-generated and it would be overriden at some point . We would need to update the engine.

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -957,6 +957,10 @@ def _fetch_and_cache_recipe_config(self):
                     if not self.image_uri:
                         self.image_uri = config.get("EcrAddress")
 
+                    # Cache environment variables from recipe config
+                    if not self.env_vars:
+                        self.env_vars = config.get("Environment", {})
```
nargokul, line 962: If it does exist, should these be appended ?

sagemaker-core/src/sagemaker/core/shapes/shapes.py
joshuatowner: we should sort out where this is coming from under the hood, i did get errors describing the endpoint without it

sagemaker-serve/src/sagemaker/serve/model_builder.py
joshuatowner: guessing yes, we should fix

PR #5629, opened 2026-03-14
fix: resolve PermissionError during local mode cleanup of root-owned Docker files

mollyheamazon:
## Issue
Local container training fails with `PermissionError` during post-training cleanup.
Docker containers write files as root, and `shutil.rmtree` cannot remove them when running as a non-root user (e.g., `sagemaker-user` in Studio).

## Root Cause
The original `_rmtree` attempted `os.chmod` on PermissionError, which also fails on root-owned files when running as non-root.

## Fix
On PermissionError, use `docker run chmod -R 777` (as root inside the container) to fix permissions, then retry `shutil.rmtree`. In SageMaker Studio, `--network sagemaker` is added since Studio restricts Docker to that network. The already-pulled training image is reused to avoid pulling additional images.

## Changes
- `sagemaker-train/.../local/local_container.py` — updated `_rmtree`, pass image and is_studio from `train()` cleanup calls
- `sagemaker-core/.../local_core/local_container.py` — same fix, kept in sync
- Added unit tests for `_rmtree` covering normal, fallback, Studio, and error cases

## Testing
- Reproduced and verified fix on SageMaker Studio and local IDE
- Unit tests pass for both sagemaker-train (5 new) and sagemaker-core (31 existing)

### DIFF
--- a/migration.md
+++ b/migration.md
@@ -23,6 +23,122 @@
 * **Intelligent Defaults**: Smart default values based on context
 * **Advanced Features**: Access to latest SageMaker capabilities
 
+## Migration Tool (MCP Server)
+
+An AI-powered migration tool is available as an MCP server that can analyze your V2 code, transform it to V3, validate the results, and answer migration questions interactively through your IDE.
+
+### Installation
+
+```bash
+pip install --no-cache-dir https://d3azyja9oqj8z1.cloudfront.net/sagemaker_sdk_helper-0.2.0.tar.gz --force-reinstall
+```
+
+Verify installation:
+
+```bash
+which sagemaker-migration-mcp   # Should output the path to the executable
+sagemaker-migration-mcp --help   # Test the server runs
+```
+
+### IDE Setup
+
+Add the following to your MCP configuration file:
+
+| IDE | Config Location |
+|-----|----------------|
+| Kiro | `~/.kiro/settings/mcp.json` |
+| Kiro CLI | `~/.kiro/settings/mcp.json` |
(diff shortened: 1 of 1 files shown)


sagemaker-core/src/sagemaker/core/modules/local_core/local_container.py
```diff
@@ -58,15 +57,34 @@
...
-    def _onerror(func, path, exc_info):
-        if isinstance(exc_info[1], PermissionError):
-            os.chmod(path, stat.S_IRWXU)
-            func(path)
-        else:
-            raise exc_info[1]
-    shutil.rmtree(path, onerror=_onerror)
+    try:
+        shutil.rmtree(path)
+    except PermissionError:
```
nargokul, line 64: What is there is another Error here?

Do we still need `onerror=_onerror`?

sagemaker-core/src/sagemaker/core/modules/local_core/local_container.py
nargokul, line 1: MIssing unit tests for this file

sagemaker-core/src/sagemaker/core/modules/local_core/local_container.py
mollyheamazon: The old `_onerror` callback only handled `PermissionError` — for any other exception type it did `raise exc_info[1]`, which is the same as not catching it at all. 

The new code has the same behavior: `shutil.rmtree` is called without `onerror`, so any non-`PermissionError` exception propagates naturally. The explicit except `PermissionError` makes the intent clearer.

Issue #5610, opened 2026-03-09
Sagemaker HyperParameterTuner fails to start distributed training


CoolFish88:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**
When submitting a standalone Sagemaker Training Job using ModelTrainer, the job completes successfully while the same training job terminates with the following error when executed within a HyperparameterTuner instance.  

`RuntimeError: The size of tensor a (4) must match the size of tensor b (16) at non-singleton dimension 0`
(in my config I have: per_device_train_batch_size=4, gradient_accumulation_steps=4)

**Note:** The standalone job uses the sm_drivers channel and torchrun_driver.py, launching 4 processes across 4 GPUs. The tuning job uses the legacy framework container path (sagemaker_pytorch_container), which invokes python train.py directly — single GPU only.

**Claude log analysis**: 

> In TRL 0.24.0's compute_loss:
> entropy = torch.sum(per_token_entropy * attention_mask) / attention_mask.sum()
> With your config: per_device_train_batch_size=4, gradient_accumulation_steps=4. In single-GPU mode, TRL's num_items_in_batch logic produces per_token_entropy with batch dim 4, but attention_mask reflects the accumulated batch 4×4=16. This mismatch doesn't occur in distributed mode because batch dimensions are correctly partitioned across 4 ranks.
> Quick verification: Check what model_trainer.input_data_config contains when submitted standalone vs. via tuner — the sm_drivers channel is very likely missing from the tuner's job definition, which is the proximate cause of the different execution path.

**To reproduce**
**Launching Sagemaker Training Job:**
```
training_image = "763104351884.dkr.ecr.us-west-2.amazonaws.com/huggingface-pytorch-training:2.8.0-transformers4.56.2-gpu-py312-cu129-ubuntu22.04-v1.0"
training_mode = Mode.SAGEMAKER_TRAINING_JOB
distributed_strategy = Torchrun()

model_trainer = ModelTrainer(
    training_mode=training_mode,
    sagemaker_session=sagemaker_session,
    role=role,
    training_image=training_image,
    base_job_name=job_name,
    source_code=source_code,
    compute=compute,
    distributed=distributed_strategy,
    output_data_config=output,
    checkpoint_config=checkpoint_config,
    stopping_condition=stopping,
    environment={"PYTORCH_CUDA_ALLOC_CONF": "expandable_segments:True"},
[...]

PR #5634, opened 2026-03-17
fix(tuner): Include sm_drivers channel in HyperparameterTuner jobs

nargokul:
*Issue #, if available:*
https://github.com/aws/sagemaker-python-sdk/issues/5610

*Description of changes:*
When ModelTrainer has distributed=Torchrun(), the sm_drivers channel contains torchrun_driver.py and sm_train.sh which are required for multi-GPU execution. The tuner was not building this channel, causing the framework container to fall back to the legacy single-GPU entry point (python train.py) instead of torchrun.

This caused a tensor size mismatch (batch_size vs accumulated_batch) in TRL's compute_loss when gradient_accumulation_steps > 1, because the single-process path doesn't partition batches across ranks.

Fix: Replace _upload_source_code_and_configure_hyperparameters with _build_driver_and_code_channels that replicates ModelTrainer's channel building logic (sm_drivers, code, distributed.json, sourcecode.json, sm_train.sh). Also pass through environment and VPC config.

### DIFF
--- a/sagemaker-train/tests/unit/train/test_tuner_driver_channels.py
+++ b/sagemaker-train/tests/unit/train/test_tuner_driver_channels.py
@@ -1,0 +1,549 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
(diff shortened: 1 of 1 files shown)


mollyheamazon (dismissed):
Can you add unit tests?

PR #5642, opened 2026-03-18
Bedrock fix

nargokul:
Source (bedrock_model_builder.py):

Extract _is_nova_model() module-level helper using safe getattr with defaults, replacing all duplicated hasattr chains
Add ValueError guards in deploy() for missing model_package, custom_model_name, and role_arn
Add ValueError guard in create_deployment() for empty/None model_arn
Initialize status = None before the polling loop to prevent UnboundLocalError
Move json and urlparse imports to module level
Replace f-string logging with lazy %s formatting throughout

### DIFF
--- a/sagemaker-serve/example_notebooks/bedrock_nova_deployment.ipynb
+++ b/sagemaker-serve/example_notebooks/bedrock_nova_deployment.ipynb
@@ -4,26 +4,26 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Deploy a Nova Model to Amazon Bedrock\n",
+    "# Deploy a Fine-Tuned Nova Model to Amazon Bedrock\n",
     "\n",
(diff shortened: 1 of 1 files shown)


sagemaker-serve/src/sagemaker/serve/bedrock_model_builder.py
```diff
@@ -22,10 +26,33 @@
 from sagemaker.core.telemetry.telemetry_logging import _telemetry_emitter
 from sagemaker.core.telemetry.constants import Feature
 
+logger = logging.getLogger(__name__)
+
+
+def _is_nova_model(container) -> bool:
```
rsareddy0329, line 32: not a blocker, but I think this can be moved to common utils as model builder also has this method

rsareddy0329 (approved):

PR #5656, opened 2026-03-19
fix: Replace HMAC with SHA based integrity checks in model builder.

pravali96:
*Issue #, if available:*

*Description of changes:*
Replace HMAC with SHA based integrity checks in model builder.

## Fix

Switch from HMAC-SHA256 (requires a secret key) to plain SHA-256 (no key needed)

### Changes

**`check_integrity.py`**
- Removed `generate_secret_key()` — no longer needed
- `compute_hash()` now uses `hashlib.sha256()` instead of `hmac.new()`
- `perform_integrity_check()` no longer reads `SAGEMAKER_SERVE_SECRET_KEY` from environment
 

### DIFF
--- a/src/sagemaker/serve/model_server/multi_model_server/prepare.py
+++ b/src/sagemaker/serve/model_server/multi_model_server/prepare.py
@@ -26,7 +26,6 @@ from sagemaker.session import Session
 from sagemaker.serve.spec.inference_spec import InferenceSpec
 from sagemaker.serve.detector.dependency_manager import capture_dependencies
 from sagemaker.serve.validations.check_integrity import (
-    generate_secret_key,
     compute_hash,
 )
 from sagemaker.remote_function.core.serialization import _MetaData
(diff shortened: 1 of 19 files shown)


src/sagemaker/serve/model_server/multi_model_server/prepare.py
```diff
@@ -120,11 +119,10 @@ def prepare_for_mms(
...
-    secret_key = generate_secret_key()
     with open(str(code_dir.joinpath("serve.pkl")), "rb") as f:
         buffer = f.read()
-    hash_value = compute_hash(buffer=buffer, secret_key=secret_key)
+    hash_value = compute_hash(buffer=buffer)
     with open(str(code_dir.joinpath("metadata.json")), "wb") as metadata:
         metadata.write(_MetaData(hash_value).to_json())
 
-    return secret_key
+    return ""
```
nargokul: Nit: We need not return anything here

pravali96: The 11 failing tests are in JumpStart, not model builder. They are unrelated to these changes. is that a blocker?

pravali96: it's not needed for long-term but its better to keep `return ""` for backward compatibility in this security fix. Callers (`self.secret_key = prepare_for_mms(...)`) still assign this return value, so it cannot be none.

pravali96: Removed the return statements.

PR #5666, opened 2026-03-23
fix: aws_batch integ test resources are now uniquely named by test run.

davlind-amzn:
*Description of changes:* aws_batch integ test resources are now uniquely named by test run.  This is required to prevent overlapping test runtimes (e.g. when multiple pysdk PRs are open) clobbering each other's AWS Batch test resources.  Resource creation is also now simplified by eliminating the option to reuse resources between independent test executions (wasn't used prior).

Other changes:
- Improved exception handling around integ tests, cleaning up resources when an exception is thrown during resource creation or test runtime.
- Small syntax bug fix in the quota management example notebook.

### DIFF
--- a/sagemaker-train/tests/integ/train/aws_batch/test_queue.py
+++ b/sagemaker-train/tests/integ/train/aws_batch/test_queue.py
@@ -15,6 +15,8 @@ from __future__ import absolute_import
 import boto3
 import botocore
 import pytest
+import random
+import string
 
 from sagemaker.train.model_trainer import ModelTrainer
 from sagemaker.train.configs import SourceCode, InputData, Compute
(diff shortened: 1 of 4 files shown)


sagemaker-train/tests/integ/train/aws_batch/test_queue.py
```diff
@@ -29,16 +31,34 @@
...
+    print(f"Integration test ID (used in AWS Batch resource naming): {test_id}")
+
+    resource_manager = BatchTestResourceManager(batch_client=batch_client, test_id=test_id)
+
+    try:
+        resource_manager.get_or_create_resources()
+        yield resource_manager
+    except Exception as e:
+        print(f"Exception thrown while creating or yielding AWS Batch resources: {str(e)}")
+
```
mollyheamazon, line 61: should raise exception here
add `raise`

davlind-amzn: I intentionally didn't raise here because I want cleanup to still attempt if we catch an exception

PR #5670, opened 2026-03-24
feat(pipeline): Make _PipelineExecution a public class

### LINKED ISSUE #4391: Make _PipelineExecution public

[_PipelineExecution](https://sagemaker.readthedocs.io/en/stable/workflows/pipelines/sagemaker.workflow.pipelines.html#sagemaker.workflow.pipeline._PipelineExecution) is a handy class that allow users to view and control executions, even if they are not started by the Python SDK. For example, I have an execution "blah" started from the Studio UI

```
execution = _PipelineExecution(arn="blah")
execution.result()
```

However, it is marked as protected. 

Make it a public class and add documentation with code examples. 

```
execution = PipelineExecution(arn="blah")
execution.result()
```

**Additional context**

Rishabh0255:
Renames _PipelineExecution to PipelineExecution and exports it from the workflow module. Keeps the old private name available for backward compatibility.

Closes #4391


sagemaker-mlops/src/sagemaker/mlops/workflow/pipeline.py
```diff
@@ -1087,6 +1100,10 @@ def result(self, step_name: str):
         )
 
 
+# Backward-compatible alias for the previously private class name
+_PipelineExecution = PipelineExecution
```
nmadan: Don't think this is needed. The class was private to begin with. We only need to update all references in this repository

Rishabh0255: Understood removed backward compatibility line

nmadan (approved):

nargokul (approved):

PR #5692, opened 2026-03-27
Fix: hardcode handler_name = "lambda_function.lambda_handler" to match the zip entry name.

mollyheamazon:
*Issue #, if available:*
Ticket: https://tiny.amazon.com/12ooeo0ys/tcorpamazP405over

*Description of changes:*
- One-line fix in evaluator.py line 385
- Updated test_create_with_byoc to assert the correct Handler value is passed to create_function

### DIFF
--- a/sagemaker-train/tests/integ/ai_registry/test_evaluator.py
+++ b/sagemaker-train/tests/integ/ai_registry/test_evaluator.py
@@ -103,6 +103,11 @@ class TestEvaluatorIntegration:
         assert evaluator.method == EvaluatorMethod.BYOC
         assert evaluator.reference is not None
 
+        # Wait for Lambda to become Active before invoking
+        lambda_client = boto3.client("lambda")
+        waiter = lambda_client.get_waiter("function_active_v2")
+        waiter.wait(FunctionName=evaluator.reference)
+
         # Invoke the Lambda directly to verify the handler is correct
         lambda_client = boto3.client("lambda")
         response = lambda_client.invoke(


sagemaker-train/src/sagemaker/ai_registry/evaluator.py
```diff
@@ -382,7 +382,7 @@ def _create_lambda_function(cls, name: str, source_file: str, role: Optional[str
         # Create Lambda function
         lambda_client = boto3.client("lambda")
         function_name = f"SageMaker-evaluator-{name}-{datetime.now().strftime('%Y%m%d_%H%M%S')}"
-        handler_name = f"{os.path.splitext(os.path.basename(source_file))[0]}.lambda_handler"
+        handler_name = "lambda_function.lambda_handler"
```
mufaddal-rohawala, line 385: The fix itself makes sense if the zip archive always packages the source file as `lambda_function.py` regardless of the original filename. However, this introduces a subtle coupling — the hardcoded string `"lambda_function.lambda_handler"` must match whatever entry name is used when building the zip. Consider extracting this as a module-level constant (e.g., `LAMBDA_HANDLER_NAME = "lambda_function.lambda_handler"`) and adding a comment explaining *why* it's hardcoded (i.e., the zip archive always renames the source to `lambda_function.py`). This will prevent future contributors from "fixing" it back to the dynamic version.

Also, as a pre-existing issue: this method calls `boto3.client("lambda")` directly (line 383). Per V3 architecture tenets, all AWS API interactions should go through `sagemaker-core` rather than calling boto3 directly. This is out of scope for this PR, but worth tracking.

sagemaker-train/tests/unit/ai_registry/test_evaluator.py
```diff
@@ -77,6 +77,9 @@ def test_create_with_byoc(self, mock_air_hub, mock_boto3):
         
         assert evaluator.method == EvaluatorMethod.BYOC
         mock_air_hub.upload_to_s3.assert_called_once()
```
mufaddal-rohawala, line 79: Good addition to validate the handler value. Minor suggestion: consider also asserting that the `Handler` key is specifically `"lambda_function.lambda_handler"` using a more descriptive test or a comment explaining the invariant — e.g., that the zip always uses `lambda_function.py` as the module name. This helps future readers understand *why* this assertion exists.

Also, accessing `call_args[1]` assumes keyword arguments. This is fine for the current mock setup, but using `mock_lambda_client.create_function.call_args.kwargs` (Python 3.8+) would be slightly more readable and explicit.

mufaddal-rohawala (commented):
## 🤖 AI Code Review

This PR fixes a bug where the Lambda handler name was dynamically derived from the source file name, but the zip entry always uses 'lambda_function.py' as the module name. The fix is straightforward and the test update correctly validates the new behavior. However, there are two notable concerns: a direct boto3 call (pre-existing) and a missing explanation of why the handler name must be hardcoded.

PR #5774, opened 2026-04-17
Fix/wait respects sagemaker session 5765

aviruthen:
*Issue #, if available:*

Bug fix for the first issue mentioned in GitHub issue 5765! Ports waiting logic from PySDK V2 to V3 and allows users to optionally wait for Processing Jobs or Training Jobs without running into the aforementioned NoCredentialsError

*Description of changes:*

Did local integration testing to verify job waiting was working as expected, added unit tests

### DIFF
--- a/sagemaker-train/src/sagemaker/train/model_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/model_trainer.py
@@ -790,7 +790,11 @@ class ModelTrainer(BaseModel):
             self._latest_training_job = training_job
 
             if wait:
-                training_job.wait(logs=logs)
+                from sagemaker.train.common_utils.trainer_wait import wait as trainer_wait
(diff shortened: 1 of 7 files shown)


sagemaker-train/src/sagemaker/train/model_trainer.py
```diff
@@ -790,7 +791,10 @@ def train(
             self._latest_training_job = training_job
 
             if wait:
-                training_job.wait(logs=logs)
+                trainer_wait(
```
mufaddal-rohawala, line 794: Could you help me understand the reasoning for moving away from the autogenerated waiter here? I noticed we introduced a dedicated waiter for ModelCustomization tailored to those specific flows — I'm curious whether the intent is to also adopt custom waiters across all high-level abstractions going forward?

PR #5778, opened 2026-04-20
fix: make local mode health-check timeout configurable via session config

### LINKED ISSUE #3362: Configurable (or just much longer?) health-check timeout in local mode

Today, local-mode endpoint deployment uses a [hard-coded health check time-out](https://github.com/aws/sagemaker-python-sdk/blob/ace07d72f4f44c43fe95b05574968decc7e806ac/src/sagemaker/local/entities.py#L44) of 120s for the container to become healthy.

This does not appear to be consistent with the start-up requirements for actual SageMaker endpoints, and even if it was, it may not be appropriate to assume local environments have similar network bandwidth or compute capabilities to target instance types.

I'm currently testing a use case with large (e.g. ~5GB+) model archives, and finding local mode deployment fails due to this healthcheck time-out, even though actual SageMaker endpoint deployments succeed without any issue.

If the default timeout was significantly longer, I think it should work okay. If the default timeout was configurable somehow, I could force it to wait longer for my use case.

Possible options could include:
- Extending the timeout
- Making the timeout configurable
- Somehow excluding tarball download/extract time from the coverage of the timeout check
- Supporting decompressed local folders as `model_data` targets for local models/endpoints - instead of requiring S3/tarball.

**Additional context**

N/A

egorkosaretsky:
## Problem

Local mode endpoint deployment uses a hard-coded `HEALTH_CHECK_TIMEOUT_LIMIT` of 120s ([entities.py:43](https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/local/entities.py#L43)). This causes failures when containers take longer to start — e.g. large model archives (~5GB+), slow pip installs, or low-bandwidth environments — even though the actual SageMaker endpoint deployment succeeds fine.

Fixes #3362.

## Solution

- Added an optional `timeout` parameter to `_wait_for_serving_container()` (defaults to `HEALTH_CHECK_TIMEOUT_LIMIT` for backwards compatibility)
- Both `_LocalEndpoint.serve()` and `_LocalTransformJob.start()` now read `local.health_check_timeout` from the session config and pass it through

**Usage:**
```python
sess = LocalSession()
sess.config = {'local': {'health_check_timeout': 600}}  # 10 minutes
```

## Testing

Added unit tests covering:
- Custom timeout value is respected by `_wait_for_serving_container`
- `health_check_timeout` from session config is passed through in `_LocalEndpoint.serve()`
- Default behaviour (120s) is unchanged when config is not set

## Checklist

- [x] Unit tests added/updated
- [x] Backwards compatible (default unchanged)
- [x] Both endpoint and transform job code paths updated

### DIFF
--- a/sagemaker-core/src/sagemaker/core/local/entities.py
+++ b/sagemaker-core/src/sagemaker/core/local/entities.py
@@ -332,7 +332,11 @@ class _LocalTransformJob(object):
         self.container.serve(self.primary_container["ModelDataUrl"], environment)
 
         serving_port = get_config_value("local.serving_port", self.local_session.config) or 8080
-        _wait_for_serving_container(serving_port)
+        health_check_timeout = (
+            get_config_value("local.health_check_timeout", self.local_session.config)
+            or HEALTH_CHECK_TIMEOUT_LIMIT
+        )
+        _wait_for_serving_container(serving_port, timeout=health_check_timeout)
 
         # Get capabilities from Container if needed
         endpoint_url = "http://%s:%d/execution-parameters" % (get_docker_host(), serving_port)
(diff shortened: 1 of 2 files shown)


egorkosaretsky:
@nargokul Could you please take a look, this should fix https://github.com/aws/sagemaker-python-sdk/issues/3362

PR #5781, opened 2026-04-21
feature: support custom hub_name in fine-tuning trainers and model evaluators for private hub testing

mollyheamazon:
## Issue
Fine-tuning trainers (`SFTTrainer`, `DPOTrainer`, `RLVRTrainer`, `RLAIFTrainer`)
and model evaluators (`BenchMarkEvaluator`, `CustomScorerEvaluator`) hardcode
`hub_name="SageMakerPublicHub"` when fetching recipes, override parameters, and
studio tags. This blocks development and E2E testing of new recipes — a recipe
can only be exercised end-to-end after it ships to the public JumpStart hub,
delaying any PySDK fixes or support work that depends on the recipe.

Studio already uses private hubs for its own E2E suite; PySDK trainers need to
support the same pattern so test accounts can publish a pre-release recipe to a
private hub and exercise the full trainer path against it.

## Description
Adds an optional `hub_name` parameter that threads through all hub-content
lookups in the trainer and evaluator flows. Defaults to `"SageMakerPublicHub"`
so existing callers are unaffected.

## Usage
```
# Default behavior — unchanged
SFTTrainer(model="meta-textgeneration-llama-3-2-1b-instruct", ...)

# Test a pre-release recipe from a private hub
SFTTrainer(
   model="my-prerelease-model",
   model_package_group="my-group",
   hub_name="MyPrivateHub",
)

# Same pattern works for DPOTrainer, RLVRTrainer, RLAIFTrainer,
# BenchMarkEvaluator, CustomScorerEvaluator
```

### DIFF
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -4,6 +4,7 @@ from sagemaker.core.helper.session_helper import Session
 from sagemaker.core.training.configs import Tag, Networking, InputData, Channel
 from sagemaker.core.shapes import shapes
 from sagemaker.core.resources import TrainingJob
+from sagemaker.train.constants import HUB_NAME
 
 
 class BaseTrainer(ABC):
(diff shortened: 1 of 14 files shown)


sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -36,6 +37,10 @@ class BaseTrainer(ABC):
             Can include training and validation datasets.
         environment (Optional[Dict[str, str]]):
             Environment variables to set in the training container.
+        hub_name (Optional[str]):
```
zachgk, line 40: This isn't a customer accessible feature so we should avoid having this documented where customers can see it

PR #5785, opened 2026-04-22
fix: improve error messages for waiter timeouts

joshuatowner:
Improve timeout error messages for TrainingJob and EvaluationJob waits

When wait() times out, the current error says "Increase the timeout and try again" — which implies the job stopped and needs to be re-created. In reality, the job is still running server-side; only the client-side polling stopped.

Changes:

- Add message parameter to TimeoutExceededError so callers can provide context-specific guidance (defaults to existing message for backward compat)
- TrainingJob.wait(): timeout now says "Your training job is still running. Call .refresh() to check its current status."
- EvaluationPipelineExecution.wait(): timeout now says "Your evaluation job is still running. Use .refresh() to check its current status." Also changes resource_type from PipelineExecution to EvaluationJob for clarity.
- Fix pre-existing typo resouce_type → resource_type in TrainingJob.wait() (would have caused TypeError at runtime)

### DIFF
--- a/sagemaker-core/src/sagemaker/core/resources.py
+++ b/sagemaker-core/src/sagemaker/core/resources.py
@@ -972,7 +972,7 @@ class Algorithm(Base):
                     )
 
                 if timeout is not None and time.time() - start_time >= timeout:
-                    raise TimeoutExceededError(resouce_type="Algorithm", status=current_status)
+                    raise TimeoutExceededError(resource_type="Algorithm", status=current_status)
                 time.sleep(poll)
 
     @Base.add_validate_call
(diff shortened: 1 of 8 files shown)


sagemaker-core/src/sagemaker/core/resources.py
```diff
@@ -35883,7 +35883,11 @@ def wait(
                     return
 
                 if timeout is not None and time.time() - start_time >= timeout:
-                    raise TimeoutExceededError(resouce_type="TrainingJob", status=current_status)
+                    raise TimeoutExceededError(
+                        resource_type="TrainingJob",
+                        status=current_status,
```
mollyheamazon, line 35888: You have helped caught typo of this autogenerated script: should be `resource_type` instead of `resouce_type`, thanks! Can you help update the source for this script 
1. https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/tools/templates.py#L335, also line 388, 439.
2. Regenerate `resources.py`:
```
   cd sagemaker-core
   python -m sagemaker.core.tools.codegen
```
3. That single template fix will correct all 85+ resources in the generated 
resources.py at once

gauravmadarkal (approved):
LGTM

mollyheamazon (approved):

PR #5791, opened 2026-04-22
Fix output data config in HyperparameterTuner

zhaoqizqwang:
fix(tuner): pass through full OutputDataConfig from ModelTrainer
HyperparameterTuner._build_training_job_definition was reconstructing a new
OutputDataConfig with only s3_output_path, silently dropping kms_key_id,
compression_type, and other fields set on the ModelTrainer.

Pass model_trainer.output_data_config directly to preserve all fields.

Also update _create_mock_model_trainer in tests to use a real OutputDataConfig
instead of MagicMock, and add a test verifying kms_key_id and compression_type
are preserved through the tuning job definition.


sagemaker-train/src/sagemaker/train/tuner.py
```diff
@@ -1472,13 +1472,10 @@ def _build_training_job_definition(self, inputs):
...
-        output_config = OutputDataConfig(
-            s3_output_path=(
-                model_trainer.output_data_config.s3_output_path
-                if model_trainer.output_data_config
-                else None
-            )
+        # Pass through the full OutputDataConfig from ModelTrainer so that
+        # kms_key_id, compression_type, and any other fields are preserved.
+        output_config = model_trainer.output_data_config or OutputDataConfig(
+            s3_output_path=None
```
nargokul, line 1478: Why do we need to set this to None.

sagemaker-train/src/sagemaker/train/evaluate/Resolved template parameters: {'role_arn.md
```diff
@@ -1,172 +0,0 @@
-Resolved template parameters: {'role_arn':                       base_evaluator.py:757
-                             'arn:aws:iam::634683118556:role/service-role/AmazonSageMaker-Exe
```
mollyheamazon, line 2: What is this file? Looks like it is committed accidentally

mollyheamazon (changes requested):

sagemaker-train/src/sagemaker/train/tuner.py
zhaoqizqwang: `OutputDataConfig` is a required field in `create_hyperparameter_tuning_job`. We use an empty object when we don't get the config from ModelTrainer, which is what the old code does

sagemaker-train/src/sagemaker/train/evaluate/Resolved template parameters: {'role_arn.md
zhaoqizqwang: Yes this file is not used

mollyheamazon (approved):

PR #5800, opened 2026-04-28
fix: refresh sagemaker-core service model and regenerate SDK

lucasjia-aws:
# Issue

The `service-2.json` bundled in sagemaker-core had drifted from the [public botocore service model](https://github.com/boto/botocore/blob/master/botocore/data/sagemaker/2017-07-24/service-2.json). This caused the generated SDK to contain stale API definitions and shape fields that no longer match the public SageMaker API surface.

# Fix

Replaced `service-2.json` with the latest [public botocore service model](https://github.com/boto/botocore/blob/master/botocore/data/sagemaker/2017-07-24/service-2.json), then re-ran codegen (`python3 -m sagemaker.core.tools.codegen`) to regenerate all downstream files.

**Files regenerated by codegen:**

File | Effect
-- | --
`src/sagemaker/core/resources.py` | Updated resource classes and fields to match the current public API surface
`resource_plan.csv` | Updated resource plan entries
`src/sagemaker/core/config_schema.py` | Updated config schema definitions
`src/sagemaker/core/utils/code_injection/shape_dag.py` | Updated shape dependency graph
`src/sagemaker/core/tools/api_coverage.json` | Updated API coverage numbers (SupportedAPIs: 361→387, UnsupportedAPIs: 6→18)

**Additional changes:**

File | What and Why
-- | --
`src/sagemaker/core/shapes/shapes.py` | Copied from the regenerated top-level `shapes.py` to `shapes/shapes.py`. Python resolves `from sagemaker.core.shapes import *` to the `shapes/` package, so this copy must stay in sync.
`src/sagemaker/core/resources.py` (L30878) | Fixed a pre-existing codegen bug: the `TrainingJob.wait()` method had an `if` block indented at 16 spaces instead of 8, causing `SyntaxError`. Manually dedented to the correct level.
~43 other files (various `.py` and test files) | Code formatting only (whitespace, line wrapping) produced by `black -l 100 .` as the final codegen step. No functional changes.

# Follow-up Commits

These commits fix CI failures and runtime issues discovered after the initial model refresh.

Commit | File(s) | What & Why
-- | -- | --
`f252f4da` | `sagemaker-train`, `sagemaker-mlops`, `sagemaker-core` codegen tools, `shapes.py`, `resources.py` | Removed references to shapes/fields that no longer exist in the updated model (`enable_ffm`, `AddOnlineStoreReplicaAction`). Added `IntPipeVar` support and `REQUIRED_TO_OPTIONAL_OVERRIDES` to codegen, then regenerated.
`ae707dc` | `sagemaker-core/.../resources.py` | Added `Session = Boto3Session` alias to fix `NameError`.
`30c4c0c` | `sagemaker-core/.../shapes/shapes.py` | Synced `shapes/shapes.py` package copy with updated top-level `shapes.py`.
`7aac55c` | `sagemaker-mlops/.../test_iceberg_properties.py` | Replaced removed `description=` param with `feature_additions=[]` in test calls.


sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -30,7 +30,7 @@
             },
             # Override default instance type; the model's default
             # (ml.p3.2xlarge) is deprecated.
-            "compute": Compute(instance_type="ml.g5.xlarge"),
```
nargokul, line 33: Why are we making this change ?

sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
```diff
@@ -421,9 +420,8 @@ def _create_input_channels(dataset: str, content_type: Optional[str] = None,
         content_type=content_type,
         compression_type=input_compression_type,
         record_wrapper_type=record_wrapper_type,
-        input_mode=input_mode,
-        enable_ffm=enable_ffm,
-    )
+        input_mode=input_mode
```
nargokul, line 423: Why are we removing this ?

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
lucasjia-aws: This is a by-the-way integ-test fixing not corresponding to this issue. Confirmed with Zhaoqi.

sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
lucasjia-aws: After replacing `service-2.json`, `enable_ffm` no longer exist in shapes.py, need to remove, also it is a dead call actually, safe to remove.

PR #5804, opened 2026-04-28
fix: potential s3 path traversal

zhaoqizqwang:
Fix potential s3 path traversal by using `realpath()` and containment checks before every S3-to-local file write

### DIFF
--- a/sagemaker-core/src/sagemaker/core/common_utils.py
+++ b/sagemaker-core/src/sagemaker/core/common_utils.py
@@ -467,13 +467,7 @@ def _download_files_under_prefix(bucket_name, prefix, target, s3):
         s3_relative_path = obj_sum.key[len(prefix) :].lstrip("/")
         file_path = os.path.join(target, s3_relative_path)
 
-        # Validate that the resolved file path stays within the target directory
-        file_path_real = os.path.realpath(file_path)
-        if not file_path_real.startswith(target_real + os.sep) and file_path_real != target_real:
-            raise ValueError(
-                f"Path traversal detected: S3 key '{obj_sum.key}' resolves to "
-                f"'{file_path_real}' which is outside the target directory '{target_real}'"
-            )
+        validate_path_within_directory(file_path, target, source_description=obj_sum.key)
 
         try:
             os.makedirs(os.path.dirname(file_path))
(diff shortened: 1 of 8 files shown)


sagemaker-core/src/sagemaker/core/local/entities.py
```diff
@@ -509,14 +509,28 @@ def _perform_batch_inference(self, input_data, output_data, **kwargs):
...
-            copy_directory_structure(working_dir, relative_path)
             destination_path = os.path.join(working_dir, relative_path, filename + ".out")
 
+            # Validate that the destination stays within the working directory
+            destination_path_real = os.path.realpath(destination_path)
+            if (
+                not destination_path_real.startswith(working_dir_real + os.sep)
+                and destination_path_real != working_dir_real
+            ):
+                raise ValueError(
```
mujtaba1747: Are we okay if some customer use cases break because we throw an error here?
I think we should be fine because our justification is that this is a security issue.

sagemaker-core/src/sagemaker/core/common_utils.py
```diff
@@ -466,6 +467,14 @@ def _download_files_under_prefix(bucket_name, prefix, target, s3):
         s3_relative_path = obj_sum.key[len(prefix) :].lstrip("/")
         file_path = os.path.join(target, s3_relative_path)
 
+        # Validate that the resolved file path stays within the target directory
+        file_path_real = os.path.realpath(file_path)
+        if not file_path_real.startswith(target_real + os.sep) and file_path_real != target_real:
```
mujtaba1747: It maybe better to create a common util function as this logic is being repeated in multiple places.

Logic: file_path_real.startswith(target_real + os.sep)

I maybe wrong here, perhaps the path check is each place is slightly different. But please take a look if all of them can use a single util function.

mujtaba1747 (changes requested):

sagemaker-core/src/sagemaker/core/local/entities.py
zhaoqizqwang: yes if path traversal isdetected, workflow should stop and users should be warned against it

sagemaker-core/src/sagemaker/core/common_utils.py
zhaoqizqwang: Good call, I moved them into one common util in the new commit

zhaoqizqwang:
The failed integ tests in sagemaker-core and sagemaker-train are not related to code change in this PR

mujtaba1747 (approved):

Issue #5802, opened 2026-04-28
re-enable skipped integ tests in sagemaker-serve


mujtaba1747:
Integration tests in sagemaker serve test_deploy_from_training_job and test_optimize_build_deploy_invoke_cleanup were skipped due to intermittent failures.

PR #5807, opened 2026-04-29
fix: Un-skipping tests in sagemaker-serve

mujtaba1747:
*Issue #, if available:* https://github.com/aws/sagemaker-python-sdk/issues/5802

*Description of changes:* Fix skipped tests in sagemaker-serve

## test_model_customization_deployment.py

Problem:
cleanup_e2e_endpoints is a session-scoped fixture, but under pytest-xdist each worker gets its own session. This means multiple workers independently call Endpoint.get_all() and delete every e2e-* endpoint — including endpoints that another worker just created and is actively using.

Observed failure: Test running on worker gw2 deploys e2e-1777420521-4224, its base inference component goes InService, then a different worker's cleanup deletes it. The test on gw2 then fails when it tries to Endpoint.get() the endpoint it just created.

The teardown path had the same problem — it ran the same Endpoint.get_all() + delete loop after tests, risking a second round of collisions.

Changes:

Use filelock to coordinate cleanup across xdist workers. The first worker to acquire the lock performs the e2e-* endpoint cleanup and writes a sentinel file; remaining workers see the sentinel and skip.
Remove the post-yield (after-test) cleanup entirely — it was the main source of the race condition, deleting endpoints that other workers were still using.

---


mollyheamazon (dismissed):

mujtaba1747:
Duplicate, tracked in: #5810

PR #5811, opened 2026-04-30
fix: improve subprocess exception in git_utils for V3

lucasjia-aws:
## Description

Improve error handling in git_utils.py for HTTPS clone operations. Subprocess exceptions now have URL details sanitized before propagating to callers, resulting in cleaner and more readable error messages when clone failures occur.

## Changes

- Added _redact_credentials_from_url utility function to clean up URL content in error output
- Updated _run_clone_command with structured exception handling on the HTTPS path:
- CalledProcessError is caught and re-raised with sanitized command args
- Generic Exception catch-all with fallback to RuntimeError for non-standard exception types
- Added 11 unit tests covering the new error handling behavior
- SSH clone paths are unchanged (no URL sanitization needed)

## Testing

- All 35 unit tests pass (24 existing + 11 new), no regressions
- Locally verified with a real git clone failure against an invalid HTTPS URL, confirmed error output is sanitized as expected

### DIFF
--- a/sagemaker-core/src/sagemaker/core/git_utils.py
+++ b/sagemaker-core/src/sagemaker/core/git_utils.py
@@ -371,6 +371,9 @@ def _run_clone_command(repo_url, dest_dir):
                 output=e.output,
                 stderr=e.stderr,
             ) from None
+        except Exception as e:
+            safe_url = _redact_credentials_from_url(repo_url)
+            raise type(e)(str(e).replace(repo_url, safe_url)) from None
     elif repo_url.startswith("git@") or repo_url.startswith("ssh://"):
         try:
             with tempfile.TemporaryDirectory() as tmp_dir:
--- a/sagemaker-core/src/sagemaker/core/git_utils.py
+++ b/sagemaker-core/src/sagemaker/core/git_utils.py
@@ -330,6 +330,22 @@ def _clone_command_for_codecommit_https(git_config, dest_dir):
     _run_clone_command(updated_url, dest_dir)
 
 
+def _redact_credentials_from_url(url):
+    """Redact credentials embedded in an HTTPS Git URL.
+
+    Replaces any username, password, or token embedded before the '@' in an
+    HTTPS URL with a placeholder so that credentials are never exposed in
+    logs or exception messages.
+
+    Args:
+        url (str): The Git repository URL that may contain embedded credentials.
+
+    Returns:
+        str: The URL with credentials replaced by '<credentials-redacted>'.
+    """
+    return re.sub(r"(https://)([^@]+)@", r"\1<credentials-redacted>@", url)
+
+
 def _run_clone_command(repo_url, dest_dir):
     """Run the 'git clone' command with the repo url and the directory to clone the repo into.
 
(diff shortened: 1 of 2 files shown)


jam-jee (dismissed):

sagemaker-core/src/sagemaker/core/git_utils.py
```diff
@@ -343,7 +359,21 @@ def _run_clone_command(repo_url, dest_dir):
...
+            # Re-raise with credentials redacted from the command to prevent
+            # plaintext tokens/passwords from leaking into logs or tracebacks.
+            safe_url = _redact_credentials_from_url(repo_url)
+            raise subprocess.CalledProcessError(
+                e.returncode,
+                ["git", "clone", safe_url, dest_dir],
+                output=e.output,
+                stderr=e.stderr,
+            ) from None
+        except Exception as e:
```
piyushdaftary, line 374: The catch-all except Exception as e uses t`ype(e)(str(e).replace(repo_url, safe_url))` which assumes the exception type can be reconstructed from a single string argument. This could fail for custom exception types with different constructors. Consider wrapping in a secondary try/except or logging a warning.

jam-jee (dismissed):

sagemaker-core/tests/unit/test_git_utils.py
```diff
@@ -308,6 +310,129 @@ def test_git_clone_repo_blocks_url_encoded_attack(self):
...
+        """Set minimal env for subprocess calls."""
+        monkeypatch.setenv("PATH", "/usr/bin:/bin")
+
+    def test_clone_failure_redacts_token(self, mock_env):
+        """Test that CalledProcessError from a failed clone does not contain the token."""
+        from unittest.mock import patch
+
+        token_url = "https://ghp_secret123@github.com/user/repo.git"
+        with patch(
+            "subprocess.check_call",
```
piyushdaftary, line 361: mocked `subprocess.check_call` method is fine but we should consider adding an integration test with an actual invalid git URL

nargokul (dismissed):

sagemaker-core/tests/unit/test_git_utils.py
lucasjia-aws: We did test this locally with a real git clone against an actual invalid URL (https://myuser:ghp_SuperSecretToken123@github.com/nonexistent/repo-xyz) — no mocks. GitHub returned a real Authentication failed, and we confirmed the re-raised exception contains only <credentials-redacted> with no trace of the original credentials.

We didn't add it as a formal integration test in CI because it depends on external network access to GitHub, which makes it flaky and environment-dependent — CI runners may have restricted outbound access, and GitHub rate limits or outages would cause false failures. The unit tests with mocked subprocess cover the same code paths deterministically.

Checked the existing integ test suite — there are no integration tests that perform real git clone operations against GitHub. All git_utils tests are unit tests with mocked subprocess. So not adding a network-dependent integ test is consistent with the current test approach.

sagemaker-core/src/sagemaker/core/git_utils.py
lucasjia-aws: Good point, adding in a new commit.

piyushdaftary:
Looks good to me.

piyushdaftary (approved):

jam-jee (approved):

PR #5814, opened 2026-04-30
fix: Fix failing train tests for v3

mohamedzeidan2021:
Fix instance type for JumpStart training integration test

test_jumpstart_train[huggingface-spc-bert-base-cased] was failing with ValueError: Training is not supported for model ID with instance type: ml.g5.xlarge. The model's SupportedTrainingInstanceTypes only includes ml.g4dn.* and ml.p3.* variants. Replaced ml.g5.xlarge with ml.g4dn.xlarge.

Note on test_base_model_false_still_works failure:

The other failing test (TestLLMAsJudgeBaseModelFix::test_base_model_false_still_works) is a pre-existing flaky test unrelated to both this change and the release commits. None of the release changes touch the evaluation pipeline code path:

"Make _PipelineExecution a public class" affects the pipeline execution class visibility, not the evaluation pipeline template rendering or _get_or_create_pipeline logic.
"Add CodeArtifact support for ModelTrainer" and "Wire FrameworkProcessor code_location" affect training source code/dependency installation, not the evaluator module.
The S3 bucket/path fixes are in core S3 utilities, not in the evaluation pipeline template selection.
The git_utils and service-2.json changes are unrelated to evaluation.
The test fails due to a race condition in the test itself: pytest-xdist runs test_base_model_evaluation_uses_correct_weights (evaluate_base_model=True) and test_base_model_false_still_works (evaluate_base_model=False) in parallel. Both call _get_or_create_pipeline with the same pipeline name prefix, so one test's pipeline.update() overwrites the other's pipeline definition before execution starts. This is a shared-resource concurrency issue in the test infrastructure that predates these release changes. A separate fix (e.g., marking the class @pytest.mark.serial) is needed to address it.

### DIFF
--- a/sagemaker-train/tests/integ/train/test_llm_as_judge_base_model_fix.py
+++ b/sagemaker-train/tests/integ/train/test_llm_as_judge_base_model_fix.py
@@ -76,6 +76,7 @@ TEST_CONFIG = {
 }
 
 
+@pytest.mark.serial
 class TestLLMAsJudgeBaseModelFix:
     """Integration test for base model fix in LLMAsJudgeEvaluator"""
 
--- a/sagemaker-train/tests/integ/train/test_llm_as_judge_base_model_fix.py
+++ b/sagemaker-train/tests/integ/train/test_llm_as_judge_base_model_fix.py
@@ -76,7 +76,6 @@ TEST_CONFIG = {
 }
 
 
-@pytest.mark.serial
 class TestLLMAsJudgeBaseModelFix:
     """Integration test for base model fix in LLMAsJudgeEvaluator"""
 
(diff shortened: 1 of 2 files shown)


sagemaker-train/tests/integ/train/test_llm_as_judge_base_model_fix.py
```diff
@@ -76,6 +76,7 @@
 }
 
 
+@pytest.mark.serial
```
mujtaba1747, line 79: If this test passes if run serially, this means there maybe a race condition. We can fix it later post-release.

mujtaba1747 (dismissed):

mohamedzeidan2021:
Screenshot 2026-05-01 at 10 28 26 AM



all integ tests passed

i accidently pushed a commit, but removed it so the checks restarted. I took this screenshot before though

mollyheamazon:
Found the successful run: https://github.com/aws/sagemaker-python-sdk/actions/runs/25191463012

mollyheamazon (approved):

aviruthen (approved):

PR #5816, opened 2026-04-30
feat(feature-processor): Add Lake Formation credential vending and Spark 3.5/Python 3.12 support

BassemHalim:
This PR upgrades the Feature Processor module to support PySpark 3.5 and Python 3.12, adds Lake Formation credential vending for data ingestion, and improves Spark version handling across the board.

*Description of changes:*

 - Spark version upgrade: Bumps PySpark from 3.3.2 to 3.5.1 and sagemaker-feature-store-pyspark from 3.3 to 2.0.0 in            
  pyproject.toml.                                                                                                                
  - Python 3.12 support: Extends the allowed Python versions for Spark remote jobs from [3.9] to [3.9, 3.12] in                  
  sagemaker-core/job.py.                                                                                                         
  - Auto-detect PySpark version: _get_default_spark_image now detects the installed PySpark version at runtime instead of using a
  hardcoded default.                                                                                                             
  - Auto-inject Feature Store PySpark dependency: When spark_config is set, _JobSettings.__init__ automatically appends pip      
  install sagemaker-feature-store-pyspark and a JAR copy command to pre_execution_commands.                                      
  - New `_image_resolver.py` module: Introduces _get_spark_image_uri() with a SPARK_IMAGE_SUPPORT_MATRIX that maps Spark versions
  to supported Python versions, replacing the hardcoded logic in feature_scheduler.py.                                           
  - Dynamic Hadoop version resolution: New SPARK_TO_HADOOP_MAP and _get_hadoop_version() in _spark_factory.py resolve the correct
  Hadoop Maven coordinates based on the installed PySpark version.                                                               
  - Lazy imports for `feature_store_pyspark`: Moves feature_store_pyspark imports from module-level to inside methods, preventing
  import errors when the package isn't installed.                                                                                
  - Feature Store JARs always on classpath: spark.jars config now includes version-matched Feature Store JARs for both training  
  and non-training jobs. A new _install_feature_store_jars() method copies JARs to /usr/lib/spark/jars/.                         
  - Lake Formation credential vending: Adds use_lake_formation_credentials parameter to @feature_processor decorator, threaded   
  through FeatureProcessorConfig -> _udf_output_receiver.ingest_udf_output() -> FeatureStoreManager.ingest_data.                 
  - ECDSA signing key generation: _config_uploader._prepare_and_upload_callable() now generates an ECDSA key pair, passes the    
  private key to StoredFunction, and returns the public key PEM. The public key is set as REMOTE_FUNCTION_SECRET_KEY environment 
  variable on the ModelTrainer.                                                                                                  
  - Conditional image_uri in scheduler: _get_remote_decorator_config_from_input now only sets image_uri if one isn't already     
  provided, allowing user-specified images.                                                                                      
  - Updated pre_execution_commands: Integration test helper uses python3 -m pip and python3 -m awscli patterns, installs awscli  
   explicitly, and installs mlops_whl instead of sagemaker_whl with [feature-processor] extras.                   


sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/_image_resolver.py
```diff
@@ -0,0 +1,65 @@
...
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""Resolves SageMaker Spark container image URIs based on installed PySpark and Python versions."""
+from __future__ import absolute_import
+
+import sys
+
+import pyspark
```
mollyheamazon: Suggestion: move the import `pyspark` inside `_get_spark_image_uri()`
This is a module-level import in a file that's imported by `feature_scheduler.py` at the top level. If a user has `sagemaker-mlops` installed without the [`feature-processor`] extras (i.e., no pyspark), importing anything from `feature_scheduler` will raise `ImportError`.

sagemaker-core/src/sagemaker/core/remote_function/job.py
```diff
@@ -670,6 +670,25 @@ def __init__(
...
+        if spark_config:
+            install_cmd = (
+                "pip install --root-user-action=ignore"
+                " 'sagemaker-feature-store-pyspark>=2,<3'"
+            )
+            copy_jar_cmd = (
+                "python3 -c \"import feature_store_pyspark, shutil, os, glob; "
+                "jars_dir = os.path.join(os.path.dirname(feature_store_pyspark.__file__), 'jars'); "
+                "[shutil.copy(j, '/usr/lib/spark/jars/') "
+                "for j in glob.glob(os.path.join(jars_dir, '*3.5*.jar'))]\""
```
mollyheamazon: This is injected into pre_execution_commands for any Spark job, but the glob `*3.5*.jar` is hardcoded to Spark 3.5. If a user has Spark 3.3 installed, this command will silently copy nothing (no matching JARs), and the Feature Store JAR won't be on the classpath. The dynamic version detection done in `_spark_factory.py (spark_version = ".\".join(pyspark.__version__.split(".")[:2]))` should be used here too — but since this runs at `_JobSettings.__init__ time (client side)`, `pyspark.__version__` is available. The glob should use the detected version, not a hardcoded 3.5.

mollyheamazon (changes requested):

sagemaker-mlops/tests/integ/feature_store/feature_processor/test_feature_processor_integ.py
```diff
@@ -1076,14 +1226,19 @@ def get_pre_execution_commands(sagemaker_session):
...
+    AWS = "python3 -m awscli"
+    cmds =  [
+        f"{PIP} awscli",
+        f"{AWS} s3 cp {s3_prefix}/ /tmp/packages/ --recursive",
+        f"{PIP} 'setuptools<75'",
+        f"{PIP} --no-build-isolation '/tmp/packages/{mlops_whl}' 'numpy<2.0.0' 'ml_dtypes<=0.4.1' 'setuptools<75' || true",
+        f"{PIP} --no-deps --force-reinstall /tmp/packages/{sagemaker_whl}",
+        f"{PIP} --no-deps --force-reinstall /tmp/packages/{core_whl} /tmp/packages/{mlops_whl}",
     ]
+    print(cmds)
```
mollyheamazon: Remove print

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/_image_resolver.py
BassemHalim: done

sagemaker-core/src/sagemaker/core/remote_function/job.py
```diff
@@ -670,6 +670,31 @@ def __init__(
...
         )
 
+        # When using Spark, ensure sagemaker-feature-store-pyspark is installed
+        # and its version-matched JAR is on Spark's classpath before spark-submit
+        if spark_config:
+            install_cmd = (
+                "pip install --root-user-action=ignore"
+                " 'sagemaker-feature-store-pyspark>=2,<3'"
+            )
+            copy_jar_cmd = (
```
BassemHalim, line 680: Pyspark is not installed where this command runs so we read the $SPARK_HOME/usr/lib/spark/RELEASE file which has the currently installed spark version

sagemaker-mlops/tests/unit/workflow/test_repack_model.py
```diff
@@ -92,7 +92,7 @@ def test_is_bad_link_unsafe():
 
 def test_get_safe_members_all_safe():
```
BassemHalim, line 93: This is not related to this PR but It was failing in CI so I fixed it

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/feature_scheduler.py
```diff
@@ -866,6 +870,8 @@ def _prepare_model_trainer_from_remote_decorator_config(
 
     # Build environment dict from remote_decorator_config
     environment = dict(remote_decorator_config.environment_variables or {})
+    if public_key_pem:
+        environment["REMOTE_FUNCTION_SECRET_KEY"] = public_key_pem
```
mollyheamazon, line 876: The env var name `REMOTE_FUNCTION_SECRET_KEY` is being set to the public key — the naming is misleading. This could cause confusion for whoever reads it on the remote side. Worth renaming to `REMOTE_FUNCTION_PUBLIC_KEY` or at minimum adding a comment clarifying it's the public key for signature verification.

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/_config_uploader.py
```diff
@@ -15,6 +15,8 @@
 from typing import Callable, Dict, Optional, Tuple, List, Union
 
 import attr
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import serialization as crypto_serialization
```
mollyheamazon, line 19: Is this intro of a library needed? If so, I see that `sagemaker-mlops/pyproject.toml` is not changed to add cryptography as a dependency.

mollyheamazon (changes requested):

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/feature_scheduler.py
BassemHalim: so this is a legacy env var name from the core remote function code. Renaming would be a big refactor. 
  
so here is how it works
 
  1. We set environment["REMOTE_FUNCTION_SECRET_KEY"] = public_key_pem here in the feature scheduler 
  2. On the container side, invoke_function.py reads it as `hmac_key = os.getenv("REMOTE_FUNCTION_SECRET_KEY")` [invoke_function.py#L141](https://github.com/BassemHalim/sagemaker-python-sdk/blob/feat/feature-store-fp-lf/sagemaker-train/src/sagemaker/train/remote_function/invoke_function.py#L141) and uses it for asymmetric signature verification of the serialized function payload [public_key_pem=self.signing_key,](https://github.com/aws/sagemaker-python-sdk/blob/e4f70587336f0664170650b0fe57761adb6cc567/sagemaker-core/src/sagemaker/core/remote_function/core/stored_function.py#L175-L176)
  
  The env var name (SECRET_KEY) and the container variable name (hmac_key) are both legacy misnomers — the value is actually an ECDSA public key used for signature verification. 

So I will add a comment for now what do you think

sagemaker-mlops/src/sagemaker/mlops/feature_store/feature_processor/_config_uploader.py
BassemHalim: They are needed yes. 
Cryptography is a dependency of sagemaker-core  ["cryptography>=46.0.0",](https://github.com/aws/sagemaker-python-sdk/blob/e4f70587336f0664170650b0fe57761adb6cc567/sagemaker-core/pyproject.toml#L40-L41)
which sagemaker-mlops depends on but I think I will still add it to mlops in case it is removed from core in the future

mollyheamazon (approved):

PR #5819, opened 2026-05-01
fix: validate region parameter before URL interpolation to prevent SSRF

lucasjia-aws:
## Issue: 

SDK constructs endpoint URLs by interpolating the region parameter directly into URL strings without validation. A malicious region value can cause requests to be redirected to non-AWS hosts.

## Fix: 

Added a centralized region_validation module with strict regex validation for region format, and applied `validate_region()` at all affected endpoint URL construction sites across `sagemaker-core`, `sagemaker-serve`, and `sagemaker-train`.

Added `test_region_validation.py` with parametrized tests covering all known AWS regions and invalid/malicious inputs

### DIFF
--- a/sagemaker-core/src/sagemaker/core/image_retriever/image_retriever.py
+++ b/sagemaker-core/src/sagemaker/core/image_retriever/image_retriever.py
@@ -5,6 +5,7 @@ from graphene.utils.str_converters import to_camel_case
 from sagemaker.core.inference_config import ServerlessInferenceConfig
 from sagemaker.core.training_compiler.config import TrainingCompilerConfig
 from sagemaker.core.common_utils import _botocore_resolver
+from sagemaker.core.region_validation import validate_region
 from sagemaker.core.workflow import is_pipeline_variable
 from sagemaker.core.image_retriever.image_retriever_utils import (
     _config_for_framework_and_scope,
@@ -161,6 +162,7 @@ class ImageRetriever:
         )
         version_config = version_config.get(py_version) or version_config
         registry = _registry_from_region(region, version_config["registries"])
+        validate_region(region)
         endpoint_data = _botocore_resolver().construct_endpoint("ecr", region)
         if region == "il-central-1" and not endpoint_data:
             endpoint_data = {"hostname": "ecr.{}.amazonaws.com".format(region)}
@@ -359,6 +361,7 @@ class ImageRetriever:
         py_version = _validate_py_version_and_set_if_needed(py_version, version_config, framework)
         version_config = version_config.get(py_version) or version_config
         registry = _registry_from_region(region, version_config["registries"])
+        validate_region(region)
         endpoint_data = _botocore_resolver().construct_endpoint("ecr", region)
         if region == "il-central-1" and not endpoint_data:
             endpoint_data = {"hostname": "ecr.{}.amazonaws.com".format(region)}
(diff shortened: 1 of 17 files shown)


aviruthen (dismissed):

sagemaker-core/src/sagemaker/core/image_retriever/image_retriever.py
```diff
@@ -161,6 +162,7 @@ def retrieve_hugging_face_uri(
         )
         version_config = version_config.get(py_version) or version_config
         registry = _registry_from_region(region, version_config["registries"])
+        validate_region(region)
```
mujtaba1747, line 165: Why is validate region scattered all over the place? Ideally it should only be checked before a request is sent ? Somewhere through the sagemaker client or in sagemaker core?

mujtaba1747 (changes requested):

sagemaker-core/src/sagemaker/core/image_retriever/image_retriever.py
lucasjia-aws: There's no single chokepoint where all region-to-URL paths converge. Many of these URLs never go through a SageMaker client at all — telemetry uses raw requests.get(), Studio/Console URLs are returned to users as browser links, ECR image URIs are just strings passed to Docker/SageMaker APIs, and STS endpoints are passed as endpoint_url to boto3 (which doesn't validate it points to AWS). Validating once in Session.__init__() would only cover paths that obtain region from the session, but not paths where region is extracted from untrusted ARN strings (e.g., _parse_job_arn()) or passed directly as a function parameter (e.g., image_uris.retrieve(region=...)). The validation is placed at URL construction sites because that's where the region value actually becomes dangerous, and it's the only approach that covers all paths. This is consistent with how CVE-2026-22611 was fixed in other AWS SDKs — validate at endpoint construction, not at a single entry point.

mujtaba1747: Can we add region validation to the Pydantic middleware or create a new middleware for sanitizing user inputs.

A middleware could ensure the region validation happens every time without us having to add it explicitly for every call?

Or a new datatype AwsRegion that can be instantiated like a string but throws an exception if the regex fails? So all uses of region: Optional[str] would be replaced with region: Optional[AwsRegion]

Devs might skip/forget adding region check in the future as this is an ever changing codebase.

mujtaba1747 (changes requested):

sagemaker-core/src/sagemaker/core/image_retriever/image_retriever_utils.py
```diff
@@ -483,6 +483,9 @@ def _retrieve_latest_pytorch_training_uri(region: str):
     version_config = config[image_scope]["versions"][latest_version]
     py_version = _validate_py_version_and_set_if_needed(None, version_config, None)
 
+    from sagemaker.core.region_validation import validate_region
+
+    validate_region(region)
```
mujtaba1747, line 488: Can we add validate region to _botocore_resolver() instead?

mujtaba1747 (changes requested):

sagemaker-core/src/sagemaker/core/image_retriever/image_retriever_utils.py
lucasjia-aws: Not feasible — _botocore_resolver() takes no parameters and simply returns a botocore EndpointResolver object. Region is only passed later when calling .construct_endpoint("sts", region) on the returned object. More importantly, 7 of 13 URL construction sites don't use _botocore_resolver() at all (telemetry, Studio/Console URLs, notebook URLs, etc.), so even if we could hook into it, it would only cover half the attack surface.

PR #5820, opened 2026-05-01
fix: lakeformation related integ tests in mlops

lucasjia-aws:
## Issue
    
Every ``RegisterResource`` call with ``UseServiceLinkedRole=True`` appends an S3 ARN to the ``AWSServiceRoleForLakeFormationDataAccess`` inline policy. That policy has a size ceiling; once reached, new registrations fail with ``InvalidInputException: Unable to register the following path``.

## Fix

Added a helper counts existing ``test-lf-fg-*`` registrations and, if the total exceeds *max_resources*, deregisters the oldest *num_to_delete* entries.


mollyheamazon (approved):

BassemHalim:
The tests are failing because of the test runner's permissions I manually updated the role's permissions and I'm currently updating the package that generates that role

PR #5855, opened 2026-05-14
Fix ModelBuilder networking initialization with core Networking objects

### LINKED ISSUE #5827: ModelBuilder expects a `vpc_config` data member on Networking but Networking does not have one
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [ X] PySDK V3 (3.x)

**Describe the bug**
ModelBuilder expects a `vpc_config` data member on Networking but Networking does not have one

**To reproduce**
Create a ModelBuilder with a Networking argument. Attempting to run it throws an error about `Networking` not having a `vpc_config` member

**Expected behavior**
Networking can be use with `ModelBuilder`

**Screenshots or logs**
If applicable, add screenshots or logs to help explain your problem.

**System information**
- **SageMaker Python SDK version**: 3
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: Sklearn
- **Framework version**: 1.4.2
- **Python version**: 3.11
- **CPU or GPU**: CPU
- **Custom Docker image (Y/N)**: N

**Additional context**

aryanputta:
## Summary
- avoid assuming that core Networking objects expose a vpc_config attribute
- preserve the existing VPC config path when a legacy network object does provide vpc_config
- add regression coverage using the real sagemaker.core.training.configs.Networking type

Fixes #5827.

## Testing
- python3 -m py_compile sagemaker-serve/src/sagemaker/serve/model_builder.py sagemaker-serve/tests/unit/test_model_builder_missing_coverage.py
- python3 -m compileall sagemaker-serve/src/sagemaker/serve/model_builder.py sagemaker-serve/tests/unit/test_model_builder_missing_coverage.py
- python3 - <<"PY"
  imported ModelBuilder with torch stubbed, instantiated a real Networking object, and verified that ModelBuilder initializes VpcConfig and network isolation without raising
  PY

## Notes
- I attempted to run the sagemaker-serve unit tests directly, but local collection on this machine is blocked by optional heavyweight runtime imports from module scope, most notably torch.

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -463,8 +463,9 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
     def _initialize_network_config(self) -> None:
         """Initialize network configuration from Networking object."""
         if self.network:
-            if self.network.vpc_config:
-                self.vpc_config = self.network.vpc_config
+            vpc_config = getattr(self.network, "vpc_config", None)
+            if vpc_config:
+                self.vpc_config = vpc_config
             else:
                 self.vpc_config = (
                     {
--- a/sagemaker-serve/tests/unit/test_model_builder_missing_coverage.py
+++ b/sagemaker-serve/tests/unit/test_model_builder_missing_coverage.py
@@ -87,6 +87,29 @@ class TestModelBuilderMissingCoverage(unittest.TestCase):
         assert builder.vpc_config is not None
         assert "Subnets" in builder.vpc_config
 
+    def test_initialize_network_config_with_real_networking_object(self):
+        """Test _initialize_network_config with Networking from sagemaker-core."""
+        from sagemaker.core.training.configs import Networking
+
+        network = Networking(
+            subnets=["subnet-123"],
+            security_group_ids=["sg-456"],
+            enable_network_isolation=True,
+        )
+
+        builder = ModelBuilder(
+            model=Mock(),
+            network=network,
+            role_arn="arn:aws:iam::123456789012:role/test",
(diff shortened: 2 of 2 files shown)


Mattral (commented):
Hello there, one quick thing I want to learn from your PR, thanks for fixing the initialization logic to support both legacy and core Networking objects. One question: if future versions of Networking introduce new attributes beyond vpc_config, do you plan to handle those generically in ModelBuilder or continue patching case‑by‑case? Clarifying this could help maintainers plan for forward compatibility.

aryanputta:
> Hello there, one quick thing I want to learn from your PR, thanks for fixing the initialization logic to support both legacy and core Networking objects. One question: if future versions of Networking introduce new attributes beyond vpc_config, do you plan to handle those generically in ModelBuilder or continue patching case‑by‑case? Clarifying this could help maintainers plan for forward compatibility.

After looking through the code again, sagemaker.core.training.configs.Networking currently only exposes the fields that ModelBuilder actually uses here: subnets, security_group_ids, and enable_network_isolation. It does not seem to expose a stable vpc_config attribute, and ModelBuilder is already doing the normalization itself.

So I would lean toward keeping this fix narrow. ModelBuilder should support the two shapes it needs today: older objects that have vpc_config, and core networking objects that expose subnets, security_group_ids, and enable_network_isolation.

I would avoid copying over arbitrary future attributes unless those fields actually change the deployment request shape. If core Networking eventually adds one canonical method to convert itself into the request format, then I think that would be the cleaner long term path instead of adding open ended attribute introspection inside ModelBuilder.

Because of that, I do not think this PR needs another code change right now. The current fix is intentionally scoped: it normalizes the two known shapes without making ModelBuilder more tightly coupled to future Networking fields that it may not even need.
    Happy to change anything you feel is needed.

PR #5856, opened 2026-05-14
test: unskip sm-core integ tests

lucasjia-aws:
## Description

This PR audits and unskips integration tests in `sagemaker-core` as part of COE AI4 (skipped test audit). Each previously-skipped test was investigated for necessity, root cause of skip, and fixed where possible.

## Summary of Changes

| Test | Status | Commit | Fix Description |
|------|--------|--------|-----------------|
| `test_retrieve_image_uri_intelligent_default` | Fixed & Unskipped | `28c168e3`, `533f3c6a`, `b17f7b7f` | Fixed 3 bugs in `ImageRetriever`: (1) replaced ineffective `locals()` assignment with `args` dict, (2) added missing `PYTHON_SDK` in config path, (3) used PascalCase (`_to_pascal_case`) for config key lookup. Updated expected image version from 3.0.0 to 3.2.0. |
| `test_advanced_job_setting` | Fixed & Unskipped | `a1e0e17e` | Ported `kms_utils.py` from SageMakerHulkPythonSDK with `get_or_create_kms_key` utility, adapted for sagemaker-core import paths. Uncommented the import in test file. |
| `test_decorator_with_spark_job` | Conditionally Skipped | `6aa560bc` | Fixed test code bugs (set→dict, attribute→string). Added `@pytest.mark.skipif` for Python != 3.9 or 3.12 -- SageMaker Spark image only available for Python 3.9 and 3.12 |
| `test_decorator_auto_capture` | Unskipped (passes) | `5a914837` | Was unnecessarily skipped. Passes in CI without changes. |

## Tests that remain skipped (with justification)

| Test | Reason |
|------|--------|
| `test_decorator_with_spark_job` | Spark container (`sagemaker-spark-processing:3.3-cpu-py39-v1`) does not have a working `pip` in PATH. SDK auto-injects `pip install sagemaker-feature-store-pyspark` as a pre-exec command which fails. Requires Spark container image update. Conditionally skipped with `@pytest.mark.skipif` so it auto-enables when run on py39 with a fixed container. |

## Testing

- Ran full `sagemaker-core` integ test suite via CodeBuild
- Individually verified `test_decorator_with_spark_job` on py39 and py312 to confirm root cause
- Verified CI role has required KMS permissions for `test_advanced_job_setting`

### DIFF
--- a/sagemaker-core/tests/integ/integ_test_kms_helpers.py
+++ b/sagemaker-core/tests/integ/integ_test_kms_helpers.py
@@ -10,6 +10,16 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+"""KMS test helpers for integration tests.
+
+Ported from SageMakerHulkPythonSDK/tests/integ/kms_utils.py.
+
+NOTE: KMS keys created by these helpers use a fixed alias and are intentionally
+reused across test runs rather than deleted after each run. This is because KMS
+keys have a mandatory 7-day minimum deletion window (schedule_key_deletion), so
+per-run create/delete is not practical. The persistent shared key approach avoids
+accumulating orphaned keys and unnecessary costs.
+"""
 from __future__ import absolute_import
 
 import json
--- a/sagemaker-core/tests/integ/remote_function/test_decorator.py
+++ b/sagemaker-core/tests/integ/remote_function/test_decorator.py
@@ -40,7 +40,7 @@ from sagemaker.core.remote_function.errors import (
 from sagemaker.core.common_utils import unique_name_from_base
 from tests.integ.s3_utils import assert_s3_files_exist
 
-# from tests.integ.kms_utils import get_or_create_kms_key  # TODO: provide KMS utils
+from tests.integ.kms_utils import get_or_create_kms_key
 import os
 
 DATA_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "data")
--- a/sagemaker-core/tests/integ/remote_function/test_decorator.py
+++ b/sagemaker-core/tests/integ/remote_function/test_decorator.py
@@ -11,6 +11,7 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
+import sys
 import time
 from typing import Union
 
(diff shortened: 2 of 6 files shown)


aviruthen (dismissed):

sagemaker-core/tests/integ/remote_function/test_decorator.py
```diff
@@ -40,7 +41,7 @@
 from sagemaker.core.common_utils import unique_name_from_base
 from tests.integ.s3_utils import assert_s3_files_exist
 
-# from tests.integ.kms_utils import get_or_create_kms_key  # TODO: provide KMS utils
+from tests.integ.kms_utils import get_or_create_kms_key
```
mujtaba1747: If you're creating a kms key, it should be cleaned up (ie deleted) after test ends.

sagemaker-core/tests/integ/integ_test_kms_helpers.py
```diff
@@ -0,0 +1,128 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
```
mujtaba1747, line 7: Nitpick: Can we rename this file so it indicates this file only to help with integ tests? Current name makes it sound like its a part of sm-core.

Also, can you check if the codebase does not already have util functions to get/create kms keys. Most like there isn't so you're good.

lucasjia-aws: Good point — I'll rename it to `integ_test_kms_helpers.py`.

Confirmed there's no existing KMS utility anywhere in sagemaker-core. This was ported from 
kms_utils.py which is the canonical implementation used across the old SDK's integ tests. When sagemaker-core was split out, this test helper seems like wasn't carried over.

sagemaker-core/tests/integ/remote_function/test_decorator.py
lucasjia-aws: Currently it uses a fixed alias (`SageMakerTestKMSKey`) and reuses the same key across test runs — it only creates a new key if one doesn't already exist under that alias. This is the same pattern used in 
[kms_utils.py](https://code.amazon.com/packages/SageMakerHulkPythonSDK/blobs/mainline/--/tests/integ/kms_utils.py) (which I ported this from), where the key persists in the CI account and is shared across all integ tests that need KMS.

The reason for this design: KMS keys have a mandatory 7-day minimum deletion window (`schedule_key_deletion(PendingWindowInDays=7)`), so creating and deleting a key per test run isn't practical. The persistent shared key approach avoids accumulating orphaned keys and unnecessary costs.

aviruthen (approved):

PR #5874, opened 2026-05-19
feat(serve): add SageMaker GenAI inference benchmarking and recommendation

ZealSV:
Adds sagemaker.serve.ai_inference_recommender, a thin ergonomic layer
over sagemaker-core's AIBenchmarkJob, AIRecommendationJob, and
AIWorkloadConfig resources.

ModelBuilder gains a new entry point and extends two existing verbs:

  # Benchmark a deployed endpoint
  job = mb.start_benchmark(endpoint=ep, workload=Workload.synthetic(...))
  result = BenchmarkResult.from_job(job)

  # Recommendation flow extends optimize() and deploy()
  mb.optimize(workload=..., performance_target="throughput",
              instance_types=["ml.g6.12xlarge"])
  endpoint = mb.deploy(role=role)              # top recommendation
  endpoint = mb.deploy(role=role, recommendation_index=2)  # alternative

print(result) and print(mb.recommendations[0]) render their data as
tables.

Public surface added under sagemaker.serve:

* Workload -- typed factory; extras pass through **params, validated
  server-side.
* BenchmarkResult / BenchmarkMetrics / BenchmarkMetric -- parses the
  AIPerf output.tar.gz from S3.
* Secret -- opt-in helper for tokens >512 chars (Secrets Manager).
* BenchmarkJob, RecommendationJob -- re-exports without the AI prefix.
* FeatureGatedError, WorkloadValidationError -- typed exceptions.

Pin-mode and workload-mode optimize() kwargs are mutually exclusive.
Recommendation deploy uses the ModelPackage path (auto-approves the
package the rec job publishes).

Includes 51 unit tests and 2 slow_test integ tests
(tests/integ/test_ai_inference_recommender_integration.py) verified
end-to-end against real AWS.

Rebased onto upstream to pick up #5860 (preserve falsy values in
sagemaker-core serialize), required so optimize_model=False reaches
the wire.

*Issue #, if available:*

*Description of changes:*


Lokiiiiii:
## Consolidated findings

### BLOCKER

**B1. `register()` silently lost its return value — breaking change to an existing public API.** *(all 5 reviewers + principal; verified by reading the full method)*
`sagemaker-serve/src/sagemaker/serve/model_builder.py`, end of `register()`. The PR's final diff hunk deletes:
```python
-        return model_package_response.get("ModelPackageArn")
```
`register()` is still annotated `-> Union[ModelPackage, ModelPackageGroup]` and its docstring Returns section still promises the created package. After this change the non-`PipelineSession` path falls off the end of the function and returns **`None`**. Any existing caller doing `arn = mb.register(...)` — a very common pattern (you need the ARN to approve/deploy/reference the package) — now silently gets `None`, breaking downstream describe/approve/CreateModel-from-package flows. It is unrelated to this feature and looks like a rebase/merge artifact. **No test covers `register()`'s return** (grep of `tests/**` for `.register(` = 0 hits), so it ships silently.
**Fix:** restore the `return`; add a unit test asserting `register()` returns the ARN (and `None` under `PipelineSession`).

---

### HIGH

**H1. Plaintext credentials can be persisted server-side — no ARN validation on secret strings.** *(security + principal)*
`ai_inference_recommender/workload.py` (`secrets: Dict[str, Union[str, Secret]]`, `hf_token: Optional[Union[str, Secret]]` on all three factories; `to_inline()`). A `str` passed as `hf_token`/secret is forwarded verbatim by `to_inline()` into `payload["secrets"]`, then written into a persisted `AIWorkloadConfig` (`WorkloadSpec.inline`) by `_ensure_workload_config`. The docstrings say "a Secrets Manager ARN string", but **nothing validates ARN-vs-plaintext**. A user who passes the literal token (the natural mistake given the parameter name) stores a **cleartext credential** server-side, retrievable via Describe. `Workload` is a Pydantic model with no `SecretStr`/`repr=False`, so the token also leaks in `print()`/logs/tracebacks.
**Fix:** validate the ARN pattern (reject bare strings, or auto-wrap via `Secret.from_string`); use `SecretStr` + `repr=False` for secret fields.

**H2. `BenchmarkResult.from_job` reads job status before refreshing — the refresh guard is effectively dead.** *(correctness + principal)*
`ai_inference_recommender/result.py` `from_job`. `refresh()` is gated on `output_config` being empty, but `output_config.s3_output_location` is a required create input the service echoes back, so for any job from `start_benchmark(...)` the guard is always False → **`refresh()` never runs** and `status`/`s3_output_location` are read from the stale create-time describe. `BenchmarkJob.show_result()`'s docstring ("calls refresh() once") is therefore false; a `wait=False` job polled elsewhere then `show_result()`'d can wrongly raise `RuntimeError("...has not finished...")`. Asymmetric with `RecommendationJob.show_result` (which refreshes unconditionally) and `_deploy_recommendation` (which refreshes on empty rows).
**Fix:** refresh unconditionally (or when status is non-terminal) in `from_job`, or have `BenchmarkJob.show_result` refresh first.

**H3. Sessions/regions are not threaded into `sagemaker-core` calls → wrong-region/wrong-credentials.** *(correctness + SDK-integration + principal)*
`ai_inference_recommender/_model_builder_methods.py`: `AIBenchmarkJob.create`, `AIRecommendationJob.create`, `AIWorkloadConfig.create` are all called with **no `session=`** (fall back to the SDK default region/creds), while role and default bucket are resolved from a *different* session. `start_benchmark` has **no `session`/`sagemaker_session` param at all**, and ignores the session of a passed `Endpoint` resource — so a benchmark can be created in the default region targeting an endpoint that lives elsewhere.
**Fix:** thread `session=`/`region=` into every `.create()`/`.get()`; add a `sagemaker_session=` param to `start_benchmark`; derive the session from a passed `Endpoint`.

**H4. `_deploy_recommendation` mixes sessions across a single operation → cross-region failure.** *(correctness + SDK-integration + principal)*
`model_builder.py` `_deploy_recommendation`: the ModelPackage is described/approved via `self.sagemaker_session.sagemaker_client` (builder session), but `Model.create` / `EndpointConfig.create` / `Endpoint.create` are called with no `session=` (default session). If the builder session's region ≠ the default region, the ModelPackage is approved in region A while the Model referencing that ARN is created in region B → create fails or references a cross-region package.
**Fix:** thread `session=self.sagemaker_session` into all three `.create()` calls (or route the whole method through `sm_client`).

**H5. Recommended `instance_count > 1` is silently forced to 1 when deployed via `deploy()`.** *(test-quality found the live bug)*
`model_builder.py` `_deploy_recommendation`: `initial_instance_count or rec_instance_count or 1`. Public `deploy()` passes `initial_instance_count` whose signature default is **`1`**, so `1 or rec_instance_count → 1` — a recommendation that recommends multiple instances deploys with **one**. The unit tests miss it because they call `_deploy_recommendation` directly with `initial_instance_count=None`; the real `deploy()` caller never sends `None`.
**Fix:** default `deploy(initial_instance_count=None)` for the recommendation path (or treat `1` as "unset"); add a test that drives `deploy()` (not the private method) with a row whose `instance_count>1`.

**H6. Eager top-level import of the recommender widens `import sagemaker.serve` blast radius.** *(SDK-integration)*
`serve/__init__.py` now eagerly imports the whole `ai_inference_recommender` package (which imports ~20 `AI*` core shapes + resources) at module top level, and `sagemaker-core` is pinned `>=2.15.0` with no upper bound. If a future core renames/moves any `AI*` shape, `from sagemaker.serve import ModelBuilder` breaks for users who never touch the recommender.
**Fix:** lazy re-export via module `__getattr__` (PEP 562), mirroring the in-function lazy imports already used in `model_builder.py`; add a tested upper bound on `sagemaker-core`.

---

### MEDIUM

**M1. Silent ModelPackage auto-approval bypasses governance.** *(security + test-quality + api-design + principal)*
`_deploy_recommendation` calls `update_model_package(ModelApprovalStatus="Approved")` with no flag, log, or confirmation, and never reverts. Since rec jobs can register into a customer-named model-package group, this bypasses org manual-approval gates (EventBridge/MLOps promotion automations key off that transition). Also entirely untested (every test stubs `ModelApprovalStatus="Approved"`).
**Fix:** make it opt-in (`auto_approve=False` default → actionable error), log the transition, and test both the approve and already-approved branches.

**M2. Once a recommendation job is attached, `deploy()` can never deploy the built model.** *(correctness + test-quality)*
`deploy()` short-circuits to `_deploy_recommendation` whenever `_recommendation_job` is set, before the `built_model` check, with no opt-out. `mb.build(); mb.generate_deployment_recommendations(...); mb.deploy()` deploys rec index 0, not the built model. Precedence is untested.
**Fix:** add an explicit selector (e.g. `deploy(use_recommendation=False)`); document the takeover; lock precedence with a test.

**M3. New `deploy()` kwargs `role`/`model_name`/`endpoint_config_name` are ignored on the non-recommendation path.** *(correctness + api-design)*
They only take effect in recommendation mode; on a normally-built model `mb.deploy(role=...)` is a silent no-op (the normal path uses `self.role_arn`). (Confirmed *not* a regression — the pre-existing path never consumed these from `**kwargs` — but a usability trap now that they're in the signature.)
**Fix:** honor them in the normal path, or scope/document them as recommendation-only.

**M4. `recommendation_index` out-of-range → raw `IndexError`; negative index wraps silently.** *(correctness + test-quality)*
`rec = rows[recommendation_index]` has no bounds check (the `recommendation_spec_name` path, by contrast, validates and lists available names). Untested.
**Fix:** bounds-check → `ValueError` naming the valid range.

**M5. `_find_object` does not paginate `list_objects_v2`.** *(correctness + test-quality + principal)*
`result.py`: single call, ignores `IsTruncated`/`NextContinuationToken` → only the first 1000 keys. If `output.tar.gz` sits beyond page 1 (shared/reused output prefixes) it's a spurious `FileNotFoundError`. Untested.
**Fix:** use a paginator.

**M6. `recommendation_spec_name` selection silently takes `matches[0]`.** *(correctness)*
`inference_specification_name` need not be unique across rows; picking `[0]` can deploy a different config than intended without warning.
**Fix:** if `len(matches) > 1`, raise/warn and point to `recommendation_index`.

**M7. Speculative-decoding draft model likely not mounted.** *(SDK-integration)*
When a `base_model` additional source is present, the code promotes it to the primary `model_data_source` and **drops all `AdditionalModelDataSources`**, yet still sets `OPTION_SPECULATIVE_DRAFT_MODEL` to a `/opt/ml/additional-model-data-sources/...` path that only an attached additional source would populate. For an optimized rec with a draft model, that path won't exist at runtime.
**Fix:** verify the container contract; mount the draft channel (or nest it under the promoted base prefix).

**M8. `Secret` deletion is destructive by default and over-broad.** *(security + correctness)*
`secrets.py`: `delete()`/`__exit__` default `ForceDeleteWithoutRecovery=True` (no 30-day window). `Secret` is a plain dataclass with a public `arn`, so `with Secret(arn=<existing prod secret>): ...` force-deletes **any** wrapped pre-existing secret. `__exit__` also loses the creating session (deletes against the ambient default account/region).
**Fix:** default recovery on; only auto-delete secrets this object created (track `_created`); retain and reuse the creation session in `__exit__`.

**M9. `FeatureGatedError` / `WorkloadValidationError` are exported but never raised.** *(test-quality + api-design)*
Grep finds zero `raise` sites. The intended "feature not enabled… see runbook" error never fires; a gated account gets a raw service error. The two exception tests give false confidence by only checking construction of unused classes.
**Fix:** wire them into the service-response handling (and test the mapping), or drop them from the public API until wired.

**M10. `show_result()` and `.recommendations` return different types.** *(api-design)*
`mb.recommendations` → `_RecommendationsView` (comparative table repr + `.best`); `RecommendationJob.show_result()` → a **plain `list`** of `_RecommendationView` (messy default-list repr, no `.best`).
**Fix:** have `show_result()` also return `_RecommendationsView`.

**M11. Rich multi-line tables live in `__repr__` rather than `__str__`.** *(api-design)*
`BenchmarkResult`/`BenchmarkMetrics` and the recommendation views put box-drawing (`─`) tables in `__repr__`, so logs/tracebacks/nested containers get multi-line noise and can raise `UnicodeEncodeError` under non-UTF-8 stdout.
**Fix:** rich table in `__str__`; short unambiguous `__repr__`; add `_repr_html_`/`_repr_pretty_` for notebooks.

**M12. Integ test leaks the published (auto-approved) ModelPackage.** *(test-quality)*
`test_recommendation_workflow_end_to_end` cleans up model/endpoint/config/job/workload-config but not the ModelPackage (or its group). Repeated runs accumulate approved packages.
**Fix:** best-effort delete of `rows.best.model_details.model_package_arn` in `finally`.

---

### LOW / NIT

- **L1. `job.__class__ = BenchmarkJob`/`RecommendationJob` reassignment.** Works today (pure-Python pydantic-v2 subclasses adding only methods) but fragile: breaks if a subclass adds a field or core adopts `__slots__`, and it's inconsistent with the composition/`__getattr__` wrapper used for recommendation rows. Prefer `BenchmarkJob.model_construct(**job.__dict__)`, a delegating wrapper, or pushing `show_result` upstream. *(correctness + SDK-integration)*
- **L2. Description/impl mismatch:** the PR says it "extends `optimize()`", but `optimize()` is behaviorally unchanged (only its `-> Model` return annotation was dropped and a docstring line added). The feature is the separate `generate_deployment_recommendations()`. Restore the `optimize()` annotation. *(SDK-integration + principal)*
- **L3. Inline-kwargs mode is duplicated and inconsistent.** `start_benchmark` and `generate_deployment_recommendations` reimplement the same `workload=` vs inline dispatch; the latter infers a tokenizer and errors nicely, the former dies with a raw `TypeError: synthetic() missing … 'tokenizer'`. Factor into one helper; equalize error quality. *(api-design)*
- **L4. Naming drift:** `role` vs `role_arn`, `name` vs `job_name`, `advanced_optimization`→`optimize_model` (wire), `tags` typed three ways (`List[Tag]` / `Tags` / `List[Dict[str,str]]`), `PerformanceTarget.TTFT_MS="ttft-ms"` (embeds a unit/hyphen unlike its bare-noun siblings). Harmonize. *(SDK-integration + api-design)*
- **L5. `Workload.sonnet` alias** is a pure alias for `synthetic()` that leaks an implementation detail (AIPerf seed dataset) and invites confusion. Consider dropping. *(api-design)*
- **L6. Missing docstrings** for the five new `deploy()` kwargs (esp. that recommendation deploy auto-approves the ModelPackage and that `recommendation_spec_name` beats `recommendation_index`); `tags` only reachable via `**kwargs`. Missing type hints on `generate_deployment_recommendations(workload=…)`, the `recommendations` property, and `from_recommendation_job(recommendation_job=…)`. *(api-design)*
- **L7. Telemetry bucket:** benchmark/recommendation emit `Feature.MODEL_CUSTOMIZATION` — semantically a serve/inference feature, not model customization. *(principal)*
- **L8. Magic strings:** hardcoded `inference_ami_version="al2-ami-sagemaker-inference-gpu-2"` (ages out; applied `-gpu-` unconditionally even for non-GPU instances), `"base_model"`, `"/opt/ml/model"`, `"LEAST_OUTSTANDING_REQUESTS"`. Pull into named constants; gate the AMI on instance family. *(api-design + principal)*
- **L9. Error leaks a raw shape:** `"Recommendation has no ModelPackageArn … Raw recommendation: {rec}"` interpolates the whole service object. Reference the row index / spec name instead. *(api-design)*
- **L10. Docs:** no narrative guide page under `docs/model_customization/` for a sizeable feature (workload modes, template mode, deploy-from-recommendation, secrets). `BenchmarkMetric`/`BenchmarkMetrics` are in the submodule `__all__` but not re-exported to `sagemaker.serve`, so autodoc won't render them on `sagemaker_serve.rst`. *(SDK-integration + api-design)*
- **L11. Copyright-header inconsistency:** new **source** files use the full 13-line Apache header; new **test** files use a truncated 7-line header — a third variant matching neither the repo norm nor the PR's own source files. *(api-design)*
- **L12. Local aliased imports** `import time as _time` / `import uuid as _uuid` inside `_deploy_recommendation` despite module-level `time`/`uuid`. Remove. *(api-design)*
- **L13. Unrelated churn** in the branch: `sagemaker-train/…` files + `docs/conf.py`/`docs/robots.txt` appear out of scope; confirm they belong or split out (relevant since B1 also looks like accidental branch state). *(SDK-integration)*
- **L14. `deploy()` sets `self._deployed = True` before the recommendation branch;** a raise in `_deploy_recommendation` leaves a misleading "already deployed" state on retry. *(correctness)*
- **L15. Tarfile read** (`_read_member_from_tar_gz`) is in-memory only (no path-traversal risk) but does an unbounded read of a job-owned S3 archive (zip-bomb/OOM, low likelihood). *(security)*

---

ZealSV:
> ## Consolidated findings
> ### BLOCKER
> **B1. `register()` silently lost its return value — breaking change to an existing public API.** _(all 5 reviewers + principal; verified by reading the full method)_ `sagemaker-serve/src/sagemaker/serve/model_builder.py`, end of `register()`. The PR's final diff hunk deletes:
> 
> ```python
> -        return model_package_response.get("ModelPackageArn")
> ```
> 
> `register()` is still annotated `-> Union[ModelPackage, ModelPackageGroup]` and its docstring Returns section still promises the created package. After this change the non-`PipelineSession` path falls off the end of the function and returns **`None`**. Any existing caller doing `arn = mb.register(...)` — a very common pattern (you need the ARN to approve/deploy/reference the package) — now silently gets `None`, breaking downstream describe/approve/CreateModel-from-package flows. It is unrelated to this feature and looks like a rebase/merge artifact. **No test covers `register()`'s return** (grep of `tests/**` for `.register(` = 0 hits), so it ships silently. **Fix:** restore the `return`; add a unit test asserting `register()` returns the ARN (and `None` under `PipelineSession`).
> 
> ### HIGH
> **H1. Plaintext credentials can be persisted server-side — no ARN validation on secret strings.** _(security + principal)_ `ai_inference_recommender/workload.py` (`secrets: Dict[str, Union[str, Secret]]`, `hf_token: Optional[Union[str, Secret]]` on all three factories; `to_inline()`). A `str` passed as `hf_token`/secret is forwarded verbatim by `to_inline()` into `payload["secrets"]`, then written into a persisted `AIWorkloadConfig` (`WorkloadSpec.inline`) by `_ensure_workload_config`. The docstrings say "a Secrets Manager ARN string", but **nothing validates ARN-vs-plaintext**. A user who passes the literal token (the natural mistake given the parameter name) stores a **cleartext credential** server-side, retrievable via Describe. `Workload` is a Pydantic model with no `SecretStr`/`repr=False`, so the token also leaks in `print()`/logs/tracebacks. **Fix:** validate the ARN pattern (reject bare strings, or auto-wrap via `Secret.from_string`); use `SecretStr` + `repr=False` for secret fields.
> 
> **H2. `BenchmarkResult.from_job` reads job status before refreshing — the refresh guard is effectively dead.** _(correctness + principal)_ `ai_inference_recommender/result.py` `from_job`. `refresh()` is gated on `output_config` being empty, but `output_config.s3_output_location` is a required create input the service echoes back, so for any job from `start_benchmark(...)` the guard is always False → **`refresh()` never runs** and `status`/`s3_output_location` are read from the stale create-time describe. `BenchmarkJob.show_result()`'s docstring ("calls refresh() once") is therefore false; a `wait=False` job polled elsewhere then `show_result()`'d can wrongly raise `RuntimeError("...has not finished...")`. Asymmetric with `RecommendationJob.show_result` (which refreshes unconditionally) and `_deploy_recommendation` (which refreshes on empty rows). **Fix:** refresh unconditionally (or when status is non-terminal) in `from_job`, or have `BenchmarkJob.show_result` refresh first.
> 
> **H3. Sessions/regions are not threaded into `sagemaker-core` calls → wrong-region/wrong-credentials.** _(correctness + SDK-integration + principal)_ `ai_inference_recommender/_model_builder_methods.py`: `AIBenchmarkJob.create`, `AIRecommendationJob.create`, `AIWorkloadConfig.create` are all called with **no `session=`** (fall back to the SDK default region/creds), while role and default bucket are resolved from a _different_ session. `start_benchmark` has **no `session`/`sagemaker_session` param at all**, and ignores the session of a passed `Endpoint` resource — so a benchmark can be created in the default region targeting an endpoint that lives elsewhere. **Fix:** thread `session=`/`region=` into every `.create()`/`.get()`; add a `sagemaker_session=` param to `start_benchmark`; derive the session from a passed `Endpoint`.
> 
> **H4. `_deploy_recommendation` mixes sessions across a single operation → cross-region failure.** _(correctness + SDK-integration + principal)_ `model_builder.py` `_deploy_recommendation`: the ModelPackage is described/approved via `self.sagemaker_session.sagemaker_client` (builder session), but `Model.create` / `EndpointConfig.create` / `Endpoint.create` are called with no `session=` (default session). If the builder session's region ≠ the default region, the ModelPackage is approved in region A while the Model referencing that ARN is created in region B → create fails or references a cross-region package. **Fix:** thread `session=self.sagemaker_session` into all three `.create()` calls (or route the whole method through `sm_client`).
> 
> **H5. Recommended `instance_count > 1` is silently forced to 1 when deployed via `deploy()`.** _(test-quality found the live bug)_ `model_builder.py` `_deploy_recommendation`: `initial_instance_count or rec_instance_count or 1`. Public `deploy()` passes `initial_instance_count` whose signature default is **`1`**, so `1 or rec_instance_count → 1` — a recommendation that recommends multiple instances deploys with **one**. The unit tests miss it because they call `_deploy_recommendation` directly with `initial_instance_count=None`; the real `deploy()` caller never sends `None`. **Fix:** default `deploy(initial_instance_count=None)` for the recommendation path (or treat `1` as "unset"); add a test that drives `deploy()` (not the private method) with a row whose `instance_count>1`.
> 
> **H6. Eager top-level import of the recommender widens `import sagemaker.serve` blast radius.** _(SDK-integration)_ `serve/__init__.py` now eagerly imports the whole `ai_inference_recommender` package (which imports ~20 `AI*` core shapes + resources) at module top level, and `sagemaker-core` is pinned `>=2.15.0` with no upper bound. If a future core renames/moves any `AI*` shape, `from sagemaker.serve import ModelBuilder` breaks for users who never touch the recommender. **Fix:** lazy re-export via module `__getattr__` (PEP 562), mirroring the in-function lazy imports already used in `model_builder.py`; add a tested upper bound on `sagemaker-core`.
> 
> ### MEDIUM
> **M1. Silent ModelPackage auto-approval bypasses governance.** _(security + test-quality + api-design + principal)_ `_deploy_recommendation` calls `update_model_package(ModelApprovalStatus="Approved")` with no flag, log, or confirmation, and never reverts. Since rec jobs can register into a customer-named model-package group, this bypasses org manual-approval gates (EventBridge/MLOps promotion automations key off that transition). Also entirely untested (every test stubs `ModelApprovalStatus="Approved"`). **Fix:** make it opt-in (`auto_approve=False` default → actionable error), log the transition, and test both the approve and already-approved branches.
> 
> **M2. Once a recommendation job is attached, `deploy()` can never deploy the built model.** _(correctness + test-quality)_ `deploy()` short-circuits to `_deploy_recommendation` whenever `_recommendation_job` is set, before the `built_model` check, with no opt-out. `mb.build(); mb.generate_deployment_recommendations(...); mb.deploy()` deploys rec index 0, not the built model. Precedence is untested. **Fix:** add an explicit selector (e.g. `deploy(use_recommendation=False)`); document the takeover; lock precedence with a test.
> 
> **M3. New `deploy()` kwargs `role`/`model_name`/`endpoint_config_name` are ignored on the non-recommendation path.** _(correctness + api-design)_ They only take effect in recommendation mode; on a normally-built model `mb.deploy(role=...)` is a silent no-op (the normal path uses `self.role_arn`). (Confirmed _not_ a regression — the pre-existing path never consumed these from `**kwargs` — but a usability trap now that they're in the signature.) **Fix:** honor them in the normal path, or scope/document them as recommendation-only.
> 
> **M4. `recommendation_index` out-of-range → raw `IndexError`; negative index wraps silently.** _(correctness + test-quality)_ `rec = rows[recommendation_index]` has no bounds check (the `recommendation_spec_name` path, by contrast, validates and lists available names). Untested. **Fix:** bounds-check → `ValueError` naming the valid range.
> 
> **M5. `_find_object` does not paginate `list_objects_v2`.** _(correctness + test-quality + principal)_ `result.py`: single call, ignores `IsTruncated`/`NextContinuationToken` → only the first 1000 keys. If `output.tar.gz` sits beyond page 1 (shared/reused output prefixes) it's a spurious `FileNotFoundError`. Untested. **Fix:** use a paginator.
> 
> **M6. `recommendation_spec_name` selection silently takes `matches[0]`.** _(correctness)_ `inference_specification_name` need not be unique across rows; picking `[0]` can deploy a different config than intended without warning. **Fix:** if `len(matches) > 1`, raise/warn and point to `recommendation_index`.
> 
> **M7. Speculative-decoding draft model likely not mounted.** _(SDK-integration)_ When a `base_model` additional source is present, the code promotes it to the primary `model_data_source` and **drops all `AdditionalModelDataSources`**, yet still sets `OPTION_SPECULATIVE_DRAFT_MODEL` to a `/opt/ml/additional-model-data-sources/...` path that only an attached additional source would populate. For an optimized rec with a draft model, that path won't exist at runtime. **Fix:** verify the container contract; mount the draft channel (or nest it under the promoted base prefix).
> 
> **M8. `Secret` deletion is destructive by default and over-broad.** _(security + correctness)_ `secrets.py`: `delete()`/`__exit__` default `ForceDeleteWithoutRecovery=True` (no 30-day window). `Secret` is a plain dataclass with a public `arn`, so `with Secret(arn=<existing prod secret>): ...` force-deletes **any** wrapped pre-existing secret. `__exit__` also loses the creating session (deletes against the ambient default account/region). **Fix:** default recovery on; only auto-delete secrets this object created (track `_created`); retain and reuse the creation session in `__exit__`.
> 
> **M9. `FeatureGatedError` / `WorkloadValidationError` are exported but never raised.** _(test-quality + api-design)_ Grep finds zero `raise` sites. The intended "feature not enabled… see runbook" error never fires; a gated account gets a raw service error. The two exception tests give false confidence by only checking construction of unused classes. **Fix:** wire them into the service-response handling (and test the mapping), or drop them from the public API until wired.
> 
> **M10. `show_result()` and `.recommendations` return different types.** _(api-design)_ `mb.recommendations` → `_RecommendationsView` (comparative table repr + `.best`); `RecommendationJob.show_result()` → a **plain `list`** of `_RecommendationView` (messy default-list repr, no `.best`). **Fix:** have `show_result()` also return `_RecommendationsView`.
> 
> **M11. Rich multi-line tables live in `__repr__` rather than `__str__`.** _(api-design)_ `BenchmarkResult`/`BenchmarkMetrics` and the recommendation views put box-drawing (`─`) tables in `__repr__`, so logs/tracebacks/nested containers get multi-line noise and can raise `UnicodeEncodeError` under non-UTF-8 stdout. **Fix:** rich table in `__str__`; short unambiguous `__repr__`; add `_repr_html_`/`_repr_pretty_` for notebooks.
> 
> **M12. Integ test leaks the published (auto-approved) ModelPackage.** _(test-quality)_ `test_recommendation_workflow_end_to_end` cleans up model/endpoint/config/job/workload-config but not the ModelPackage (or its group). Repeated runs accumulate approved packages. **Fix:** best-effort delete of `rows.best.model_details.model_package_arn` in `finally`.
> 
> ### LOW / NIT
> * **L1. `job.__class__ = BenchmarkJob`/`RecommendationJob` reassignment.** Works today (pure-Python pydantic-v2 subclasses adding only methods) but fragile: breaks if a subclass adds a field or core adopts `__slots__`, and it's inconsistent with the composition/`__getattr__` wrapper used for recommendation rows. Prefer `BenchmarkJob.model_construct(**job.__dict__)`, a delegating wrapper, or pushing `show_result` upstream. _(correctness + SDK-integration)_
> * **L2. Description/impl mismatch:** the PR says it "extends `optimize()`", but `optimize()` is behaviorally unchanged (only its `-> Model` return annotation was dropped and a docstring line added). The feature is the separate `generate_deployment_recommendations()`. Restore the `optimize()` annotation. _(SDK-integration + principal)_
> * **L3. Inline-kwargs mode is duplicated and inconsistent.** `start_benchmark` and `generate_deployment_recommendations` reimplement the same `workload=` vs inline dispatch; the latter infers a tokenizer and errors nicely, the former dies with a raw `TypeError: synthetic() missing … 'tokenizer'`. Factor into one helper; equalize error quality. _(api-design)_
> * **L4. Naming drift:** `role` vs `role_arn`, `name` vs `job_name`, `advanced_optimization`→`optimize_model` (wire), `tags` typed three ways (`List[Tag]` / `Tags` / `List[Dict[str,str]]`), `PerformanceTarget.TTFT_MS="ttft-ms"` (embeds a unit/hyphen unlike its bare-noun siblings). Harmonize. _(SDK-integration + api-design)_
> * **L5. `Workload.sonnet` alias** is a pure alias for `synthetic()` that leaks an implementation detail (AIPerf seed dataset) and invites confusion. Consider dropping. _(api-design)_
> * **L6. Missing docstrings** for the five new `deploy()` kwargs (esp. that recommendation deploy auto-approves the ModelPackage and that `recommendation_spec_name` beats `recommendation_index`); `tags` only reachable via `**kwargs`. Missing type hints on `generate_deployment_recommendations(workload=…)`, the `recommendations` property, and `from_recommendation_job(recommendation_job=…)`. _(api-design)_
> * **L7. Telemetry bucket:** benchmark/recommendation emit `Feature.MODEL_CUSTOMIZATION` — semantically a serve/inference feature, not model customization. _(principal)_
> * **L8. Magic strings:** hardcoded `inference_ami_version="al2-ami-sagemaker-inference-gpu-2"` (ages out; applied `-gpu-` unconditionally even for non-GPU instances), `"base_model"`, `"/opt/ml/model"`, `"LEAST_OUTSTANDING_REQUESTS"`. Pull into named constants; gate the AMI on instance family. _(api-design + principal)_
> * **L9. Error leaks a raw shape:** `"Recommendation has no ModelPackageArn … Raw recommendation: {rec}"` interpolates the whole service object. Reference the row index / spec name instead. _(api-design)_
> * **L10. Docs:** no narrative guide page under `docs/model_customization/` for a sizeable feature (workload modes, template mode, deploy-from-recommendation, secrets). `BenchmarkMetric`/`BenchmarkMetrics` are in the submodule `__all__` but not re-exported to `sagemaker.serve`, so autodoc won't render them on `sagemaker_serve.rst`. _(SDK-integration + api-design)_
> * **L11. Copyright-header inconsistency:** new **source** files use the full 13-line Apache header; new **test** files use a truncated 7-line header — a third variant matching neither the repo norm nor the PR's own source files. _(api-design)_
> * **L12. Local aliased imports** `import time as _time` / `import uuid as _uuid` inside `_deploy_recommendation` despite module-level `time`/`uuid`. Remove. _(api-design)_
> * **L13. Unrelated churn** in the branch: `sagemaker-train/…` files + `docs/conf.py`/`docs/robots.txt` appear out of scope; confirm they belong or split out (relevant since B1 also looks like accidental branch state). _(SDK-integration)_
> * **L14. `deploy()` sets `self._deployed = True` before the recommendation branch;** a raise in `_deploy_recommendation` leaves a misleading "already deployed" state on retry. _(correctness)_
> * **L15. Tarfile read** (`_read_member_from_tar_gz`) is in-memory only (no path-traversal risk) but does an unbounded read of a job-owned S3 archive (zip-bomb/OOM, low likelihood). _(security)_

Addressed all High and Medium findings

Lokiiiiii (approved):

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -4418,6 +4426,467 @@ def _model_builder_optimize_wrapper(
         self.built_model = self._create_model()
         return self.built_model
 
+    @_telemetry_emitter(
+        feature=Feature.MODEL_CUSTOMIZATION,
```
zhaoqizqwang, line 4430: this is not correct? Do we have an inference feature to use for telemetry?

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -4418,6 +4426,467 @@ def _model_builder_optimize_wrapper(
...
+        func_name="ModelBuilder.generate_deployment_recommendations",
+    )
+    def generate_deployment_recommendations(
+        self,
+        workload=None,
+        performance_target: Optional[Union[str, "PerformanceTarget"]] = None,
+        *,
```
zhaoqizqwang, line 4437: What does this input do?

ZealSV: workload is the benchmark traffic spec the recommender uses to evaluate candidate configs. Two ways to pass it:
- a Workload instance or name/ARN of an existing AIWorkloadConfig or
- omit it and pass workload params inline as kwargs

sagemaker-serve/src/sagemaker/serve/model_builder.py
ZealSV: PySDK currently doesn't have an inference feature for telemetry, I used the standard across other serve methods

zhaoqizqwang (approved):

PR #5835, opened 2026-05-07
fix: unskip gpu intensive tests

mujtaba1747:
Re-enable GPU intensive integration tests for SFT, DPO, RLAIF, and RLVR trainers. Updates test configurations to use S3 datasets and unique job names to avoid `ResourceAlreadyExists` errors. Nova tests are refactored to use a dedicated us-east-1 session fixture but remain skipped pending region enablement.

| Test | Action | Changes |
|------|--------|---------|
| `test_dpo_trainer_integration.py::test_dpo_trainer_lora_complete_workflow` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_dpo_trainer_integration.py::test_dpo_trainer_with_validation_dataset` | Unskip + Modify | Dataset ARN → S3 path (train + val); added `base_job_name` |
| `test_rlaif_trainer_integration.py::test_rlaif_trainer_lora_complete_workflow` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_rlaif_trainer_integration.py::test_rlaif_trainer_with_custom_reward_settings` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_rlaif_trainer_integration.py::test_rlaif_trainer_continued_finetuning` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_rlvr_trainer_integration.py::test_rlvr_trainer_lora_complete_workflow` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_rlvr_trainer_integration.py::test_rlvr_trainer_with_custom_reward_function` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_rlvr_trainer_integration.py::test_rlvr_trainer_nova_workflow` | Refactor (still skipped) | Use `sagemaker_session_us_east_1` fixture; removed manual env vars and hyperparameter overrides; added `base_job_name` |
| `test_sft_trainer_integration.py::test_sft_trainer_lora_complete_workflow` | Unskip + Modify | Dataset ARN → S3 path; added `base_job_name` |
| `test_sft_trainer_integration.py::test_sft_trainer_with_validation_dataset` | Unskip + Modify | Dataset ARN → S3 path (train + val); added `base_job_name` |
[...]

PR #5892, opened 2026-05-26
test: unskip gpu intensive tests

lucasjia-aws:
Cherry pick fixes from https://github.com/aws/sagemaker-python-sdk/pull/5835 with minor changes

### DIFF
--- a/.github/workflows/gpu-integ-tests.yml
+++ b/.github/workflows/gpu-integ-tests.yml
@@ -1,0 +1,24 @@
+name: GPU Integ Tests
+on:
+  schedule:
+    - cron: "0 */8 * * *"
+  workflow_dispatch:
+
+permissions:
+    id-token: write # This is required for requesting the JWT
+
+jobs:
+  gpu-integ-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+          aws-region: us-west-2
+          role-duration-seconds: 10800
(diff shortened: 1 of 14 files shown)


.github/workflows/gpu-integ-tests.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  gpu-integ-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
+          aws-region: us-west-2
```
mujtaba1747, line 18: region us-east-1?

lucasjia-aws: From my comprehension: the CodeBuild execution role possesses cross-region SageMaker permissions; the `aws-region: us-west-2` setting in the GitHub Action only affects the STS endpoint used to trigger CodeBuild, and does not influence the region in which the tests are executed. So should be fine.

mujtaba1747: IMO cross region call is a slight anti-pattern. What if there is an outage in us-west-2. Infra/services in us-east-1 should be unaffected ideally.

This is just a CI/CD workflow so it's not really Tier0/Tier 1 service, so we might be okay for now.

mujtaba1747 (approved):

.github/workflows/gpu-integ-tests.yml
lucasjia-aws: ill include all the modifications regarding nova/us-east-1 tests in another PR, those are remaining skipped for now

zhaoqizqwang (approved):

PR #5893, opened 2026-05-26
feat: bedrock-oss-provisioned-throughput-polling

lucasjia-aws:
## Description

Add synchronous polling for Bedrock OSS (non-Nova) model deployments.

Previously, `BedrockModelBuilder.deploy()` for OSS models would fire off a `create_model_import_job` call and return immediately, leaving the user to manually wait for the import to complete. This PR adds polling so `deploy()` waits until the import job completes and the model is ready for on-demand inference, matching the synchronous behavior of the Nova deployment path.

`CreateProvisionedModelThroughput` is provided as a separate public method (`create_provisioned_throughput()`) for users who need dedicated throughput, rather than being mandatory in the deploy flow.

### Changes

**Implementation** (`bedrock_model_builder.py`):
- Non-Nova `deploy()` now waits for the import job to complete and returns the job details. Model is ready for on-demand inference after `deploy()` returns.
- New public method: `create_provisioned_throughput()` — creates provisioned throughput for an imported model and polls until InService. This is a standalone method, not called by `deploy()`.
- New private methods: `_wait_for_import_job_complete()`, `_wait_for_provisioned_throughput_in_service()`.

**Unit tests** (`test_bedrock_model_builder.py`):
- Updated existing non-Nova deploy tests to verify `deploy()` waits for import completion and does NOT call PT APIs.
- Added test classes: `TestWaitForImportJobComplete`, `TestCreateProvisionedThroughput`, `TestWaitForProvisionedThroughputInService`.

**Integration test** (`test_bedrock_provisioned_throughput.py`):
- New integ test for import job polling (`test_deploy_oss_model_waits_for_import_completion`).
- New integ test for `create_provisioned_throughput()` using a pre-existing Bedrock fine-tuned Llama 3.1 8B custom model.

### Testing

- Unit tests cover all new methods (happy path, polling, failure, timeout) — 66 tests passing.
- Integration test for `deploy()` validated locally against real Bedrock APIs — import job creation, polling to completion (~9 min), and model readiness all confirmed.
- Integration test for `create_provisioned_throughput()` validated locally end-to-end — PT creation, polling to InService (~18 min), and automatic cleanup all confirmed.

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/bedrock_model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/bedrock_model_builder.py
@@ -101,6 +101,7 @@ class BedrockModelBuilder:
         self.model = model
         self._bedrock_client = None
         self._sagemaker_client = None
+        self._imported_model_id = None
         self.boto_session = Session().boto_session
         self.model_package = self._fetch_model_package() if model else None
         self._is_rmp = is_restricted_model_package(self.model_package)
@@ -253,10 +254,11 @@ class BedrockModelBuilder:
             job_arn = import_response.get("jobArn")
             self._wait_for_import_job_complete(job_arn)
 
-            # Return the completed job details
+            # Return the completed job details and store imported model ID
             job_details = self._get_bedrock_client().get_model_import_job(
                 jobIdentifier=job_arn
             )
+            self._imported_model_id = job_details.get("importedModelName")
             return job_details
 
     def create_deployment(
@@ -316,8 +318,8 @@ class BedrockModelBuilder:
 
     def create_provisioned_throughput(
         self,
-        model_id: str,
-        provisioned_model_name: str,
+        model_id: Optional[str] = None,
+        provisioned_model_name: str = None,
         model_units: int = 1,
         commitment_duration: Optional[str] = None,
         tags: Optional[list] = None,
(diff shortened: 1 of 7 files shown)


nargokul (commented):
Can we also update example notebooks ?

sagemaker-serve/src/sagemaker/serve/bedrock_model_builder.py
```diff
@@ -303,6 +314,140 @@ def create_deployment(
...
+
+        Returns:
+            Response from Bedrock create_provisioned_model_throughput API.
+
+        Raises:
+            RuntimeError: If the provisioned throughput fails or times out.
+            ValueError: If model_id or provisioned_model_name is not provided.
+        """
+        if not model_id:
+            raise ValueError("model_id is required for create_provisioned_throughput.")
```
nargokul: Can this be gotten from the constructor or from class variables ?

Like If I do bedrock_model_builder.deplo

lucasjia-aws: Good point! I've updated PR with two commits:
1. make mode_id optional in `create_provisioned_throughput()`, if not provided it will now falls back to the stored model ID after `deploy()` completes
2. update example notebooks for new OSS deploy polling behavior:
    a. `bedrock-modelbuilder-deployment.ipynb`: deploy() now waits for import completion, removed manual polling cell, added PT usage example
    b. `mtrl_finetuning_example_notebook_v3_prod.ipynb`: removed manual polling loop, updated description to reflect automatic waiting, added optional create_provisioned_throughput() example

nargokul (approved):

PR #5894, opened 2026-05-26
tests: unskip three tests with individual issues in train

lucasjia-aws:
`test_hp_contract_mpi_script` -- unskip with fix
`test_benchmark_evaluation_base_model_only` -- unskip
`test_benchmark_evaluation_nova_model` -- remain skipped but eliminate dependency to cross account resources

### `test_hp_contract_mpi_script` fix

**Problem:** `mpi_driver.py` used absolute imports (`from sagemaker.train.container_drivers.distributed_drivers.mpi_utils import ...`), but the `sagemaker-train` package is not installed in the training container. The driver scripts are uploaded as standalone files to `/opt/ml/input/data/sm_drivers/` and executed directly, so the absolute imports fail with `ModuleNotFoundError: No module named 'sagemaker.train'`.

**Fix:** Changed `mpi_driver.py` to use `sys.path.insert` + relative module name imports, matching the pattern already used by `torchrun_driver.py`, which works correctly in the container environment.

### DIFF
--- a/sagemaker-train/src/sagemaker/train/container_drivers/distributed_drivers/mpi_driver.py
+++ b/sagemaker-train/src/sagemaker/train/container_drivers/distributed_drivers/mpi_driver.py
@@ -17,7 +17,11 @@ import os
 import sys
 import json
 
-from sagemaker.train.container_drivers.distributed_drivers.mpi_utils import (
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+from distributed_drivers.mpi_utils import (  # noqa: E402 # pylint: disable=C0413,E0611
     start_sshd_daemon,
     bootstrap_master_node,
     bootstrap_worker_node,
@@ -27,7 +31,7 @@ from sagemaker.train.container_drivers.distributed_drivers.mpi_utils import (
 )
 
 
-from sagemaker.train.container_drivers.common.utils import (
+from common.utils import (  # noqa: E402 # pylint: disable=C0413,E0611
     logger,
     hyperparameters_to_cli_args,
     get_process_count,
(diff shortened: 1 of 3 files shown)


sagemaker-train/src/sagemaker/train/container_drivers/distributed_drivers/mpi_driver.py
aviruthen, line 1: Is this failure of sagemaker.train not being available a cause of sagemaker not being installed within the training container (or the correct version of sagemaker)? I'm not sure relative paths is the correct fix here, although feel free to correct me if I'm wrong

lucasjia-aws: Good question! The driver scripts (mpi_driver.py, torchrun_driver.py, etc.) are not meant to run as part of an installed sagemaker-train package. They're uploaded as standalone files to /opt/ml/input/data/sm_drivers/ in the training container and executed directly by the entrypoint bash script. The container uses a standard AWS DLC image (e.g., pytorch-training:2.0.0-cpu-py310) which doesn't have sagemaker-train installed — and it shouldn't need to.

torchrun_driver.py already uses sys.path.insert + relative module imports and works correctly. mpi_driver.py was the only driver using absolute imports, which was inconsistent with the rest of the codebase. I'm actually mirroring what we have done and is running fine here.

aviruthen: Makes sense, I see in PySDK V2 that a similar approach is being used here too, thanks for the clarification!

PySDK V2 link: https://github.com/aws/sagemaker-python-sdk/blob/master-v2/src/sagemaker/modules/train/container_drivers/distributed_drivers/mpi_driver.py#L31

aviruthen (approved):

PR #5935, opened 2026-06-08
fix: lazy-init boto3 clients in AIRHub to avoid NoRegionError at import time

aviruthen:
*Issue #, if available:*

*Description of changes:*

The AIRHub class in sagemaker-train constructs boto3.client("sagemaker") and boto3.client("s3") as class-level attributes, which execute at import time. Since the MTRL Launch PR (v3.13.0) added from sagemaker.train.multi_turn_rl_trainer import MultiTurnRLTrainer to model_builder.py, import sagemaker.serve now transitively imports AIRHub, causing a NoRegionError in any environment without AWS region configured (e.g., conda-forge build CI).

This change converts those class-level client constructions to lazy initialization via _get_sagemaker_client() and _get_s3_client() classmethods. The clients are still created once and cached — the only difference is they're created on first use rather than at import time.

### DIFF
--- a/sagemaker-train/src/sagemaker/ai_registry/air_hub.py
+++ b/sagemaker-train/src/sagemaker/ai_registry/air_hub.py
@@ -27,9 +27,23 @@ from sagemaker.core.telemetry.constants import Feature
 class AIRHub:
     """AI Registry Hub class for managing hub content operations."""
     
-    # Use production SageMaker endpoint (default)
-    _sagemaker_client = boto3.client("sagemaker")
-    _s3_client = boto3.client("s3")
+    # Lazily initialized clients to avoid requiring AWS region at import time.
+    _sagemaker_client = None
+    _s3_client = None
+
+    @classmethod
+    def _get_sagemaker_client(cls):
+        """Get or create the default SageMaker client (lazy initialization)."""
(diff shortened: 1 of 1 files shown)


zhaoqizqwang (approved):

lucasjia-aws (approved):

sagemaker-train/src/sagemaker/ai_registry/air_hub.py
```diff
@@ -27,9 +27,23 @@
...
-    _s3_client = boto3.client("s3")
+    # Lazily initialized clients to avoid requiring AWS region at import time.
+    _sagemaker_client = None
+    _s3_client = None
+
+    @classmethod
+    def _get_sagemaker_client(cls):
+        """Get or create the default SageMaker client (lazy initialization)."""
+        if cls._sagemaker_client is None:
+            cls._sagemaker_client = boto3.client("sagemaker")
```
mujtaba1747, line 38: Can we use Session() instead of boto3?

mujtaba1747 (changes requested):

mujtaba1747 (approved):

papriwal (approved):

PR #5940, opened 2026-06-10
test: fix canaries-v3

lucasjia-aws:
## Description

This PR fixes two flaky integ tests. Both are caused by cross-run or cross-worker resource contention rather than product bugs, and each is made self-healing so it no longer fails intermittently. These are test-only changes; no product code is modified.

### 1. MTRL evaluator integ fixtures — race under `pytest-xdist`

`sagemaker-train/tests/integ/train/test_mtrl_evaluator.py`

The fixtures for the shared model package group and package used a check-then-create pattern. Under `pytest -n auto`, two workers could both see the group missing, both call `CreateModelPackageGroup`, and the loser would get `ValidationException: Model Package Group already exists`. That crashed the module-scoped fixture and errored the dependent test `test_model_package_config_fields`. Fix: wrap the create calls and treat "already exists" / concurrent creation as success, falling back to reusing the existing resource.

### 2. Bedrock provisioned throughput integ test — leaked PTs exhausting quota

`sagemaker-serve/tests/integ/.../test_bedrock_provisioned_throughput.py`

The test only deleted the provisioned throughput (PT) it created in its own teardown, so a process killed before teardown leaked the PT. These accumulated and exhausted the model-unit quota, causing `ServiceQuotaExceededException` in later runs. Fix: add eager cleanup in the `bedrock_client` fixture to delete `test-pt-integ-*` throughputs older than 2h on setup, matching the self-healing pattern already used by the import-job test.

### DIFF
--- a/sagemaker-core/src/sagemaker/core/user_agent.py
+++ b/sagemaker-core/src/sagemaker/core/user_agent.py
@@ -74,3 +74,5 @@ def get_user_agent_extra_suffix():
         suffix = "{} md/{}#{}".format(suffix, STUDIO_PREFIX, studio_app_type)
 
     return suffix
+
+# Trigger PR check: run full integ test suite.
--- a/sagemaker-serve/tests/integ/test_bedrock_provisioned_throughput.py
+++ b/sagemaker-serve/tests/integ/test_bedrock_provisioned_throughput.py
@@ -17,6 +17,7 @@ import json
 import time
 import random
 import logging
+from datetime import datetime, timezone, timedelta
 from urllib.parse import urlparse
 
 import boto3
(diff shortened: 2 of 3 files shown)


lucasjia-aws:
test_bedrock_model_invoke succeeded:
https://us-west-2.console.aws.amazon.com/codesuite/codebuild/729646638167/projects/sagemaker-python-sdk-ci-sagemaker-serve-integ-tests/build/sagemaker-python-sdk-ci-sagemaker-serve-integ-tests%3Ab82feae2-1147-48a3-bb36-4fc55b4c99b1/log?region=us-west-2

mujtaba1747 (approved):

PR #5946, opened 2026-06-12
fix(train): fall back to subscription recipes for models without base recipes

aviruthen:
*Issue #, if available:*

*Description of changes:*

**Issue**

SFTTrainer raises ValueError: No recipes found with Smtj for technique: SFT,training_type:TrainingType.LORA when used with models that only have subscription-based recipes.

The recipe selection logic exclusively looks for non-subscription recipes as the primary recipe, then merges subscription recipe params on top as an overlay. Some models only ship subscription recipes in the hub — there is no non-subscription "base" recipe — so the lookup always returns None.

**Fix:** 

Fall back to subscription recipes when no non-subscription recipe exists for the requested training type. This aligns with how Nova Forge SDK handles subscription recipes (they're treated as first-class when data_mixing=True).

Handle subscription S3 access point URIs in the primary recipe download path — resolve {customer_id} placeholders and parse access point ARN formats.

Raise a clear, actionable error if the S3 access point denies access, informing the user that their account may not have an active Forge subscription (with a link to docs). This replaces the confusing "no recipes found" error.

Skip redundant subscription overlay merge when the selected primary recipe is already a subscription recipe.

**Testing**

Verified locally that:

1. The original ValueError no longer occurs
2. Unsubscribed users get a clear error: "Could not access subscription recipe for model <model name>. This model only provides subscription-based recipes. Please verify that your account has an active Nova Forge subscription."
3. Models with both recipe types (e.g., Nova Lite v2) continue to work as before (non-subscription recipe selected first)

Reference PR with full e2e testing: https://github.com/aws/sagemaker-python-sdk/pull/5948/changes

### DIFF
--- a/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
@@ -525,7 +525,7 @@ def _get_fine_tuning_options_and_model_arn(model_name: str, customization_techni
                                 v_copy['default'] = None  # No default — won't appear in to_dict() unless set
                             options_dict[k] = v_copy
                 except Exception as e:
-                    logger.debug(f"Could not fetch subscription recipe override_params: {type(e).__name__}: {e}")
+                    logger.warning(f"Could not fetch subscription recipe override_params: {type(e).__name__}: {e}")
 
         if options_dict:
             return FineTuningOptions(options_dict), model_arn, is_gated_model
--- a/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
@@ -448,56 +448,84 @@ def _get_fine_tuning_options_and_model_arn(model_name: str, customization_techni
             raise ValueError(f"No recipes found with Smtj for technique: {customization_technique}")
 
         # Select recipe based on training type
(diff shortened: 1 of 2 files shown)


rsareddy0329 (dismissed):

sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
```diff
@@ -448,56 +448,84 @@ def _get_fine_tuning_options_and_model_arn(model_name: str, customization_techni
...
+                    # Merge: subscription params into _specs only (don't set defaults)
+                    # This makes them settable but not serialized unless user explicitly sets them
+                    for k, v in sub_options.items():
+                        if k not in options_dict:
+                            v_copy = v.copy() if isinstance(v, dict) else v
+                            if isinstance(v_copy, dict):
+                                v_copy['default'] = None  # No default — won't appear in to_dict() unless set
+                            options_dict[k] = v_copy
+                except Exception as e:
+                    logger.debug(f"Could not fetch subscription recipe override_params: {type(e).__name__}: {e}")
```
nargokul: Nit: This needs to be a warning . 
Are we swallowing the exception here ?

nargokul (dismissed):

rsareddy0329 (approved):

mollyheamazon (approved):

PR #5958, opened 2026-06-17
fix(serve): repair HuggingFace -> JumpStart redirect in ModelBuilder

ZealSV:
Two bugs prevented `ModelBuilder(model="<HF_id_with_JS_mirror>").build()` from working with any bare HuggingFace id (gpt2, Qwen/Qwen3-0.6B, microsoft/phi-2, etc.), reproducible end-to-end:

  ModelBuilder.build()
    -> _build_single_modelbuilder              (model_builder.py:2579)
    -> _use_jumpstart_equivalent               (model_builder_utils.py)
    -> _hf_schema_builder_init                 raises TaskNotFoundException
    -> redirect aborts before HF -> JS id swap

1. `sagemaker/serve/schema/task.json` was missing from the published wheel. setuptools needs an explicit `[tool.setuptools.package-data]` declaration; `include-package-data = true` alone is a no-op without a MANIFEST.in or version-controlled file list. Without the JSON in the wheel, `retrieve_local_schemas("text-generation")` raised FileNotFoundError, the caller wrapped it in TaskNotFoundException, and the redirect aborted before swapping to the JumpStart id.

2. `_use_jumpstart_equivalent` called `_build_for_jumpstart()` itself, then `_build_single_modelbuilder` called it again on return. The second call hit `_prepare_for_mode` with consumed state and raised FileNotFoundError. Removed the inner call; the caller already handles it.

Also wraps the HF schema bootstrap in a try/except so a missing local task or flaky remote schema fetch can't abort the redirect, the JumpStart build path supplies its own schema downstream.

Verified end-to-end: ModelBuilder(model="gpt2").build() and ModelBuilder(model="Qwen/Qwen3-0.6B").build() both succeed (model created with TGI / DJL_SERVING respectively). Pure-boto3 control with the same JumpStart djl-inference image + S3 artifacts was already accepted by CreateModel/CreateEndpointConfig/CreateEndpoint, so the AWS contract and JumpStart artifacts were never the issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
@@ -942,16 +942,19 @@ class _ModelBuilderUtils:
                 if not model_task:
                     model_task = hf_model_md.get("pipeline_tag")
                 if model_task:
-                    self._hf_schema_builder_init(model_task)
+                    try:
+                        self._hf_schema_builder_init(model_task)
+                    except (TaskNotFoundException, FileNotFoundError, OSError) as e:
+                        logger.warning(
+                            "Could not initialize HF schema builder for task %r "
+                            "(%s: %s); falling back to the JumpStart-supplied schema.",
+                            model_task, type(e).__name__, e,
+                        )
 
             huggingface_model_id = self.model
             jumpstart_model_id = self._jumpstart_mapping[huggingface_model_id]["jumpstart-model-id"]
             self.model = jumpstart_model_id
             merged_date = self._jumpstart_mapping[huggingface_model_id].get("merged-at")
-
-            # Call _build_for_jumpstart if method exists
-            if hasattr(self, "_build_for_jumpstart"):
-                self._build_for_jumpstart()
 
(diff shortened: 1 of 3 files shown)


sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
```diff
@@ -942,17 +942,20 @@ def _use_jumpstart_equivalent(self) -> bool:
                 if not model_task:
                     model_task = hf_model_md.get("pipeline_tag")
                 if model_task:
-                    self._hf_schema_builder_init(model_task)
+                    try:
+                        self._hf_schema_builder_init(model_task)
+                    except TaskNotFoundException:
```
mollyheamazon: consider broadening to `except (TaskNotFoundException, Exception) as e:`, or at minimum `except (TaskNotFoundException, FileNotFoundError, OSError):`

mollyheamazon (commented):
Add unit test for your change

mollyheamazon (approved):

Issue #5588, opened 2026-02-27
ModelBuilder fails w/ DJL container using HF_MODEL_ID


hqudah-aws:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [X] PySDK V3 (3.4.0)

**Describe the bug**
When using ModelBuilder in "passthrough" mode (image_uri + env_vars, no model
or inference_spec), `build()` fails with a ValidationException because a
non-existent S3 path is injected as `ModelDataUrl` in the CreateModel request.

Passthrough mode is the intended path for deploying models on custom/DJL
containers where the container itself handles model loading (e.g. via
HF_MODEL_ID env var pointing to S3).

**Expected behavior**
Expected: Model created with no ModelDataUrl (container loads via HF_MODEL_ID)
Actual:   ValidationException -- "Could not find model data at
          s3://<bucket>/model-builder/<name>/<uuid>/"

**Suggested Fix**

In `_build_for_passthrough()`, also clear `s3_model_data_url`:

    def _build_for_passthrough(self) -> Model:
        if not self.image_uri:
            raise ValueError("image_uri is required for pass-through cases")
        self.s3_upload_path = None
        self.s3_model_data_url = None   # <-- add this line
        return self._create_model()

**To reproduce**
```
from sagemaker.core.helper.session_helper import Session, get_execution_role
from sagemaker.serve.model_builder import ModelBuilder

sagemaker_session = Session()
region = sagemaker_session.boto_session.region_name
role = get_execution_role(sagemaker_session, use_default=True)
bucket = sagemaker_session.default_bucket()

image_uri = f"763104351884.dkr.ecr.{region}.amazonaws.com/djl-inference:0.33.0-lmi15.0.0-cu128"

model_builder = ModelBuilder(
    image_uri=image_uri,
    role_arn=role,
    sagemaker_session=sagemaker_session,
    instance_type="ml.g5.2xlarge",
    env_vars={
        "HF_MODEL_ID": f"s3://{bucket}/models/any-model",
        "OPTION_ROLLING_BATCH": "vllm",
    },
)

model_builder.build(model_name="passthrough-bug-repro")

sagarneeldubey:
[...]

Issue #5529, opened 2026-02-01
bug: ModelBuilder overwrites user-provided HF_MODEL_ID for DJL Serving, preventing S3 model loading


michaelraskansky:
## Description

When using `ModelBuilder` with `ModelServer.DJL_SERVING`, the SDK unconditionally overwrites `HF_MODEL_ID` with the `model` parameter value. This prevents users from loading models from S3, even when explicitly setting `HF_MODEL_ID` in `env_vars`.

## Expected Behavior

Users should be able to specify an S3 URI for model loading via `env_vars`:
```python
env_vars={"HF_MODEL_ID": "s3://bucket/model/"}
```

Per [DJL documentation](https://docs.djl.ai/docs/serving/serving/docs/lmi/deployment_guide/configurations.html), `option.model_id` accepts both HuggingFace Hub IDs and S3 URIs.

## Actual Behavior

The SDK ignores user-provided values and always sets `HF_MODEL_ID` to the HuggingFace model ID from the `model` parameter.

## Root Cause

In `sagemaker-serve/src/sagemaker/serve/model_builder_servers.py`, multiple build methods use:
```python
self.env_vars.update({"HF_MODEL_ID": self.model})
```

This unconditionally overwrites any user-provided `HF_MODEL_ID`. Affected lines: 139, 215, 323, 429, 535.

## Suggested Fix

Use `setdefault()` to preserve user-provided values:
```python
self.env_vars.setdefault("HF_MODEL_ID", self.model)
```

## Reproduction

```python
from sagemaker.serve.model_builder import ModelBuilder
from sagemaker.serve.utils.types import ModelServer
from sagemaker.serve.builder.schema_builder import SchemaBuilder
from sagemaker.core.image_uris import retrieve as get_image_uri

S3_PATH = 's3://my-bucket/models/Qwen/'

mb = ModelBuilder(
    model='Qwen/Qwen3-VL-4B-Instruct',
    image_uri=get_image_uri(framework='djl-lmi', version='0.30.0', region='us-west-2'),
    model_server=ModelServer.DJL_SERVING,
    schema_builder=SchemaBuilder({'inputs': 'Hello', 'parameters': {}}, [{'generated_text': 'Hi'}]),
    role_arn='arn:aws:iam::123456789012:role/SageMakerRole',
    env_vars={'HF_MODEL_ID': S3_PATH},
    instance_type='ml.g5.2xlarge',
)

[...]

PR #5964, opened 2026-06-21
fix(tgi): honor S3 model_path as weight source for TGI builds (#5943)

### LINKED ISSUE #5943: bug: ModelBuilder TGI path ignores S3 model inputs (model_path and s3_model_data_url), forcing HuggingFace Hub download
# Description

**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**

When using `ModelBuilder` with `ModelServer.TGI`, there is no working way to
load model weights from S3 — the deployed TGI container always pulls weights
from huggingface.co at startup. Both parameters that look like they should
enable S3 loading are silently ignored (neither raises at build time):

1. **`model_path="s3://..."`** — the TGI build path calls
   `_create_dir_structure(self.model_path)`, which runs
   `Path(model_path).mkdir(parents=True, exist_ok=True)`. On a POSIX filesystem
   this does **not** fail on an S3 URI; it silently creates a literal local
   directory tree named `s3:/<bucket>/<prefix>` in the working directory and
   continues. The S3 URI never reaches the container, and `HF_MODEL_ID` still
   resolves to the HF repo id.
   _Source: `sagemaker/serve/model_server/tgi/prepare.py::_create_dir_structure`_

2. **`s3_model_data_url="s3://..."`** — on the TGI path this is treated as an
   *upload destination prefix* for SDK scaffolding, not as a weight source. The
   build method sets `self.s3_upload_path = None` (comment: "TGI downloads
   models directly from HuggingFace Hub") and overwrites the attribute via
   `self.s3_model_data_url, _ = self._prepare_for_mode()`. The container ends up
   with `HF_MODEL_ID=<hf_repo_id>` regardless.
   _Source: `sagemaker/serve/model_builder_servers.py::_build_for_tgi`_

Impact: a 7B model cold-starts in ~10–12 min on every scale-out (re-downloading
from HF), which makes scale-to-zero async endpoints impractical and creates a
hard runtime dependency on huggingface.co. It also blocks deploying custom
fine-tuned weights that are not on the HF Hub at all.

This is closely related to #5529 (DJL Serving overwrites a user-provided
`HF_MODEL_ID`) and #5588 (DJL passthrough injects a bad `ModelDataUrl`), but
this report is the **TGI** path and the required fix differs — see *Additional
context*.

**To reproduce**

Build-only repro (no GPU, no deploy, ungated model so no token needed).
`sagemaker==3.12.0`.

```python
from pathlib import Path

from sagemaker.serve import ModelBuilder
from sagemaker.serve.builder.schema_builder import SchemaBuilder
from sagemaker.serve.utils.types import ModelServer

ROLE = "arn:aws:iam::123456789012:role/SageMakerRole"   # any valid role ARN
S3_URI = "s3://my-bucket/models/qwen2.5-7b-instruct/"    # need not exist for the repro
schema = SchemaBuilder(
    sample_input={"inputs": "Hello", "parameters": {"max_new_tokens": 16}},
    sample_output=[{"generated_text": "Hi"}],
)

# --- Case 1: model_path as an S3 URI ---
mb = ModelBuilder(
    model="Qwen/Qwen2.5-7B-Instruct",
    model_path=S3_URI,
    model_server=ModelServer.TGI,
    schema_builder=schema,
    role_arn=ROLE,
    instance_type="ml.g5.2xlarge",
)
mb.build()  # does NOT raise
print("Case 1 — local junk dir created:", Path(S3_URI).exists())   # True (e.g. ./s3:/my-bucket/...)
print("Case 1 — HF_MODEL_ID:", mb.env_vars.get("HF_MODEL_ID"))     # 'Qwen/Qwen2.5-7B-Instruct' (WRONG)

# --- Case 2: s3_model_data_url as an S3 URI ---
mb2 = ModelBuilder(
    model="Qwen/Qwen2.5-7B-Instruct",
    s3_model_data_url=S3_URI,
    model_server=ModelServer.TGI,
    schema_builder=schema,
    role_arn=ROLE,
    instance_type="ml.g5.2xlarge",
)
mb2.build()  # does NOT raise
print("Case 2 — HF_MODEL_ID:", mb2.env_vars.get("HF_MODEL_ID"))    # 'Qwen/Qwen2.5-7B-Instruct' (WRONG)
```

**Expected behavior**

With weights pre-staged in S3, a TGI endpoint should load them from S3 (mounted
to `/opt/ml/model`) instead of downloading from huggingface.co — and at minimum
the SDK should not silently ignore an S3 value passed via `model_path` /
`s3_model_data_url` (either honor it or raise a clear error).

**Screenshots or logs**

Output of the repro above (SDK 3.12.0):

```
Case 1 — local junk dir created: True
Case 1 — HF_MODEL_ID: Qwen/Qwen2.5-7B-Instruct
Case 2 — HF_MODEL_ID: Qwen/Qwen2.5-7B-Instruct
```

For comparison, the working lower-level deployment logs this within ~3s of
container start (weights served from S3, no HF download):

```
INFO text_generation_launcher: Args { model_id: "/opt/ml/model", ... }
INFO download: Files are already present on the host. Skipping download.
INFO download: Successfully downloaded weights for /opt/ml/model
```

**System information**

- **SageMaker Python SDK version**: 3.12.0 (also present in 3.3.x–3.13.x by source inspection)
- **Framework name (eg. PyTorch) or algorithm (eg. KMeans)**: Hugging Face TGI (Text Generation Inference)
- **Framework version**: TGI DLC `huggingface-pytorch-tgi-inference:2.4.0-tgi2.3.1-gpu-py311-cu124-ubuntu22.04`
- **Python version**: 3.12
- **CPU or GPU**: GPU (`ml.g5.2xlarge`, `ml.g5.12xlarge`)
- **Custom Docker image (Y/N)**: N (AWS-provided HF TGI DLC)

**Additional context**

**Working workaround** — bypass `ModelBuilder` and construct the resources
directly, attaching the S3 weights as an uncompressed `ModelDataSource` and
pointing `HF_MODEL_ID` at the mount path:

```python
from sagemaker.core.resources import Endpoint, EndpointConfig, Model
from sagemaker.core.shapes.shapes import (
    ContainerDefinition, ModelDataSource, ProductionVariant, S3ModelDataSource,
)
from sagemaker.core.image_retriever.image_retriever import ImageRetriever

image_uri = ImageRetriever.retrieve(
    framework="huggingface-llm", region="us-east-1",
    version="2.3.1", image_scope="inference", instance_type="ml.g5.2xlarge",
)

primary = ContainerDefinition(
    image=image_uri,
    environment={
        "HF_MODEL_ID": "/opt/ml/model",   # mount path, not the HF repo id
        "HF_HUB_OFFLINE": "1",
        "SM_NUM_GPUS": "1",
        "MESSAGES_API_ENABLED": "true",
    },
    model_data_source=ModelDataSource(
        s3_data_source=S3ModelDataSource(
            s3_uri="s3://my-bucket/models/qwen2.5-7b-instruct/",
            s3_data_type="S3Prefix",
            compression_type="None",
        )
    ),
)
Model.create(model_name=..., primary_container=primary, execution_role_arn=ROLE, ...)
EndpointConfig.create(..., production_variants=[ProductionVariant(...)])
Endpoint.create(...)
```

This drops cold start from ~12 min to ~3–5 min for a 7B model.

**Relationship to existing issues**

| | #5529 | #5588 | This issue |
|---|-------|-------|-----------|
| Model server | DJL Serving (LMI) | DJL Serving (LMI) | **TGI** |
| Mode | `model` + `env_vars` | passthrough (image_uri + env_vars, no model) | `model` + `model_path`/`s3_model_data_url` |
| Symptom | user `HF_MODEL_ID` overwritten | bad `ModelDataUrl` injected → ValidationException | S3 inputs silently ignored; junk local dir created |
| Does the engine accept S3 in `HF_MODEL_ID`? | Yes (LMI `option.model_id`) | Yes | **No** — TGI expects HF repo id or local path |
| Sufficient fix | `setdefault` on `HF_MODEL_ID` | clear `s3_model_data_url` in passthrough | `setdefault` is **necessary but not sufficient**; TGI also needs weights mounted at `/opt/ml/model` via `ModelDataSource` |

**Suggested fix (TGI path)** — when an S3 URI is supplied:
1. Skip `_create_dir_structure` local-dir creation for S3 inputs.
2. Attach the S3 prefix as an uncompressed `ModelDataSource` on the container —
   the early branch in
   `sagemaker/serve/model_server/tgi/server.py::SageMakerTgiServing._upload_tgi_artifacts`
   already emits the correct shape when `model_path` is an S3 URI, but it's
   currently unreachable because `_create_dir_structure` runs first.
3. Set `HF_MODEL_ID=/opt/ml/model` instead of the HF repo id.

More broadly, consider unifying S3-as-weight-source handling across all
`model_server` branches so behavior is consistent (ties together #5529, #5588,
and this issue).

sagarneeldubey:
## Issue

Fixes aws/sagemaker-python-sdk#5943

## Summary

`ModelBuilder` with `ModelServer.TGI` silently ignored an S3 weight source supplied via `model_path="s3://..."` or `s3_model_data_url="s3://..."`. It created a literal local `s3:/...` directory and set `HF_MODEL_ID` to the HF repo id, so the deployed container always downloaded weights from huggingface.co. This forces a ~10-12 min cold start on every scale-out, makes scale-to-zero async endpoints impractical, creates a hard dependency on huggingface.co, and blocks deploying custom fine-tuned weights not published on the Hub.

`_build_for_tgi` now:
- Detects an S3 weight source before any local directory is created and **skips the local mkdir** for it.
- Attaches the S3 prefix as an **uncompressed `ModelDataSource`** (`S3DataType=S3Prefix`, `CompressionType=None`) by routing through `_prepare_for_mode(model_path=...)`.
- Sets `HF_MODEL_ID=/opt/ml/model` and `HF_HUB_OFFLINE=1` via `setdefault`, preserving any user-supplied `HF_MODEL_ID`.

Because TGI's `HF_MODEL_ID` does not accept an S3 URI (it expects an HF repo id or a local path), the fix mounts the weights at `/opt/ml/model` rather than passing the URI through the env var.

Genuine local paths, HF-Hub downloads, JumpStart, and all non-TGI servers are unchanged. The change is scoped strictly to the TGI path and does not touch the DJL behavior of #5529 / #5588.

## Additional defects fixed (surfaced during real deployment)

- **`HF_HUB_OFFLINE` reset:** the end-of-build reset set it back to `"0"`, defeating offline loading. It now stays `"1"` for the S3-mounted path so TGI loads from `/opt/ml/model` instead of phoning home.
- **Doubled trailing slash:** the `ModelDataSource` `S3Uri` could become `s3://.../prefix//` when the input prefix already ended in `/`. With `S3Prefix` matching, no objects match `...prefix//`, so zero files mount into `/opt/ml/model` and TGI fails to find weights. Normalized to exactly one trailing slash.

## Testing

- New `TestBuildForTGI` cases: S3 `model_path` skips `_create_dir_structure`; container gets `HF_MODEL_ID=/opt/ml/model` + `HF_HUB_OFFLINE=1`; `s3_model_data_url` routes through the S3 branch; user-supplied `HF_MODEL_ID` preserved; `HF_HUB_OFFLINE` survives the post-build reset.
- Preservation tests: local-path, HF-Hub, non-TGI (DJL/TEI), and JumpStart behavior unchanged.
- `test_tgi_server.py`: regression test asserting the `S3Uri` has exactly one trailing slash (no `//`).
- All affected unit tests pass (21 passed). Changed files are black/flake8 clean with zero new docstyle/pylint findings vs. baseline.

### End-to-end validation (real SageMaker inference endpoint, TGI DLC, weights from S3)

Beyond unit tests, the fix was validated against a live SageMaker inference endpoint using the TGI Deep Learning Container, loading model weights directly from an S3 prefix (no HuggingFace download). The build step was sanity-checked to confirm the container config before deploying, and the endpoint then mounted the weights from S3 and served successfully:

```python
schema_builder = SchemaBuilder(
    sample_input={"inputs": "What is deep learning?", "parameters": {"max_new_tokens": 64}},
    sample_output=[{"generated_text": "Deep learning is..."}],
)

builder = ModelBuilder(
    model=args.model_id,
    model_path=s3_uri,                 # the #5943 fix: S3 weights, no HF download
    model_server=ModelServer.TGI,
    schema_builder=schema_builder,
    env_vars=build_env(args),
    instance_type=args.instance_type,
    role_arn=args.role,
    sagemaker_session=sm_session,
)

print("\nBuilding model (patched ModelBuilder)...")
builder.build()

# Sanity-check the fix produced the right container config before deploying.
env = builder.env_vars or {}
junk = Path(s3_uri)
print(f"  no junk dir: {not junk.exists()}  |  HF_MODEL_ID={env.get('HF_MODEL_ID')}  |  HF_HUB_OFFLINE={env.get('HF_HUB_OFFLINE')}")

print(f"\nDeploying endpoint '{args.endpoint_name}' (should mount weights from S3)...")
start = time.time()
endpoint = builder.deploy(
    endpoint_name=args.endpoint_name,
    initial_instance_count=1,
    instance_type=args.instance_type,
    container_startup_health_check_timeout=args.startup_timeout,
)
```

Observed: no local `s3:/...` directory was created; the built container had `HF_MODEL_ID=/opt/ml/model`, `HF_HUB_OFFLINE=1`, and a `ModelDataSource` pointing at the S3 prefix; the endpoint reached `InService` and served inference with weights mounted from S3 (CloudWatch shows the container loading from the mounted path rather than downloading from huggingface.co).

## Future work (out of scope for this PR)

This PR is intentionally scoped to the TGI backend (the subject of #5943). The same "S3 as a weight source" capability should be extended to the other HF-Hub-download backends so behavior is consistent across `ModelBuilder` (see https://sagemaker.readthedocs.io/en/stable/ ):

- **TEI** — nearly identical to the TGI path; smallest follow-up.
- **DJL/LMI** — needs a decision on mechanism, since (unlike TGI) DJL's `HF_MODEL_ID`/`OPTION_MODEL_ID` can accept an S3 URI directly. Must avoid regressing the DJL env-var handling in #5529 and the DJL passthrough in #5588.
- A shared helper for S3 weight-source detection + the uncompressed `ModelDataSource` construction (the trailing-slash normalization currently exists per-adapter) to avoid duplicating this logic across backends.

Related: aws/sagemaker-python-sdk#5529, aws/sagemaker-python-sdk#5588.

## Backward compatibility

Additive and TGI-scoped. Every non-S3 / non-TGI code path is reached exactly as before.


sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
```diff
@@ -220,11 +221,30 @@ def _build_for_tgi(self) -> Model:
...
-        _create_dir_structure(self.model_path)
+        # Detect an S3 weight source (model_path="s3://..." or
+        # s3_model_data_url="s3://...") before any local directory is created. TGI's
+        # HF_MODEL_ID does not accept an S3 URI, so an S3 source is attached as an
+        # uncompressed ModelDataSource (mounted at /opt/ml/model) instead of being
+        # downloaded from the HuggingFace Hub.
+        s3_model_source = None
+        if _is_s3_uri(self.model_path):
+            s3_model_source = self.model_path
+        elif _is_s3_uri(self.s3_model_data_url):
```
mujtaba1747: Not sure if we should accept s3_model_data_url for model artifacts. It may have different use in the code

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
```diff
@@ -299,7 +324,10 @@ def _build_for_tgi(self) -> Model:
 
         model = self._create_model()
 
-        if "HF_HUB_OFFLINE" in self.env_vars:
+        # Reset the in-memory HF_HUB_OFFLINE flag after the container is built,
```
mujtaba1747, line 322: Do we really need to reset?

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
```diff
@@ -220,11 +221,30 @@ def _build_for_tgi(self) -> Model:
...
+        if not s3_model_source:
+            _create_dir_structure(self.model_path)
 
         if isinstance(self.model, str) and not self._is_jumpstart_model_id():
             # Configure HuggingFace model for TGI
-            self.env_vars.setdefault("HF_MODEL_ID", self.model)
+            if s3_model_source:
+                # Weights are mounted at /opt/ml/model; do not download from the Hub.
+                self.env_vars.setdefault("HF_MODEL_ID", "/opt/ml/model")
+                self.env_vars.setdefault("HF_HUB_OFFLINE", "1")
```
mujtaba1747, line 240: Setting HF_MODEL_ID to /opt/ml/model is a sane default when s3_model_source is used.

But HF_HUB_OFFLINE, this may not need to be set for fetching model weights from s3. If customers want to explicitly set it, they should do so by passing env_vars when ModelBuilder is created.

sagarneeldubey: This is a design decision that follows the pattern here https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-serve/src/sagemaker/serve/model_builder_servers.py#L274 

```python
if self.mode in LOCAL_MODES:
     self.env_vars.update({"HF_HUB_OFFLINE": "1"})
```
The idea is that if weights are on local disk, don't go to the Hub. S3 case is semantically same (weights on disk at /opt/ml/model). Setting it keeps that intent consistent.  
Also if the s3 url is incorrect, it would silently fall back to downloading it from HF as HF lib will assume that it is a HF link. The container will stand up after 10 mins with no errors with no obvious indication that S3 link didnt work. The offline mode would surface the error immediately. So I suggest keeping this pattern.

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
sagarneeldubey: This is an existing logic (which is if this var is set, we must be in a local build, so reset it for cleanliness). 
But I had to add another condition for  not s3_model_source because HF_HUB_OFFLINE is set for SAGEMAKER_ENDPOINT too and we don't want to reset it if so.

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
sagarneeldubey: Agree on this. As I dig through the code, it is becoming clear that the intention for `s3_model_data_url` is to be a destination for uploading the model weights and is used for other backend frameworks like TorchServe, TF etc. 
I will drop this condition and keep `model_path` as the documented source for S3-stored-model-weights. This worked in my tests too, so we should be good.

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
mujtaba1747: This is good reasoning. Will test this out as well.

mujtaba1747 (approved):

PR #5966, opened 2026-06-24
fix(tags): use sagemaker-sdk: prefix instead of sagemaker-studio: for…

Narrohag:
… JumpStart tags

The SDK should tag resources with sagemaker-sdk:jumpstart-model-id (and related keys) to distinguish SDK-initiated operations from Studio UI ones. This fixes incorrect source attribution in usage data.

- Rename _get_studio_tags() to _get_jumpstart_tags() in train/utils.py
- Change tag key prefix from sagemaker-studio: to sagemaker-sdk: everywhere
- Update all trainer call sites and unit tests

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-train/src/sagemaker/train/utils.py
+++ b/sagemaker-train/src/sagemaker/train/utils.py
@@ -240,14 +240,14 @@ def _run_clone_command_silent(repo_url, dest_dir):
             logger.error(f"Error output:\n{e}")
             raise
 
-def _get_studio_tags(model_id: str, hub_name: str):
+def _get_jumpstart_tags(model_id: str, hub_name: str):
     return [
         {
-            "key": "sagemaker-studio:jumpstart-model-id",
+            "key": "sagemaker-sdk:jumpstart-model-id",
             "value": model_id
         },
         {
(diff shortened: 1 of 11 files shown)


evakravi (approved):

sagemaker-train/src/sagemaker/train/utils.py
```diff
@@ -240,14 +240,14 @@ def _run_clone_command_silent(repo_url, dest_dir):
             logger.error(f"Error output:\n{e}")
             raise
 
-def _get_studio_tags(model_id: str, hub_name: str):
+def _get_jumpstart_tags(model_id: str, hub_name: str):
```
mujtaba1747, line 243: Why name it get jumpstart tags? Can we call it something that indicates it's SageMaker Python SDK tags

mujtaba1747: It makes sense now. Discussed with author.

mujtaba1747 (approved):

aviruthen (approved):

mujtaba1747:
Some unit and integ tests fail but those are unrelated and being tracked to be fixed in a separate PR.

PR #5989, opened 2026-07-06
fix: always apply evaluator identity keywords and allow explicit domain_id

jam-jee:
## Issue

`Evaluator.create()` inconsistently applies `HubContentSearchKeywords`, so some evaluators are silently invisible in SageMaker Studio despite registering successfully in the hub (confirmable via `describe-hub-content`). Studio filters hub content by keywords such as `@subtype:aws/evaluator` and `@evaluatortype:<type>`; when they are missing, the asset exists but never renders in the UI, and no error is raised.

Internal ref: P467494019.

## Root cause

In `Evaluator.create()` the identity keywords were gated behind `method is not None`:

```python
if method is not None:
    tags = [
        (TAG_KEY_METHOD, method.value),
        ("@subtype", ...), ("@evaluatortype", ...), ("@contenttype", ...),
    ]
else:
    tags = []
```

`method` is only resolved for `RewardFunction`. For `RewardPrompt`, `method` stays `None`, so `tags` was empty and the only keyword ever attached was `@domain` (and only inside Studio). The identity keywords describe *what the asset is* and should never depend on whether a method was resolved.

## Changes

- **Always** attach the identity keywords (`@subtype` / `@evaluatortype` / `@contenttype`) in `Evaluator.create()`, regardless of evaluator type. Only `method` remains conditional (reward prompts have no method).
- Add an optional `domain_id` parameter to `Evaluator.create()` **and** `DataSet.create()`. `@domain` is already auto-injected inside Studio, but the domain cannot be inferred outside Studio (laptop / CI), leaving assets un-scoped and hidden. Callers can now pass `domain_id` explicitly; auto-detection is skipped when it is provided. Fully backward compatible (defaults to `None`).
- Regression tests: identity keywords always present for both `RewardFunction` and `RewardPrompt`; explicit `domain_id` is tagged and auto-detection is not invoked, for both `Evaluator` and `DataSet`.

## Testing

- `tests/unit/ai_registry` passes (90 tests), including the 3 new regression tests.
- Also tested manually in model customization aws account. Created evaluator in a jupyter lab space and able to see those evaluators in Studio. 
- 
Screenshot 2026-07-09 at 3 34 44 PM

### General

- [x] I have updated any necessary documentation, including [READMEs](https://github.com/aws/sagemaker-python-sdk/blob/master/README.rst) and [API docs](https://github.com/aws/sagemaker-python-sdk/blob/master/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

### Tests

### DIFF
--- a/sagemaker-train/src/sagemaker/ai_registry/dataset.py
+++ b/sagemaker-train/src/sagemaker/ai_registry/dataset.py
@@ -230,13 +230,14 @@ class DataSet(AIRHubEntity):
         description: str = "",
         tags: Optional[List[Tuple[str, str]]] = None,
         role: Optional[str] = None,
+        domain_id: Optional[str] = None,
         sagemaker_session: Optional[Session] = None,
     ) -> "DataSet":
         """Create a new DataSet Hub AIR entity.
-        
+
         Creates a new version if entity already exists. This is the primary entry point
         for users. Uploads to S3 internally if local file input is provided.
-        
+
         Args:
             name: Name of the dataset
             source: S3 URI or local file path for the dataset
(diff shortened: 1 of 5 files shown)


mujtaba1747 (approved):
Nice

jam-jee:
Some of the integ tests are broken. We will need to get 100% integ test completed before release.

PR #5993, opened 2026-07-07
feat(serve): add opt-in model source tag-based resource reuse

amazeAmazing:
# Model Source Tag-Based Resource Reuse

## Summary

Adds **opt-in** resource reuse to the V3 serve builders so redeploying the same model artifacts doesn't create duplicate resources (or exhaust quota-limited Bedrock imported models). Each deploy tags the resources it creates (Model + Endpoint) with a stable model-source identifier; when reuse is requested, a later call for the same source discovers and returns the existing resources instead of creating new ones.

Controlled by `reuse_resources: bool = False`. On a hit we **log a warning and return the existing resource** (we do not raise).

## Behavior

| | BedrockModelBuilder | ModelBuilder (SageMaker) |
|---|---|---|
| Flag | `deploy(reuse_resources=)` | `build(reuse_resources=)` + `deploy(reuse_resources=)` |
| Build discovery | — | `list_models` + model-source tag → reuses existing Model |
| Deploy discovery | Bedrock `list_custom_models` + tags | SageMaker `list_endpoints` + endpoint tag |
| Reuses | custom model + active deployment | Model (build) + endpoint (deploy) |
| Config re-validation | no | yes (env vars, image URI, instance type on model-on-variant endpoints) |
| On match | warn + return | warn + return |

`reuse_resources` is not inherited: set it on `build()` to skip Model creation and on `deploy()` to reuse the endpoint.

**Inference components:** IC deployments (`inference_config` is a `ResourceRequirements`, or a `modelbuilder_list` build) are **not** intercepted by `reuse_resources` at the deploy level. They manage their own reuse by `endpoint_name` (infrastructure reuse) and `inference_component_name` (IC update). Passing `reuse_resources=True` on an IC deploy logs a warning that the flag is ignored. However, `build(reuse_resources=True)` still works for IC-based models — it finds the existing Model by tag and avoids creating a duplicate.

## Nova inference-component deployment

Nova model-customization deploys support inference components. When a `ResourceRequirements` `inference_config` is supplied, the Nova checkpoint is hosted as a single inference component referencing the built Model (which carries the Nova image, escrow artifacts, and env); without one, it keeps the direct model-on-variant path.

- `_is_nova_model` identifies Nova from the model package recipe/hub-content name, and also from a package-less source (raw S3 checkpoint or trainer) via `base_model_name`.
- The IC endpoint config sets `EnableNetworkIsolation` to match the built Model (always `True` for Nova), fixing a `CreateInferenceComponent` rejection on mismatched network isolation.
- Model-package-dependent logic (restricted-package path, PEFT/recipe metadata, lineage tracking) is guarded so package-less Nova checkpoints deploy cleanly.
- Endpoints created on the shared IC path are tagged with the model-source tag so they stay discoverable.

## Model-tag reuse

Both Nova and OSS (non-Nova) Models are tagged with the model-source identifier at build time. `build(reuse_resources=True)` searches existing Models by tag (`_find_reusable_model`) and skips `Model.create` on a match, regardless of whether the endpoint is model-on-variant or IC-based.

## Checkpoint URI resolution

`resolve_nova_checkpoint_uri` handles all three Nova output layouts:
- HyperPod: `<output>/<job>/manifest.json`
- Serverless: `<output>/<job>/output/output/manifest.json`
- Serverful: `<output>/<job>/output/output.tar.gz` (manifest inside)

Tries each in turn. If all fail, surfaces every error in the raised exception.

## When reuse works and when it does not

| Scenario | Reuse supported? | Notes |
|----------|:---:|-------|
| Same source, same config, model-on-variant endpoint (Nova) | ✅ | Model found by tag (build skips); endpoint found by tag + config match (deploy returns existing) |
| Same source, same config, IC-based endpoint (OSS, Nova IC) | ✅ | Model found by tag (build skips); endpoint found by tag (deploy returns existing) |
| Same source, different config (env/image/instance) | ❌ creates new | Config mismatch detected at deploy → new endpoint created. Model still reused at build if source matches. |
| Explicit IC deploy (`inference_config=ResourceRequirements(...)`) | Model reuse only | `build(reuse_resources=True)` finds Model by tag. `deploy()` warns flag is ignored and creates the IC normally. |
| `modelbuilder_list` / `CustomOrchestrator` build | N/A | `is_inference_component_build=True` → reuse gate skipped at build time |
| Bedrock custom model | ✅ | Tag match on the custom model; also reuses active deployment |

## Key changes

- New `sagemaker/serve/model_reuse.py`: tag helpers + service-client discovery (`find_existing_bedrock_model`, `find_active_bedrock_deployment_for_model`, `find_existing_sagemaker_endpoint`) with status gating.
- `sagemaker/core/training/utils.py`: shared Nova manifest/checkpoint helpers (consolidated from duplicated logic), three-layout resolution (HyperPod/serverless/serverful).
- `ModelBuilder`:
  - `build()`: tags the Model with model-source; `_find_reusable_model` searches by tag (paginated) to skip duplicate creation.
  - `deploy()`: endpoint reuse gate with config validation; IC deploy warns reuse has no effect.
  - `self.instance_type` set from caller's explicit value before `_deploy_model_customization`.
  - `EnableNetworkIsolation` on IC endpoint configs matches the built Model.
  - Raw-S3 `model` input via `model_metadata={"BASE_MODEL_NAME": ...}`.
- `BedrockModelBuilder`: reuse of custom model + deployment; consolidated source resolution; response always includes `modelArn`.

## Tests

- Unit: `test_model_reuse.py`, `test_model_builder.py` (Nova IC routing, network isolation, model-on-variant fallback), `test_bedrock_model_builder.py`, core `test_training_utils.py` (3-layout resolution).
- Integ (in-repo): model-source tag assertion on OSS deploy (`test_model_customization_deployment.py`), Nova Bedrock e2e (`test_nova_model_customization_deployment.py`).
- Manually tested: Bedrock reuse, SageMaker endpoint reuse (Nova), and OSS reuse (Model + endpoint),

### DIFF
--- a/sagemaker-core/src/sagemaker/core/training/utils.py
+++ b/sagemaker-core/src/sagemaker/core/training/utils.py
@@ -13,8 +13,12 @@
 """Training utilities."""
 from __future__ import absolute_import
 
+import io
+import json
 import os
+import tarfile
 from typing import Any, Literal
+from urllib.parse import urlparse
 from sagemaker.core.utils.utils import Unassigned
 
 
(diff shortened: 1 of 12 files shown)


sagemaker-core/src/sagemaker/core/training/utils.py
```diff
@@ -75,3 +79,135 @@ def _is_valid_s3_uri(path: str, path_type: Literal["File", "Directory", "Any"] =
...
+
+    Returns:
+        The checkpoint URI recorded in the manifest.
+
+    Raises:
+        ValueError: If the checkpoint URI cannot be resolved from either source.
+    """
+    manifest_uri = build_nova_manifest_s3_uri(s3_output_path, training_job_name)
+    try:
+        return read_nova_checkpoint_uri_from_manifest(s3_client, manifest_uri)
```
vaishnavi-kommaraju: what if there's no manifest.json saved in the output folder and it is not a tar.gz file (for hyperpod)? wouldn't we in that case incorrectly throw an error that manifest file not found in the tar.gz output path?

amazeAmazing: Thanks for the callout! I will adjust

vaishnavi-kommaraju (approved):

sagemaker-core/src/sagemaker/core/training/utils.py
mujtaba1747: Yup training outputs in SMTJ are tar.gz files by default.

sagemaker-core/src/sagemaker/core/training/utils.py
```diff
@@ -75,3 +79,147 @@ def _is_valid_s3_uri(path: str, path_type: Literal["File", "Directory", "Any"] =
...
+
+    Args:
+        s3_output_path: The training job's ``output_data_config.s3_output_path``.
+        training_job_name: The training job name.
+
+    Returns:
+        Fully-qualified S3 URI to the job's manifest.json.
+    """
+    output_path = s3_output_path.rstrip("/")
+    return f"{output_path}/{training_job_name}/output/output/manifest.json"
```
mujtaba1747, line 118: The output path for manifest json may be different for serverless and hyperpod.

For serverless: training_job_name/output/output/manifest.json
For SMHP: training_job_name/manifest.json

Can we test this for both flows.

amazeAmazing: Thanks for calling this out! I will push a new revision to address this gap. 

I will test all three flows - Serverless, Serverful and HP

sagemaker-serve/tests/integ/test_nova_model_customization_deployment.py
```diff
@@ -491,6 +499,14 @@ def test_nova_bedrock_deployment_active(self, deployed_nova_model, bedrock_clien
         )
         assert deployment.get("status") == "Active"
 
+    def test_nova_bedrock_custom_model_tagged_for_reuse(self, deployed_nova_model, bedrock_client):
```
mujtaba1747, line 502: can we also invoke this deployed model as part of the test, this would make it e2e.

amazeAmazing: This file's test already invokes the deployed model in the Bedrock class sequence (TestNovaBedrockDeployment)

test_nova_bedrock_invoke does the invoking.

mujtaba1747:
Approved. Tests ran on smoketest (nova) account. Any changes needed for integ tests to run on PySDK acc to be tracked separately.

mujtaba1747 (approved):

PR #5987, opened 2026-07-06
fix: refresh LLMAsJudgeEvaluator allowed evaluator models

jam-jee:
## Issue

*Addresses internal ticket P467477489.*

`LLMAsJudgeEvaluator` validates the user-supplied `evaluator_model` against a hardcoded dict, `_ALLOWED_EVALUATOR_MODELS`, in `sagemaker/train/constants.py`. That list had gone stale — several entries are now EOL on Bedrock, and every currently usable judge model (Claude Sonnet 4 / 4.5, Haiku 4.5, Opus 4.5, Nova 2 Lite / Micro / Premier) was missing. As a result the client-side check rejected valid models, and models that *passed* the check failed downstream at Bedrock with end-of-life errors. LLM-as-Judge evaluation was effectively unusable without monkey-patching the SDK.

## Description of changes

- **`constants.py`** — re-sync `_ALLOWED_EVALUATOR_MODELS` (model IDs + per-region availability) with the Bedrock Console judge-model selector, cross-checked against the [Bedrock user guide supported-evaluator-models list](https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported):
  - **Added** Claude Sonnet 4, Claude Sonnet 4.5, Claude Haiku 4.5, Claude Opus 4.5, Nova 2 Lite, Nova Micro, Nova Premier.
  - **Widened** region lists for existing models (e.g. Llama 3.1 70B `us-west-2` → `us-west-2` + `us-east-1`; Nova Pro `us-east-1` → all four regions).
  - **Removed** `anthropic.claude-3-5-haiku-20241022-v1:0` from the allowlist.
  - **Narrowed** Claude 3.5 Sonnet v1/v2 to `ap-northeast-1` to match the source of truth.
  - Added a comment citing the source of truth so the list can be re-synced in future.
- **`llm_as_judge_evaluator.py`** — updated docstring examples that pinned Claude 3.5 Sonnet v1 (now `ap-northeast-1`-only) alongside `us-west-2` ARNs.
- **`test_llm_as_judge_evaluator.py`** — updated the default judge model and the `valid_models` list to reflect the new allowlist.
- **`tests/integ/train/`** — updated the two LLM-as-Judge integ configs that used Claude 3.5 Haiku as the judge model to Claude Sonnet 4.

## Testing

`sagemaker-train` evaluate unit test suite passes locally: **557 passed, 7 skipped**.

## Merge Checklist

[...]

PR #5996, opened 2026-07-07
Fix/v3 tests

lucasjia-aws:
## Summary

Fixes the SageMaker Python SDK **v3 integration test** suite. The failures fell into five root-cause classes: models being resolved against the pinned private test hub instead of the public hub, region/session resolution breaking under `xdist` parallelism, test assertions written against synthetic shapes rather than real behavior, a stale/EOL Bedrock judge model, and transient infra errors (capacity, throttling, concurrent-build S3 races) surfacing as hard failures. No production behavior changes except the EOL judge-model swap and its allowlist entry.

### 1. Hub / model resolution (public-hub model vs pinned private `sdktest` hub)
The session-scoped `use_private_hub` fixture pins `SAGEMAKER_HUB_NAME` to the private `sdktest` hub, but these tests construct evaluators/trainers with public-hub models. Unlike the SFTTrainer path, evaluator resolution does not fall back to the public hub, so `DescribeHubContent` returned `ResourceNotFound` and construction failed with a pydantic `ValidationError`. Fix: override `SAGEMAKER_HUB_NAME` to `SageMakerPublicHub` in the affected tests.

| Test | File |
|---|---|
| `test_inspect_ai_bedrock_evaluation`, `test_inspect_ai_upload_benchmarks` | `sagemaker-train/tests/integ/train/test_inspect_ai_evaluator.py` |
| `TestLLMAJCustomModelIntegration` (LLMAJ base model resolution) | `sagemaker-train/tests/integ/train/test_llmaj_custom_model.py` |
| `test_evaluator_get_resolved_recipe_no_recipe_raises`, `test_evaluator_get_resolved_recipe_with_local_yaml` | `sagemaker-train/tests/integ/train/test_recipe_override_integration.py` |

### 2. Region routing & isolation under `xdist`
Nova models are only available in `us-east-1`, and tests that build a default `Session()` without an explicit region failed with "Must setup local AWS configuration with a region..." when a worker ran only such tests (region was previously set as a side effect of another fixture). Fix: route Nova tests to `us-east-1` via marker/session, pin the LLMAsJudge full-flow test to `us-west-2`, pass an explicit session to the no-recipe SFTTrainer test, and add a session-scoped autouse `ensure_default_region` fixture.

| Test | File |
|---|---|
| Nova `InspectAI` tests, `test_sft_rejects_max_steps_below_minimum` | `test_inspect_ai_evaluator.py`, `test_recipe_override_integration.py` |
| LLMAsJudge full-flow evaluator (pinned `us-west-2`) | `test_llm_as_judge_evaluator.py` |
| `test_sft_get_resolved_recipe_no_recipe_raises` (explicit session) | `test_recipe_override_integration.py` |
| `ensure_default_region` autouse fixture | `sagemaker-train/tests/integ/train/conftest.py` |

### 3. Assertion / expectation correctness
Several tests asserted against synthetic shapes or stale messages instead of real runtime behavior. The SFT recipe tests read flat `training_config.<hyperparameter>` paths matching unit-test mocks, but the real Hub SFT LoRA recipe nests hyperparameters under `training_config.training_args` with different field names (`max_epochs`, `train_batch_size`, `max_len`, `micro_train_batch_size`, `max_norm`). Fix: align assertions and recipe-file paths with the real recipe, update stale error-message regexes, correct the evaluator-ARN expectation, and validate the LLMAJ pipeline via step status.

| Test | Fix |
|---|---|
| `TestSFTTrainerRecipeOverrideInteg`, `TestSFTTrainerFullRecipeOverrideInteg`, `TestSFTTrainerValidationFailuresInteg` | Read hyperparameters at the real nested path with real field names; `test_sft_recipe_file_with_invalid_value_raises` now actually reaches validation; skip `test_sft_save_steps_equal_to_max_steps_passes` (llama SFT trains by epochs, not steps) |
| No-recipe resolver tests | Update regex to current message ("requires a 'recipe', 'overrides', or direct hyperparameter assignments...") |
| `test_extract_evaluator_arn_lambda_creates_retrievable_evaluator` | Lambda ARN is rejected by `_extract_evaluator_arn` (dispatched upstream) — expect `ValueError` |
| `TestLLMAJCustomModelIntegration` | Validate pipeline via step status; make `show_results` best-effort |
| `test_huggingface_integration` | Use canonical HF repo id `openai-community/gpt2` |

### 4. Stale / EOL Bedrock judge model
The default LLM-as-Judge model `anthropic.claude-3-5-haiku-20241022-v1:0` reached end-of-life in Bedrock, so evaluation pipelines failed with a `ValidationException`. Fix: switch to the active successor `anthropic.claude-haiku-4-5-20251001-v1:0` and add it to the allowed evaluator models list.

| Test / File | File |
|---|---|
| `test_llm_as_judge_base_model_fix`, `test_llm_as_judge_evaluator`, allowlist + unit test | `sagemaker-train/.../test_llm_as_judge_*.py`, `sagemaker-train/src/sagemaker/train/constants.py` |

### 5. Flaky / transient infra tolerance
These failures were not code defects but transient conditions under concurrent load. Fix: append a `uuid` suffix to per-test S3 prefixes so concurrent builds no longer share paths (and cleanup only removes their own data); tolerate IAM `SimulatePrincipalPolicy` throttling; add test-level rerun logic; and `xfail` on `InsufficientInstanceCapacity`.

| Test / Area | File |
|---|---|
| `test_clarify`, `test_hyperparameter_tuning`, `test_pipeline_train_registry` (unique S3 prefixes) | `sagemaker-mlops/tests/integ/**` |
| IAM `SimulatePrincipalPolicy` throttling tolerance | `sagemaker-serve/tests/integ/conftest.py` |
| `test_feature_store_lakeformation` (test-level rerun) | `sagemaker-mlops/tests/integ/test_feature_store_lakeformation.py` |
| `test_model_customization_deployment` (`xfail` on capacity) | `sagemaker-serve/tests/integ/test_model_customization_deployment.py` |

## Testing
Ran the affected integ suites under `-n auto` (xdist) across `us-east-1` (Nova) and `us-west-2` (all other) to confirm the region/hub isolation fixes hold regardless of test order.


sagemaker-train/tests/integ/train/test_extract_evaluator_arn_integration.py
```diff
@@ -96,32 +93,12 @@ def test_extract_evaluator_arn_with_evaluator_string(sagemaker_session, evaluato
...
 
 def test_extract_evaluator_arn_with_lambda_arn_string(sagemaker_session, lambda_arn):
-    """Test _extract_evaluator_arn with a Lambda ARN string.
+    """Test _extract_evaluator_arn rejects a Lambda ARN string.
 
-    Verifies that passing a Lambda ARN triggers auto-creation of an Evaluator
-    in the AI Registry and returns the newly created evaluator's hub-content ARN.
+    _extract_evaluator_arn only accepts a SageMaker hub-content evaluator ARN
+    (or an Evaluator object). Lambda ARNs are resolved/dispatched upstream (see
+    rlvr_trainer's _is_lambda_arn handling), so passing a Lambda ARN here must
```
aviruthen: I am less familiar with Evaluators, but this seems like the revised test is testing a behavior that is different than how the test was previously constructed. Can we double check that this new test is verifying the correct behavior?

sagemaker-train/tests/integ/train/test_inspect_ai_evaluator.py
```diff
@@ -121,6 +124,11 @@ def test_inspect_ai_bedrock_evaluation(self, sagemaker_session_us_east_1, inspec
         The execution role is resolved from the active credentials by
         BaseEvaluator (no ``role`` argument passed).
         """
+        # The base model lives in the public hub, not the private "sdktest"
+        # recipe hub that the session-scoped use_private_hub fixture pins
+        # SAGEMAKER_HUB_NAME to. Explicitly resolve against SageMakerPublicHub.
+        monkeypatch.setenv("SAGEMAKER_HUB_NAME", "SageMakerPublicHub")
```
aviruthen: I don't think we should expect customers to monkey patch hub names to test a specific behavior, and I feel integration tests should mimic true customer flows. Would this require a source code change instead...? Is there any way the hub is exposed to the customer that we can use instead?

sagemaker-train/src/sagemaker/train/constants.py
```diff
@@ -64,6 +64,12 @@ def get_sagemaker_hub_name() -> str:
 # allowlist.cross-checked against
 # https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
 _ALLOWED_EVALUATOR_MODELS = {
+    "anthropic.claude-3-5-sonnet-20240620-v1:0": ["us-west-2", "us-east-1", "ap-northeast-1"],
+    "anthropic.claude-3-5-sonnet-20241022-v2:0": ["us-west-2"],
+    "anthropic.claude-3-haiku-20240307-v1:0": ["us-west-2", "us-east-1", "ap-northeast-1", "eu-west-1"],
+    "anthropic.claude-3-5-haiku-20241022-v1:0": ["us-west-2"],
+    "anthropic.claude-haiku-4-5-20251001-v1:0": ["us-west-2", "us-east-1"],
+    "meta.llama3-1-70b-instruct-v1:0": ["us-west-2"],
```
papriwal: Blocking: deduplicate the entries.

sagemaker-train/tests/integ/train/test_extract_evaluator_arn_integration.py
lucasjia-aws: Good catch, and you're right that the behavior changed. From my comprehension:

`_extract_evaluator_arn` is purely a validate-and-extract helper. For a string input it just calls `_validate_evaluator_arn` (which requires a SageMaker `hub-content/.../JsonDoc/...` ARN) and returns it as-is; for an `Evaluator` object it returns `evaluator.arn`. It has no "create an Evaluator" logic.

The previous Lambda-ARN test asserted that passing a Lambda ARN would auto-create an Evaluator in the AI Registry and return a hub-content ARN. That never matched the source: I traced the full git history and `_extract_evaluator_arn` has never contained any creation logic (no `Evaluator.create`), including in the same commit (#5969) that originally introduced this test. Since a Lambda ARN (`arn:aws:lambda:...`) fails the hub-content ARN regex, the function raises `ValueError`, so the old test would fail against the actual code.

Lambda ARNs are handled a layer up in `rlvr_trainer`: it uses `_is_lambda_arn` to branch, and only non-Lambda-ARN strings are treated as evaluator ARNs and passed down to this validation path. So a Lambda ARN reaching `_extract_evaluator_arn` should be rejected, which is exactly what the revised test now asserts (`ValueError` with "must be a valid SageMaker hub-content evaluator ARN").

sagemaker-train/tests/integ/train/test_inspect_ai_evaluator.py
lucasjia-aws: Good call — this was masking a real gap. The evaluator's `_resolve_jumpstart_model` only looked in the hub set by `SAGEMAKER_HUB_NAME`, unlike the SFTTrainer path which falls back to the public hub. I've added that fallback in the source (matching what `recipe_utils` already does), so base models resolve from `SageMakerPublicHub` when they're not in the configured hub. Dropped the `monkeypatch` from all the tests so they run the real flow now, and added unit tests for the fallback.

sagemaker-train/src/sagemaker/train/constants.py
lucasjia-aws: Done. The dupes came from a master merge that pulled back entries #5987 had already removed (and with conflicting region lists). Kept the refreshed allowlist and removed the stale EOL `claude-3-5-haiku-20241022` entry, including in the unit test that referenced it.

sagemaker-train/tests/integ/train/test_extract_evaluator_arn_integration.py
lucasjia-aws: fixed in #6000

aviruthen (approved):

sagemaker-train/tests/integ/train/test_recipe_override_integration.py
```diff
@@ -102,20 +106,21 @@ def test_sft_get_resolved_recipe_overrides_only(self):
...
             training_dataset="s3://mc-flows-sdk-testing/input_data/sft/sample_data_256_final.jsonl",
             accept_eula=True,
+            sagemaker_session=sagemaker_session,
         )
 
-        with pytest.raises(ValueError, match="requires a 'recipe' or 'overrides'"):
+        with pytest.raises(ValueError, match=r"requires a 'recipe', 'overrides'"):
             sft_trainer.get_resolved_recipe()
 
     @pytest.mark.skip(reason="Skipping GPU resource intensive test - submits actual training job")
```
mujtaba1747, line 126: THis should be moved to the thrice daily tests

PR #6000, opened 2026-07-08
rlvr validation bugfix

mujtaba1747:
*Issue #, if available:* N/A

*Description of changes:*

 PR #6000 — Bug Fixes Summary

  The PR contains 2 commits on top of upstream master (one fix, one chore):

  1. fix: removed reward verifier is_lambda + compute check (d8452649)

  Three related fixes in RLVRTrainer / reward verifier:

  - Removed mandatory compute parameter check for Lambda ARNs — Previously, verify_reward_function() raised a
  ValueError requiring compute when a Lambda ARN was used with Nova models. This was incorrect for serverless training
  (no compute needed). The check is removed; Lambda ARNs now work without specifying compute.
  - Replaced Evaluator.get() with direct describe_hub_content call — _verify_reward_function used
  Evaluator.get(evaluator_name) which relied on AIRHub (hardcoded hub name). A new _get_lambda_arn_from_evaluator_arn()
  function parses the hub name/type/version directly from the ARN and calls describe_hub_content, so it works for
  evaluators stored in any hub.
  - Added Lambda→Evaluator creation to train() time — When custom_reward_function is a raw Lambda ARN, the SDK now auto-creates an Evaluator object at train() time for serverless compute, so the evaluator ARN is properly passed to the
  training job.
  - Added skip_reward_validation parameter — New boolean flag on RLVRTrainer to skip the reward function verification
  step before job submission.
  - HyperPod + local file path now logs a warning instead of silently skipping or raising — when a local Python file is
  used as the reward function on HyperPod, a warning is logged advising the user about the SageMaker naming
  convention.
  - Fixed nova_reward_fn.py test helper — Fixed logic order (check for missing content before checking role), handle
  missing ground_truth gracefully with a score of 0.0 instead of silently skipping.

  2. chore: add log to test to print job arn (5d52f386)

  Added logger.info(f"Training job submitted: {training_job.training_job_arn}") to all integration tests for easier
  debugging of submitted training jobs.


sagemaker-train/tests/integ/train/test_rlvr_trainer_integration.py
```diff
@@ -169,11 +171,13 @@ def test_rlvr_trainer_nova_workflow(sagemaker_session_us_east_1):
         validation_dataset="s3://sagemaker-us-east-1-784379639078/input_data/rlvr-nova/grpo-64-sample.jsonl",
         s3_output_path="s3://sagemaker-us-east-1-784379639078/output/",
         custom_reward_function="arn:aws:sagemaker:us-east-1:784379639078:hub-content/sdktest/JsonDoc/rlvr-nova-test-rf/0.0.1",
+        # custom_reward_function="arn:aws:lambda:us-east-1:784379639078:function:rlvr-nova-reward-function",
```
amazeAmazing, line 175: Nit: Did you mean to include this comment?

amazeAmazing (approved):

sagemaker-train/src/sagemaker/train/rlvr_trainer.py
```diff
@@ -430,7 +430,21 @@ def train(self, training_dataset: Optional[Union[str, DataSet]] = None,
...
         # Extract and validate evaluator ARN
-        evaluator_arn = _extract_evaluator_arn(self.custom_reward_function) if self.custom_reward_function else None
+        # If custom_reward_function is a Lambda ARN, create an Evaluator object first
+        if self.custom_reward_function and isinstance(self.custom_reward_function, str) and _is_lambda_arn(self.custom_reward_function):
+            lambda_arn = self.custom_reward_function
+            evaluator_name = _get_unique_name(f"rlvr-reward-{self._model_name}")
+            logger.info(f"Creating Evaluator from Lambda ARN: {lambda_arn}")
+            evaluator_obj = Evaluator.create(
+                name=evaluator_name,
+                type="RewardFunction",
```
cmahima, line 440: can we also pass role here, otherwise it uses the assumed role

mujtaba1747: What's the issue with uses assumed role.
I believe the role param is execution role. Used by SMTJ for training.

This is more like a preflight resource creation and user's assumed role should be able to create it. Otherwise, Evaluator.create should just error out. Let me know what you think.

sagemaker-train/tests/integ/train/test_rlvr_trainer_integration.py
mujtaba1747: Yes, it can be swapped to test passing a lambda arn directly as well.
I'll add a comment explaining this. Thanks for the callout.

sagemaker-train/src/sagemaker/train/rlvr_trainer.py
cmahima: Generally we would want to use the execution role for the entire job (even if it includes evaluator creation), otherwise users will need to add permissions to two separate roles

mujtaba1747: self.role is the execution role, assumed by sagemaker training job. Passed in training_job params. iam_role_resolver.py in sm-core checks if this role has required permissions. If the permissions aren't enough, it gives a warning and asks to create a new role.

It's a role meant for sagemaker training job, not for submitting the job itself, these are 2 separate roles by design


Ref:
https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/helper/iam_role_resolver.py

https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/helper/iam_policies.py#L18

mujtaba1747: We need 2 separate roles because both have different levels of permissions.
The user's assumed role can have Admin access, create training job, delete s3 etc, which we may not want to pass on to training jobs. Worst case the training job can be malicious and misuse excessive permissions.

That being said, a single role would simplify everything, in that case, the user can just use their Admin role or any other role will their desired level of permissions and pass nothing in self.role. The caller role will be picked up if self.role is None.

ref: https://github.com/aws/sagemaker-python-sdk/blob/master/sagemaker-core/src/sagemaker/core/helper/iam_role_resolver.py#L560

lucasjia-aws (approved):

sagemaker-train/src/sagemaker/train/common_utils/rlvr_reward_verifier.py
```diff
@@ -181,32 +181,34 @@ def verify_reward_function(
...
-        )
-        if function_name_match:
-            function_name = function_name_match.group(1)
-            # Check if function name contains 'SageMaker' (case-insensitive)
-            if not re.search(r"sagemaker", function_name, re.IGNORECASE):
+    if is_nova and isinstance(compute, HyperPodCompute):
+        if is_lambda:
+            # Extract function name from ARN: arn:aws[-*]:lambda:region:account:function:function-name
+            function_name_match = re.search(
+                r"arn:aws[a-zA-Z-]*:lambda:[^:]+:[^:]+:function:([^:]+)", reward_function
```
jam-jee, line 189: Why we are hardcoding pattern here? Why cant we use LAMBDA_REGEX_PATTERN ?

PR #6002, opened 2026-07-08
Feat: show_metrics() and stream_logs() helper functions

ehsu3:
*Issue #, if available:* 'Monitoring training/eval jobs with CW logs' task

*Description of changes:*
* Added cloudwatch_metrics.py with CW log parsing for training_loss, lr, reward_score
* Supports SMTJ and SMHP
* Added `show_metrics()` on BaseTrainer with start_time/end_time, step range filtering
* Added `stream_logs()` on BaseTrainer (SMTJ auto-stops on completion, SMHP via Ctrl+C)
* `show_metrics()` generates separate subplots per metric
* Added pandas to pyproject.toml for generating graphs
* Added 18 unit tests covering parsing, fetching, end-to-end, and streaming
* The two functions can be used on a 'trainer' object that has a job name attached. If the session ends, the user can make a placeholder trainer object and add the job name so it's able to use the functions again. 
* These functions only work training jobs since eval jobs don't have metrics stored in the CW logs and the logs don't have enough information for `stream_logs()` to provide information on progress. However, the latter can be added if we want. 
* Added changes requested from initial PR (clean up code, use logger > print, notebook example, update parameters)

*Tests:*
* Ran both methods on some completed and in-progress jobs to make sure they're able to capture the metrics and logs properly. 

### DIFF
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -1,7 +1,9 @@
 import copy
 import os
+import time
 import yaml
 from abc import ABC, abstractmethod
+from datetime import datetime as _datetime
 from typing import Optional, Dict, Any, List, Union
 import json
 import logging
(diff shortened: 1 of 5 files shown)


sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -268,6 +275,252 @@ def _apply_recipe_to_hyperparameters(self, final_hyperparameters: Dict[str, Any]
 
         return final_hyperparameters
 
+    def show_metrics(
```
amazeAmazing, line 278: Is this something we need to gate to the Nova models? 
The log streaming should still work for OSS models, but Nova models metric parsing is Nova model specific

ehsu3: Yes, good catch! I'm not familiar with the OSS model logs but they'd likely be different from Nova. I'll update the logic there, thanks!

amazeAmazing (approved):

PR #6004, opened 2026-07-08
feat: actionable guidance for removed v2 interfaces

jam-jee:
*Issue #, if available:* N/A

*Description of changes:*

In v3 the v2 top-level modules were **removed**, so importing one fails with a bare, dead-end error:

```
>>> import sagemaker.estimator
ModuleNotFoundError: No module named 'sagemaker.estimator'
```

This PR turns that into actionable migration guidance. (Usage **telemetry** for these events is a follow-up in a separate stacked PR.)

**How it works** — a single meta-path finder (`_RemovedV2ModuleFinder`), registered from `sagemaker-core` (namespace-package init, so it is active on any `import sagemaker.*`, including a cold first import). It is **appended** to `sys.meta_path`, so it runs only after normal import resolution fails and can never shadow a real module. Three guards keep messages truthful:

- **`_V3_REPLACEMENTS`** (34 curated modules) → **precise** message: exact v3 class, copy-pasteable import, and a direct API-docs link.
- **`_REMOVED_V2_MODULES`** (the full v2 top-level surface, derived from `master-v2`) → **generic** "was removed" message for removed modules without a clean 1:1 replacement (internal utilities, and hard-removed frameworks like MXNet/Chainer/RL).
- **Anything else** (typo / hallucinated import) → plain `ModuleNotFoundError`, so we never claim something "was removed" when it never existed.
- **`_KNOWN_V3_TOPLEVEL`** → real v3 packages (`core/train/serve/mlops/lineage/ai_registry`) are never intercepted, even when a sibling package isn't installed.

> Note: an earlier revision used 12 per-module "tombstone" files; those were removed in favor of the single data-driven finder (per review) — same precise messages, one mechanism.

## Resulting UX

**1. Curated removed interface (34 modules) — precise class + import + docs:**
```
>>> import sagemaker.pytorch
ModuleNotFoundError: `sagemaker.pytorch` was removed in the SageMaker Python SDK v3.
Use `ModelTrainer`. (from sagemaker.train import ModelTrainer)
Docs: https://sagemaker.readthedocs.io/en/stable/api/generated/sagemaker.train.model_trainer.html
See https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md for the migration guide.
```
```
>>> import sagemaker.clarify
ModuleNotFoundError: `sagemaker.clarify` was removed in the SageMaker Python SDK v3.
Use the `ProcessingJob` resource. (from sagemaker.core.resources import ProcessingJob)
Docs: https://sagemaker.readthedocs.io/en/stable/api/generated/sagemaker.core.resources.html
See https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md for the migration guide.
```
```
>>> import sagemaker.s3
ModuleNotFoundError: `sagemaker.s3` was removed in the SageMaker Python SDK v3.
Use `s3`. (from sagemaker.core import s3)
Docs: https://sagemaker.readthedocs.io/en/stable/api/generated/sagemaker.core.s3.html
See https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md for the migration guide.
```

**2. Removed v2 module without a 1:1 replacement — generic guidance:**
```
>>> import sagemaker.mxnet
ModuleNotFoundError: `sagemaker.mxnet` was removed in the SageMaker Python SDK v3.
It may have moved to a new location.
See https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md for the migration guide.
```

**3. Unknown name (typo / never existed) — plain error, no misleading guidance:**
```
>>> import sagemaker.foobar
ModuleNotFoundError: No module named 'sagemaker.foobar'
```

*Testing done:*

- `tests/unit/test_removed_v2_modules.py` — **82 tests**, all pass. Covers precise/generic/typo dispatch, finder registration (incl. cold first import in a fresh interpreter), append-ordering, non-shadowing of real packages, and name-set invariants (`_REMOVED_V2_MODULES` disjoint from real v3 packages; all curated keys are in the removed set; a real module under a removed name still wins).
- **Maintenance guards** so replacements can't silently rot: `test_curated_v3_imports_resolve` imports every curated v3 target (`pytest.importorskip`, so it runs in the all-extras lane and skips cleanly in the core-only job); `test_curated_docs_url_matches_import_module_root` asserts the docs URL is derived from the import's module path.
- All 34 replacement imports and docs pages verified against installed v3 and readthedocs.
- `black` (line-length 100) and `flake8` (repo `tox.ini`) clean.

#### General

- [x] I have read the CONTRIBUTING doc
  - Note: additive only (guidance for already-removed modules). No existing public API changes. The finder is registered at `import sagemaker` time via the namespace package init (guarded, best-effort, appended to `sys.meta_path`).
- [x] I used the commit message format described in CONTRIBUTING
- [x] I have passed the region in to all S3 and STS clients that I've initialized as part of this change. (N/A — no clients initialized in this PR)
- [ ] I have updated any necessary documentation
  - Note: `migration.md` lists two paths that do not resolve in shipped v3 (`sagemaker.mlops.pipeline`, `sagemaker.mlops.processing.DataProcessor`); this PR anchors to the verified paths (`sagemaker.mlops.workflow.pipeline`, `sagemaker.core.resources.ProcessingJob`). The guide may need a follow-up fix.

#### Tests

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added unit tests as appropriate to ensure backward compatibility of the changes
- [x] I have checked that my tests are not configured for a specific region or account
- [x] I have used `unique_name_from_base` to create resource names in integ tests (if appropriate) (N/A — unit tests only)
- [x] If adding any dependency in requirements.txt files, I have spell checked and ensured they exist in PyPi (N/A — no new dependencies)


mohanasudhan:
Can we point to samples of ModelTrainer instead of generic message - https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md ?

jam-jee:
> Can we point to samples of ModelTrainer instead of generic message - https://github.com/aws/sagemaker-python-sdk/blob/master/migration.md ?

Updated message thrown. Resulting UX detailed in PR description.

sagemaker-core/src/sagemaker/core/deprecations.py
```diff
@@ -13,13 +13,236 @@
...
+# "removed in v3" message.
+_KNOWN_V3_TOPLEVEL = frozenset({"core", "train", "serve", "mlops", "lineage", "ai_registry"})
+
+# Top-level ``sagemaker.<name>`` modules that existed in v2 but were removed in
+# v3 (some relocated under ``sagemaker.core.*``). Derived from the v2 top-level
+# module surface (the ``master-v2`` branch), minus names that still exist in v3.
+# The fallback finder only emits migration guidance for THESE names, so a typo
+# or hallucinated import (e.g. ``sagemaker.foobar``) gets a plain
+# ``ModuleNotFoundError`` rather than a misleading "was removed" message. V2 is
+# in maintenance, so this surface is effectively frozen.
```
knaresh, line 40: >minus names that still exist in v3.
# The fallback finder only emits migration guidance for THESE names, so a typo
# or hallucinated import (e.g. ``sagemaker.foobar``)


What prevents adding a new module with these names? Is there a unit test?

sagemaker-core/src/sagemaker/core/deprecations.py
```diff
@@ -13,13 +13,236 @@
...
+
+def register_removed_module_finder():
+    """Install the fallback finder for removed v2 modules (idempotent).
+
+    Appends a single ``_RemovedV2ModuleFinder`` to ``sys.meta_path`` so it acts
+    as a last resort. Safe to call multiple times -- it installs at most one
+    instance per process.
+    """
+    if any(isinstance(f, _RemovedV2ModuleFinder) for f in sys.meta_path):
+        return
```
knaresh, line 346: can we add a info log here?

sagemaker-core/src/sagemaker/base_predictor.py
```diff
@@ -0,0 +1,27 @@
...
+
+Importing ``sagemaker.base_predictor`` raises an actionable ``ModuleNotFoundError`` that
+points to the v3 replacement. See the migration guide for details.
+"""
+from __future__ import absolute_import
+
+from sagemaker.core.deprecations import raise_removed_in_v3
+
+raise_removed_in_v3(
+    module="sagemaker.base_predictor",
```
knaresh: why do we need these files as well? can we not infer which import is failing it the  ```find_spec``` again and returng appropriate error mesage? In other words, what is the value addition of having these files?

knaresh (commented):
>Resulting UX

Can you also test how claude behaviour changes with these cahnge?

sagemaker-core/src/sagemaker/core/deprecations.py
jam-jee: Thanks for this point, updated code.
 If v3 ever ships a real top-level module matching one of these names, the real module resolves first and the finder is never consulted. Also added unit tests
- `test_finder_never_shadows_a_real_module_with_removed_name`
- `test_removed_set_does_not_overlap_real_v3_packages` : asserts _REMOVED_V2_MODULES is disjoint from the real v3 top-level packages (_KNOWN_V3_TOPLEVEL), so a live package can never be flagged "removed".

sagemaker-core/src/sagemaker/core/deprecations.py
jam-jee: Added a log line when the finder registers. (used `logger.debug` rather than `info` because `register_removed_module_finder()` runs on every `import sagemaker`, so info log would become too frequent).
happy to bump it to `info` if you'd prefer it more visible.

sagemaker-core/src/sagemaker/base_predictor.py
jam-jee: Thanks for suggestion, updated overall logic, removed all 12 tombstone files and moved their curated data into a single `_V3_REPLACEMENTS` mapping that `find_spec` uses. `find_spec` receives the full module name, looks it up, and emits the same precise message (exact class + import + docs link)

jam-jee:
> Can you also test how claude behaviour changes with these cahnge?

I think this will not immediately impact the codegen by agents, as agents will need to execute code to understand the deprecation signal and corresponding V3 replacement. This will help in long term model's training and during execution time. We will track our codegen results once these changes are released but we may not see a near term win IMO.

mujtaba1747 (approved):

PR #6007, opened 2026-07-09
Fixed vllm and vllm-omni tasks to correct routing logic for HF DLCs

dwarez:
Previously, the set of vLLM and vLLM-omni tasks used in routing for Hugging Face DLCs was wrong, which resulted in the wrong DLCs selection for models like `google/gemma-4-E2B-it` and other  `image-text-to-text` models.


sagemaker-serve/src/sagemaker/serve/constants.py
```diff
@@ -71,14 +71,18 @@
...
+# HuggingFace pipeline tags for models that generate text and should prefer vLLM.
+VLLM_TASKS: Set[str] = {
+    "text-generation",
     "image-text-to-text",
-    "any-to-any",
     "audio-text-to-text",
 }
 
+# HuggingFace pipeline tags for models that can generate multiple output modalities.
+OMNI_TASKS: Set[str] = {
```
pagezyhf, line 82: is it always true?
many any-to-any run on vllm, no? https://recipes.vllm.ai/Google/gemma-4-12B-it

rsareddy0329 (commented):
Seems like a bug fix, Can we add more details in the description on the issue faced currently and how it is being addressed.

dwarez:
@rsareddy0329 in the current routing logic, tasks that should be handled with vllm (like image-text-to-text) are instead routed to vllm-omni, which results in the wrong resolution of the image uri when using the ModelBuilder abstraction. We'd like to leave only the any-to-any task to be run in vllm-omni for the moment, and even with this logic there are edge cases (like google/gemma-4-12B-it which is flagged as any-to-any but should be served with vllm)

sagemaker-serve/src/sagemaker/serve/constants.py
dwarez: iiuc, not all models on the hub flagged as any-to-any would work on vllm. In fact, also gemma-4-12B-it has a dedicated vllm image to make it work `vllm/vllm-openai:gemma4-unified`

rsareddy0329 (approved):

PR #6021, opened 2026-07-13
fix: filter full recipe template from serverless train( )

mujtaba1747:
Title: fix: filter full recipe template from serverless train()

  Summary

  - Serverless CreateTrainingJob limits HyperParameters to 100 members, but the fully resolved
  recipe template can contain hundreds of internal leaf keys the user never touched, causing API
  validation failures.
  - Adds only_user_overrides flag to _apply_recipe_to_hyperparameters() so serverless trainers
  only forward keys the user explicitly provided (direct assignments, overrides dict, or recipe
  YAML).
  - Applies the filter in SFTTrainer, DPOTrainer, and MultiTurnRLTrainer. RLVRTrainer retains
  full recipe passthrough since it renders into a YAML file with no key-count limit.

  Test plan
  - [x] Integ and unit tests to validate new logic
  - [x] Verify serverless training jobs (SFT, DPO, MultiTurnRL) no longer fail with "too many
  hyperparameters" when using a large recipe template
  - [x] Confirm user-provided overrides are still forwarded correctly
  - [x] Validate RLVRTrainer still passes the full resolved recipe (no regression for serverful
  paths)

### DIFF
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -236,19 +236,79 @@ class BaseTrainer(ABC):
             if dotpath:
                 _set_nested_value(resolved, dotpath, value)
 
-    def _apply_recipe_to_hyperparameters(self, final_hyperparameters: Dict[str, Any]) -> Dict[str, Any]:
+    def _get_user_provided_recipe_keys(self) -> set:
+        """Return the set of leaf keys the user explicitly provided.
+
+        Collects keys from every source that represents an explicit user choice:
+
+        * direct hyperparameter assignments (``trainer.hyperparameters.x = val``),
+          tracked in ``hyperparameters._user_set``;
+        * the programmatic ``overrides`` dict passed at construction;
+        * the user recipe YAML file passed at construction.
+
+        These are the keys that should be forwarded to a serverless training job,
+        alongside the Hub override spec. Full recipe-template internal keys that
+        the user never touched are intentionally excluded.
+
+        Returns:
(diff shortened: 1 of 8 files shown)


sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -236,19 +236,79 @@ def _patch_resolved_recipe(self, resolved: Dict[str, Any]) -> None:
...
+            except Exception as e:  # pragma: no cover - best-effort key discovery
+                logger.debug("Could not load user recipe to collect keys: %s", e)
+
+        return keys
+
+
+    def _apply_recipe_to_hyperparameters(
+        self,
+        final_hyperparameters: Dict[str, Any],
+        only_user_overrides: bool = False,
```
jam-jee: nit: can be renamed to `is_serverless` or something like this to improve readability.

jam-jee:
Do we need to test Nova/Serverful/Hyperpod usecases as well. If not manual can we add tests?

mujtaba1747:
We already have tests for hyperpod. Hyperpod and serverful should not be affected by this change as it's gated by the boolean flag.

mujtaba1747:
Serverful tests are being included as part of Zhaoqi's PR. I'm also doing some serverful testing manually for now.

sagemaker-train/src/sagemaker/train/base_trainer.py
mujtaba1747: Agreed, will re-use Compute param.

jam-jee (dismissed):

zhaoqizqwang (approved):

PR #6035, opened 2026-07-16
fix(serve): private hub deploys attach HubAccessConfig and support aliased content names

e-davidson:
## Scope update (2026-07-16)

#6036 (merged after this PR opened) independently fixed the `model_reference_arn` propagation defect with the same change this PR originally carried. This branch has been merged with master: the duplicate propagation block was dropped in favor of the upstream one. **This PR now provides the remaining fix plus the test coverage that would have caught the original regression.**

## Issue

Deploying a JumpStart model from a **private hub** via `ModelBuilder.from_jumpstart_config(hub_name=...)` was broken in two ways on v3.16.0:

**Defect A (fixed upstream in #6036): `HubAccessConfig` never attached to `CreateModel`.**
`get_init_kwargs` resolved `model_reference_arn`, but the builder never copied it, so `_prepare_container_def_base` omitted `S3DataSource.HubAccessConfig.HubContentArn` and the execution role needed `s3:GetObject` on the public `jumpstart-cache-prod-*` bucket.

**Defect B (fixed here): aliased hub content names unsupported.**
`CreateHubContentReference` accepts a custom `HubContentName`, but the SDK always looks up hub content by `model_id`. A reference named differently from the public model id fails with `ResourceNotFound`.

## Changes in this PR

1. **Fix:** optional `hub_content_name` on `JumpStartConfig` / `from_jumpstart_config`; when set, hub content is resolved by its actual name.
2. **Unit tests** (`tests/unit/test_private_hub_artifact_resolution.py`, extends the file from #5985):
   - `model_reference_arn` propagation regression test (guards the #6036 fix; #6036's own tests assert a subset)
   - **CreateModel payload chain test**: drives the real unmocked path `_build_for_jumpstart -> _prepare_container_def_base` and asserts the exact container definition dict carries `HubAccessConfig` (present for hub builds, absent for public). This is the assertion that would have caught the regression that shipped in v3.15.1-v3.16.0.
   - `hub_content_name` lookup + config threading tests
   - Shared setup via module-level `_build_jumpstart_builder()` helper, matching package conventions
3. **Integration tests** (`tests/integ/test_private_hub_artifact_resolution.py`, `slow_test`):
   - `test_deploy_with_no_s3_execution_role`: self-provisions an execution role with **zero S3 permissions**, deploys from a private hub, asserts the created Model carries `HubAccessConfig`. Encodes the service-side brokered-access contract that unit tests cannot reach.
[...]

PR #6039, opened 2026-07-17
fix(serve): support aliased hub content names in private hub deploys

e-davidson:
## Issue

Deploying a JumpStart model from a **private hub** via `ModelBuilder.from_jumpstart_config(hub_name=...)` fails with `ResourceNotFound` when the hub's `ModelReference` was created with a custom `HubContentName`. The `CreateHubContentReference` API accepts an aliased name, but the SDK always looks up hub content by `model_id`, so aliased references are unreachable.

This is the remaining defect after #5985 (hub_arn derivation) and #6036 (`model_reference_arn` propagation / `HubAccessConfig`).

## Fix

Optional `hub_content_name` on `JumpStartConfig`, threaded through `from_jumpstart_config`; when set, `_build_for_jumpstart` resolves hub content by the reference's actual name (3 files, +17 lines).

## Test coverage

This PR also adds the regression coverage that would have caught the `HubAccessConfig` regression that shipped in v3.15.1-v3.16.0:

**Unit** (`tests/unit/test_private_hub_artifact_resolution.py`, extends the #5985 file):
- `model_reference_arn` propagation regression test guarding the #6036 fix (supersedes the subset assertions added there)
- **CreateModel payload chain test**: drives the real unmocked path `_build_for_jumpstart -> _prepare_container_def_base` and asserts the container definition carries `S3DataSource.HubAccessConfig.HubContentArn` for hub builds and omits it for public builds
- `hub_content_name` lookup + config threading tests
- Shared setup hoisted to a module-level `_build_jumpstart_builder()` helper per package convention

**Integration** (`tests/integ/test_private_hub_artifact_resolution.py`, `slow_test`, self-provisioning with full teardown):
- `test_deploy_with_no_s3_execution_role`: creates an execution role with **zero S3 permissions**, deploys from a private hub, asserts the created Model carries `HubAccessConfig` (the service-side brokered-access contract unit tests cannot reach)
- `test_deploy_with_aliased_hub_content_name`: deploys a reference named differently from the model_id

## Verification (red -> green at every layer)

| Layer | Baseline v3.16.0 | With fix |
|---|---|---|
| Unit (17 tests) | 4 discriminating tests FAIL | 17/17 pass |
| Integ, live us-east-1 | 2 FAIL | 2 pass (38s), both with the zero-S3 role |

Live runs used a real private hub + `ModelReference` and an IAM role with no S3 permissions; all resources (hub, references, role, endpoints) cleaned up.

Replaces #6035, which predated #6036 and carried a now-redundant duplicate of the propagation fix.

### DIFF
--- a/sagemaker-core/src/sagemaker/core/jumpstart/configs.py
+++ b/sagemaker-core/src/sagemaker/core/jumpstart/configs.py
@@ -35,6 +35,9 @@ class JumpStartConfig(BaseConfig):
         model_version (Optional[str]): The version of the JumpStart model.
             Defaults to None.
         hub_name (Optional[str]): The name of the JumpStart hub. Defaults to None.
+        hub_content_name (Optional[str]): The name of the hub content reference in the
+            private hub, when it differs from the public model_id. Defaults to None,
+            which means the hub content is expected to be named after the model_id.
         accept_eula (Optional[bool]): Whether to accept the EULA. Defaults to None.
         training_config_name (Optional[str]): The name of the training configuration.
             Defaults to None.
@@ -45,6 +48,7 @@ class JumpStartConfig(BaseConfig):
     model_id: str
     model_version: Optional[str] = None
     hub_name: Optional[str] = None
+    hub_content_name: Optional[str] = None
     accept_eula: Optional[bool] = False
     training_config_name: Optional[str] = None
     inference_config_name: Optional[str] = None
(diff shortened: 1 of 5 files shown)


zhaoqizqwang:
All the added unit tests and integ tests have passed, merging the PR

zhaoqizqwang (approved):

PR #6041, opened 2026-07-17
feat(serve): support fine-tuned models in deployment-config API

Lokiiiiii:
## Summary

Make `ModelBuilder.list_deployment_configs` / `get_deployment_config` / `set_deployment_config` work for fine-tuned (model-customization) models in addition to base/JumpStart models, dispatching internally on `_is_model_customization()`. For fine-tuned models the recipe's published `HostingConfigs` are the source of truth; selection is by instance type (recipe configs are largely unnamed).

The base "deployment config" vs recipe "hosting config" distinction stays internal: fine-tuned configs are normalized to the **same shape** the base response returns (`DeploymentConfigName` + a nested `DeploymentArgs` block with `ImageUri`/`InstanceType`/`Environment`/`ComputeResourceRequirements` and the other base `DeploymentArgs` keys, present as `None` when a recipe doesn't provide them), so callers iterate either pathway identically. `BenchmarkMetrics`/`AccelerationConfigs` are `None` (recipes publish none today); an additive `IsDefault` flag marks the recipe's Default config.

`set_deployment_config(instance_type=...)` applies the whole matching config (image, env, compute requirements) at build time and raises on an unpublished/ambiguous instance type. `list_deployment_configs(instance_type=...)` filters to that instance. Configs published with only `DefaultInstanceType` are matched consistently end-to-end so they aren't silently dropped at build.

## Paired dependency

This is the SDK half of a paired change. The consumer is an Amazon-internal SageMaker agent *model-deployment* skill tracked in internal review **CR-289903643** (`code.amazon.com/reviews/CR-289903643`, not publicly accessible). The two are mutually dependent and are intended to land together.

## Testing

`python3 -m pytest sagemaker-serve/tests/unit/test_recipe_hosting_config_selection.py` (31 tests) plus the base-model regression suites (coverage boost, instance-type inference, nova hosting config, compute requirements) — **102 passed**. Includes a shape-parity test tied to the base `DeploymentArgs`/`DeploymentConfigMetadata` dataclass slots so it fails if base ever adds a field the fine-tuned shape doesn't mirror.

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -393,6 +393,7 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
     _tags: Optional[Tags] = field(default=None, init=False)
     _optimizing: bool = field(default=False, init=False)
     _deployment_config: Optional[Dict[str, Any]] = field(default=None, init=False)
+    _selected_hosting_config: Optional[Dict[str, Any]] = field(default=None, init=False)
 
     shared_libs: List[str] = field(
         default_factory=list,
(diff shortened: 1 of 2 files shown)


jam-jee:
Verified the fortress scan results are not directly relevant to the changes part of this PR.

jam-jee (approved):

jam-jee (approved):

PR #6042, opened 2026-07-17
[Feat]: Job Notifications for SMTJ

ehsu3:
*Issue #, if available:* 'Job completion notifications' task

*Description of changes:*
* Set up job notification infrastructure for SageMaker training (SMTJ) and evaluation jobs
* Enable users to receive SNS notifications on terminal states (completed, failed, stopped) via EB rules. 
* Users have to provide the SNS topic while the SDK sets up an appropriate EB rule based on the user's configurations
* Added `notifications.py` that contains core utilities -- enabling and deleting notification rules, listing existing SDK-managed rules
* Unit tests for notifications 
* Added example in the SMTJ customization notebook.

Usage: 
```
trainer = SFTTrainer(
    model="amazon.nova-lite-v2",
    training_dataset="s3://...",
    notifications={
        "sns_topic_arn": "arn:aws:sns:us-east-1:123:my-topic", # required, user-provided
        "events": ["Completed", "Failed"],       # optional statuses to trigger notifications on
        "job_name_prefix": "my-team-",           # optional
    },
)
```

### DIFF
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -34,7 +34,7 @@ from sagemaker.train.common_utils.finetune_utils import (
 )
 from sagemaker.train.common_utils.metrics_visualizer import plot_training_metrics
 from sagemaker.train.common_utils.mlflow_config_utils import resolve_mlflow_tracking_fields
-from sagemaker.train.common_utils.notifications import enable_notifications, delete_notification_rules, list_notification_rules
+from sagemaker.train.common_utils.notifications import enable_notifications, delete_notification_rule, list_notification_rules
 from sagemaker.train.common_utils.validator import validate_hyperpod_compute
 from sagemaker.train.common_utils.cloudwatch_metrics import fetch_and_plot_metrics, _get_smhp_log_group
 from sagemaker.train.defaults import TrainDefaults
@@ -476,23 +476,19 @@ class BaseTrainer(ABC):
         if not isinstance(notifications, dict):
             raise ValueError(
                 "notifications must be a dict with at least 'sns_topic_arn'. "
-                "Example: {'sns_topic_arn': 'arn:aws:sns:us-east-1:123:my-topic'}"
+                "Example: {'sns_topic_arn': 'arn:aws:sns:us-east-1:123456789012:my-topic'}"
             )
 
         sns_topic_arn = notifications.get("sns_topic_arn")
         if not sns_topic_arn:
             raise ValueError(
                 "notifications config requires 'sns_topic_arn'. "
-                "Example: {'sns_topic_arn': 'arn:aws:sns:us-east-1:123:my-topic'}"
+                "Example: {'sns_topic_arn': 'arn:aws:sns:us-east-1:123456789012:my-topic'}"
             )
-
-        sagemaker_session = TrainDefaults.get_sagemaker_session(
-            sagemaker_session=self.sagemaker_session
-        )
 
         rule_arn = enable_notifications(
             sns_topic_arn=sns_topic_arn,
-            sagemaker_session=sagemaker_session,
+            sagemaker_session=TrainDefaults.get_sagemaker_session(sagemaker_session=self.sagemaker_session),
             events=notifications.get("events"),
             event_bus_arn=notifications.get("event_bus_arn"),
             job_name_prefix=notifications.get("job_name_prefix"),
@@ -500,7 +496,7 @@ class BaseTrainer(ABC):
 
         return rule_arn
 
-    def delete_notification_rules(
+    def delete_notification_rule(
         self,
         rule_arn: str,
         event_bus_arn: Optional[str] = None,
@@ -514,12 +510,8 @@ class BaseTrainer(ABC):
         Returns:
             The name of the deleted rule.
         """
-        sagemaker_session = TrainDefaults.get_sagemaker_session(
-            sagemaker_session=self.sagemaker_session
-        )
-
-        return delete_notification_rules(
-            sagemaker_session=sagemaker_session,
+        return delete_notification_rule(
+            sagemaker_session=TrainDefaults.get_sagemaker_session(sagemaker_session=self.sagemaker_session),
             rule_arn=rule_arn,
             event_bus_arn=event_bus_arn,
         )
@@ -533,12 +525,8 @@ class BaseTrainer(ABC):
         Returns:
             List of dicts with 'name', 'arn', and 'state' for each rule.
         """
-        sagemaker_session = TrainDefaults.get_sagemaker_session(
-            sagemaker_session=self.sagemaker_session
-        )
-
         return list_notification_rules(
-            sagemaker_session=sagemaker_session,
+            sagemaker_session=TrainDefaults.get_sagemaker_session(sagemaker_session=self.sagemaker_session),
             event_bus_arn=event_bus_arn,
         )
 
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -76,6 +76,12 @@ class BaseTrainer(ABC):
         training_image (Optional[str]):
             Custom training container image URI. If not provided, the image is
             auto-resolved from the model's recipe metadata in SageMaker Hub.
+        notifications (Optional[Dict[str, Any]]):
+            Configuration for SNS notifications on job status changes. Requires 'sns_topic_arn'.
+            Optional keys: 'events' ["Completed", "Failed", "Stopped"], 'event_bus_arn',
+            and 'job_name_prefix'. If not specified, no notifications are sent.
+        notification_rule_arn (str):
+            String of the EventBridge rule that is set up when enabling job notifications.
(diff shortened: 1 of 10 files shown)


sagemaker-train/src/sagemaker/train/common_utils/notifications.py
```diff
@@ -0,0 +1,335 @@
...
+            f"EventBridge to publish to it. Add this to the topic's access policy:\n"
+            f'{{"Effect": "Allow", "Principal": {{"Service": "events.amazonaws.com"}}, '
+            f'"Action": "sns:Publish", "Resource": "{sns_topic_arn}"}}'
+        ) from e
+
+    logger.info(f"Notifications enabled: {normalized_events} -> {sns_topic_arn}")
+    return rule_arn
+
+
+def delete_notification_rules(
```
amazeAmazing: nit: this only deletes one rule so it should be `notification_rule` as a name

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -433,6 +439,109 @@ def _show_metrics_cloudwatch(
...
+        if isinstance(getattr(self, 'compute', None), HyperPodCompute):
+            raise NotImplementedError(
+                "Job notifications are not supported for HyperPod compute."
+            )
+
+        # Validate config
+        if not isinstance(notifications, dict):
+            raise ValueError(
+                "notifications must be a dict with at least 'sns_topic_arn'. "
+                "Example: {'sns_topic_arn': 'arn:aws:sns:us-east-1:123:my-topic'}"
```
amazeAmazing: 123 -> should be a full length AWS account for a better example. Here and 486

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -433,6 +439,109 @@ def _show_metrics_cloudwatch(
...
+    def list_notification_rules(
+        self,
+        event_bus_arn: Optional[str] = None,
+    ) -> List[Dict[str, str]]:
+        """List all SDK-created EventBridge notification rules.
+
+        Returns:
+            List of dicts with 'name', 'arn', and 'state' for each rule.
+        """
+        sagemaker_session = TrainDefaults.get_sagemaker_session(
```
amazeAmazing: Can we simplify this?

Can it be 

```
return list_notification_rules(
            sagemaker_session=self.sagemaker_session,
            event_bus_arn=event_bus_arn,
        )
```

sagemaker-train/tests/unit/train/common_utils/test_notifications.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def test_creates_rule_and_target(self):
+        session = self._session()
+        events_client = MagicMock()
+        sns_client = MagicMock()
+        session.boto_session.client.side_effect = lambda svc, **kw: (
+            events_client if svc == "events" else sns_client
+        )
+
+        events_client.put_rule.return_value = {"RuleArn": "arn:aws:events:us-east-1:123:rule/sm-pysdk-notif-abc"}
```
amazeAmazing: nit: should use 123456789012 as account Id here and 126, 130, 147, 205,230-231,241

sagemaker-train/src/sagemaker/train/base_trainer.py
ehsu3: `sagemaker_session` is set as 'None' when a BaseTrainer is initialized and a user doesn't provide a session at init time, so we can't assume `self.sagemaker_session` will have a value. 

`TrainDefaults.get_sagemaker_session` creates a default session if `self.sagemaker_session=None`. I can clean it up a bit by making the session creation in-line though instead of defining a new variable.

sagemaker-train/tests/unit/train/common_utils/test_notifications.py
ehsu3: Thank you for catching these! Will fix in a revision :)

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -102,6 +103,7 @@ def __init__(
         training_image: Optional[str] = None,
         base_model_name: Optional[str] = None,
         disable_output_compression: Optional[bool] = False,
+        notifications: Optional[Dict[str, Any]] = None,
```
mujtaba1747, line 112: Why not add notifications param in the trainer classes (sft, dpo etc) as well? This would make it easier to discover and use.

ehsu3: I didn't think about adding that for visibility -- I can add that so it's easier to discover. I primarily just wanted to centralize the logic/implementation and reduce how much replication we have to do across the different trainers (hence base_trainer), but just adding the param isn't bad. I'll do that!

sagemaker-train/src/sagemaker/train/common_utils/notifications.py
```diff
@@ -0,0 +1,335 @@
...
+        job_name_prefix: Optional job name prefix to filter notifications.
+
+    Returns:
+        The ARN of the created/updated EventBridge rule.
+
+    Raises:
+        ValueError: If sns_topic_arn is invalid or topic doesn't exist.
+        PermissionError: If caller lacks required permissions.
+    """
+    if not sns_topic_arn or not sns_topic_arn.startswith("arn:aws:sns:"):
```
mujtaba1747: arn needs to support all partitions like China, Govcloud etc.
would suggest using regex based matching instead

sagemaker-train/src/sagemaker/train/base_trainer.py
mujtaba1747: Pros of adding to child trainers (SFT, DPO, etc.):
  - Discoverability — users instantiate SFTTrainer, not BaseTrainer. If notifications
  doesn't appear in SFTTrainer.__init__'s signature, it won't show up in IDE
  autocomplete or help().
  - Documentation — each trainer's docstring becomes self-contained; users don't need
  to know the inheritance hierarchy.

mujtaba1747: ^^ Gen AI helped with the Pros

sagemaker-train/src/sagemaker/train/common_utils/notifications.py
ehsu3: I wasn't aware of the other partitions! Will add regex in the next revision :O Thank you!

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -114,6 +116,10 @@ def __init__(
         self.training_image = training_image
         self.base_model_name = base_model_name
         self.disable_output_compression = disable_output_compression
+
+        # Set up notifications if configured
+        if notifications:
```
mujtaba1747, line 128: nit: can we move setup notifs to def train(). Looks like the init is only meant for initializing some variables.

mujtaba1747: unless theres a good reason to have it here...

mujtaba1747: Additionally, looks like setup notifications returns an arn. We can log.debug that arn

Also store it in self.notification_arn or something, in case user wants to retrieve it later.

ehsu3: It's not in `train()` since that's defined for each subclass, so we'd have to duplicate the call in each class haha which is just a bit repetitive. **Open to moving to each subclass if that's the expected/usual pattern!**

I'll add the debug and saving the value for notifications.

sagemaker-train/src/sagemaker/train/common_utils/notifications.py
```diff
@@ -0,0 +1,335 @@
...
+
+    Args:
+        events_client: boto3 EventBridge client.
+
+    Raises:
+        PermissionError: If the caller lacks required EventBridge permissions.
+    """
+    try:
+        events_client.list_rules(NamePrefix=_RULE_NAME_PREFIX, Limit=1)
+    except events_client.exceptions.ClientError as e:
```
mujtaba1747: claude says use botocore.exceptions.ClientError instead of events_client to handle the exception.

events_client may or may not have ClientError. It has service specific exceptions.

Claude says:
```
client.exceptions in boto3 is a factory that exposes service-modeled exceptions —
  things like ResourceNotFoundException, ValidationException, etc. that are defined in
  the EventBridge service model.
```

mujtaba1747: Might want to double check. As claude raised the issue, needs confirmation.

ehsu3: Seems like both might work, but using botocore directly is both more readable and reliable.

sagemaker-train/src/sagemaker/train/base_trainer.py
mujtaba1747: Makes sense @ehsu3.

amazeAmazing (approved):

mujtaba1747 (approved):

PR #6071, opened 2026-07-21
[Fix] Remove task-type from RLVR recipe, update RLVR image selection logic

ehsu3:
*Description of changes:* Fixes two issues causing RLVR/RFT jobs on HyperPod to fail with service validation errors:

1. Strip task_type: `storm_rbs` from recipes (specifically from RFT jobs).
* Fix: In `_extract_recipe_from_helm_template()`, strip `task_type: storm_rbs` via regex. Only strips the storm_rbs value (used by HPCLI), and will preserve other task_types used by other recipes. 

2. Correct RFT training image tag
* Hub resolves the image tag `SM-HP-RFT-V2-latest`, but the correct RFT training image uses `SM-HP-RFT-TRAIN-V2-latest`. Without TRAIN in the tag, the wrong container is pulled.
* Fix: In `_train_hyperpod()`: If the resolved image contains `SM-HP-RFT-` but not `TRAIN`, replace `SM-HP-RFT-` with `SM-HP-RFT-TRAIN-`.

Testing:
* Unit tests added for both fixes
* Manually tested RLVR job submission on HyperPod (successful). 

### DIFF
--- a/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
@@ -1361,6 +1361,9 @@ def _extract_recipe_from_helm_template(template_content: str) -> str:
     with ``---`` separators). The HyperPod CLI expects a single-document recipe YAML.
     This function extracts just the ``config.yaml`` content section.
 
+    For RFT/RLVR recipes, also strips the ``task_type: storm_rbs`` field from the
+    Hub template. 
+
     Args:
         template_content: Raw Helm chart template string from S3.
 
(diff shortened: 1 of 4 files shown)


mujtaba1747:
You need to take this up with the Hub team: 
```
Hub resolves the image tag SM-HP-RFT-V2-latest, but the correct RFT training image uses SM-HP-RFT-TRAIN-V2-latest. Without TRAIN in the tag, the wrong container is pulled.
```

sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
```diff
@@ -1388,7 +1391,13 @@ def _extract_recipe_from_helm_template(template_content: str) -> str:
             "The template format may have changed."
         )
 
-    return textwrap.dedent(recipe_match.group(1)).strip()
+    result = textwrap.dedent(recipe_match.group(1)).strip()
+
+    # Strip task_type: storm_rbs from RFT/RLVR recipes - including it causes service validation failures.
```
mujtaba1747, line 1399: This code is RLVR speicfic but will run for all other training techniques too

mujtaba1747: Can we check training type (rlvr) before doing this check

mujtaba1747: Also this is very nova specific, might want to use is_nova flag too

ehsu3: Yep, I can add that check!

mujtaba1747 (approved):

PR #6077, opened 2026-07-22
fix: Fixing EULA check, relying on HostingEulaUri field

langyn:
Summary
As of sagemaker 3.15.1, ModelBuilder.deploy() refuses to proceed without accept_eula=True for JumpStart models that declare no HostingEulaUri in their hub content document. This blocks deployment of Apache-2.0 models such as huggingface-llm-qwen2-5-7b-instruct and forces users (and downstream agents) to prompt for and "accept" an EULA that doesn't exist. Meta / Llama models — which do have a real HostingEulaUri — are correctly gated. The gate should key off the model metadata; today it appears to apply universally.
Environment
- sagemaker 3.15.1 (also 3.12.0, upgraded to 3.15.1 mid-session in one repro)
- Region: us-west-2
- Python 3.12
- Path: sagemaker.serve.bedrock_model_builder.BedrockModelBuilder and the JumpStart ModelBuilder(...).deploy() path
Expected behavior
The SDK's EULA gate should be driven by the model's hub content:
- If HostingEulaUri is present in the HubContentDocument → require accept_eula=True.
- If it's absent → no acceptance required, deploy() proceeds without the flag.
Actual behavior
deploy() requires accept_eula=True regardless of whether the model declares a HostingEulaUri. The message surfaces to callers as an SDK-level check on all JumpStart models. In two independent test runs, the agent driving deployment concluded (correctly, from what it observed):
 "The ModelBuilder requires accept_eula=True for all JumpStart models regardless of license type (this is an SDK-level gate, not just for Meta models)."


mohamedzeidan2021 (approved):
failing tests unrelated to change

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -2931,7 +2931,12 @@ def _build_single_modelbuilder(
                         "Cannot deploy LORA adapter without base model artifacts."
                     )
                 accept_eula = getattr(self, "accept_eula", None)
-                if not accept_eula:
+                # Only models that declare a hosting EULA (gated models such as
+                # Meta Llama) require explicit acceptance. Ungated models
+                # (Apache-2.0, MIT, etc.) carry no HostingEulaUri and must deploy
```
aviruthen, line 2936: Nit: there's a bit too many comments here! Maybe we can make them shorter so that the code is a bit cleaner...?

aviruthen:
Were you able to test that acceptance/denial of EULA is working correctly post-fix? It's okay if this was tested locally (and not a full integ test)!

aviruthen (approved):

PR #6081, opened 2026-07-22
test(integ): absorb iam:SimulatePrincipalPolicy throttling across integ suites

jam-jee:
## Issue

Integration tests run under `pytest -n auto` (dozens of xdist workers). Many of them resolve/validate an IAM execution role via `resolve_and_validate_role` — invoked from `TrainDefaults.get_role`, `ModelBuilder.build()/.deploy()`, `Pipeline` create/upsert, and the feature-processor scheduler. That function calls the **low-TPS `iam:SimulatePrincipalPolicy`** API (plus a `get_role` for the trust check). With many workers hitting it at once, IAM throttles the request:

```
botocore.exceptions.ClientError: An error occurred (Throttling) when calling the
SimulatePrincipalPolicy operation (reached max retries: 4): Rate exceeded
```

`_evaluate_permissions` only degrades `AccessDenied`/`NoSuchEntity`; `Throttling` re-raises and fails the test — often during **fixture setup**, not just the call phase.

## Fix (test layer only — no SDK source change)

Add an identical, self-contained mitigation to the **serve**, **train**, and **mlops** integ conftests:

1. **`_configure_boto_adaptive_retries`** (autouse, session-scoped) — sets adaptive retries via the `AWS_RETRY_MODE` / `AWS_MAX_ATTEMPTS` environment variables. Env vars apply to **every** boto3 client created in the worker, so the IAM calls ride out transient throttling **whether the resolver falls back to the default session or builds its client from an explicitly-passed `Session`**. Pre-existing env values are restored on teardown.
2. **`pytest_runtest_makereport`** — belt-and-suspenders: converts a residual `SimulatePrincipalPolicy` throttle that survives retries into a **skip** (checked in both `setup` and `call` phases). Throttling on any *other* operation still fails loudly.

### Why not the previous serve-only approach?
The prior serve conftest set a retry `Config` on `boto3.DEFAULT_SESSION`. That only helps when the resolver falls back to the default session — several serve tests (and the train/mlops fixtures) pass their **own** `Session(boto_session=...)`, whose IAM client carried botocore's default 4-attempt policy and was unaffected. Env vars close that gap and let all three suites share one implementation.

### Scope
- New parent `sagemaker-train/tests/integ/conftest.py` covers `train/`, `ai_registry/`, and `jumpstart/`.
- `sagemaker-core` integ is intentionally **not** modified — `resolve_and_validate_role` has no call path there, so it never hits `SimulatePrincipalPolicy`.

## Testing
- All three conftests compile (`py_compile`) and lint clean.
- Unit-checked the throttle-detection helper: matches `Throttling`/`ThrottlingException`/`RequestLimitExceeded` **only** on `SimulatePrincipalPolicy`; ignores other ops, other error codes, and non-`ClientError` exceptions.

## Type of change
- [x] Bug fix / test-infrastructure fix (non-breaking; test-layer only)

## Checklist
- [x] Changes are limited to test conftests; no SDK source or public API is touched
- [x] No new dependencies introduced

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

### DIFF
--- a/sagemaker-mlops/tests/integ/conftest.py
+++ b/sagemaker-mlops/tests/integ/conftest.py
@@ -59,6 +59,84 @@ CONDA_YML_FILE_TEMPLATE = (
 
 
 # ---------------------------------------------------------------------------
+# IAM SimulatePrincipalPolicy throttling mitigation
+#
+# These tests run under ``pytest -n auto`` (dozens of xdist workers). Many of
+# them resolve/validate an IAM execution role via ``resolve_and_validate_role``
+# (e.g. during ``Pipeline`` create/upsert and feature-processor scheduling),
+# which internally calls the low-TPS ``iam:SimulatePrincipalPolicy`` API. With
+# many workers hitting it at once IAM throttles the request, surfacing as
+# ``ClientError: (Throttling) ... Rate exceeded``. This mitigation is a purely
+# test-harness concurrency fix and is intentionally identical to the block in
+# the sagemaker-train / sagemaker-serve integ conftests.
+# ---------------------------------------------------------------------------
+
+# botocore adaptive retry settings for throttling-prone IAM validation calls.
+# Applied via env vars so every client in the worker inherits them, regardless of
+# which boto session the SDK ends up using to build its IAM client.
+_RETRY_MODE = "adaptive"
+_MAX_ATTEMPTS = "10"
+
+# Only tolerate throttling on the IAM permission simulation used during role
+# validation, so unrelated throttling still fails loudly.
+_THROTTLE_ERROR_CODES = ("Throttling", "ThrottlingException", "RequestLimitExceeded")
+_SIMULATE_OP = "SimulatePrincipalPolicy"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _configure_boto_adaptive_retries():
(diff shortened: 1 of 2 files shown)


sagemaker-mlops/tests/integ/conftest.py
```diff
@@ -58,6 +58,84 @@
...
+_MAX_ATTEMPTS = "10"
+
+# Only tolerate throttling on the IAM permission simulation used during role
+# validation, so unrelated throttling still fails loudly.
+_THROTTLE_ERROR_CODES = ("Throttling", "ThrottlingException", "RequestLimitExceeded")
+_SIMULATE_OP = "SimulatePrincipalPolicy"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _configure_boto_adaptive_retries():
```
mujtaba1747, line 87: I think retries are a good idea.

But instead of skipping can we do the following:
Can we go to IAM team to increase rate limit
Or add feature to skip iam validations

lucasjia-aws (approved):

lucasjia-aws:
with batch-call we should not actually go into skip branch right? But we should keep monitoring if tests skips happen a lot with this change

PR #6102, opened 2026-07-24
Fix mlflow (oss models) metrics viz

mujtaba1747:
## Description

Fix related to mlflow (oss models) metrics visualization.

### 1. `plot_training_metrics` failed to render tall figures inline

When plotting many metrics, the combined figure grew too tall for the
inline backend to rasterize, so `display(fig)` produced no output. A
single-metric run also broke because `plt.subplots` returned a bare Axes
that couldn't be flattened.

Changes:
- Return early with a warning when no metric history is found
- Pass `squeeze=False` to `plt.subplots` so `axes` is always a 2D array
  that can be flattened, fixing the single-metric case
- Render the figure to an in-memory PNG and embed it inside a scrollable
  HTML `<div>`, allowing all metrics to display without hitting the
  backend's pixel-count limits

## Testing

- Verified metrics plotting renders correctly for zero, one, and many
  metrics

### DIFF
--- a/sagemaker-train/src/sagemaker/train/common_utils/metrics_visualizer.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/metrics_visualizer.py
@@ -1,7 +1,10 @@
 """MLflow metrics visualization utilities for SageMaker training jobs."""
 
+import io
+import base64
 import logging
 from typing import Optional, List, Dict, Any
+
 from sagemaker.core.resources import TrainingJob
 
 logger = logging.getLogger(__name__)
@@ -209,7 +212,6 @@ def plot_training_metrics(
     import mlflow
     from mlflow.tracking import MlflowClient
     from IPython.display import display
-    import logging
     
     logging.getLogger('botocore.credentials').setLevel(logging.WARNING)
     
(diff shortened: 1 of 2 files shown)


sagemaker-train/src/sagemaker/train/common_utils/metrics_visualizer.py
```diff
@@ -231,13 +231,24 @@ def plot_training_metrics(
...
+    if num_metrics == 0:
+        logger.warning("No metric history found for run %s. Nothing to plot.", run_id)
+        return
+
+    # Build the figure at full size (2 columns, all metrics). The inline
+    # backend can't rasterize this when it gets too tall, so instead of
+    # calling display(fig) directly we render to an in-memory PNG and embed
+    # it inside a scrollable HTML <div>. This lets the notebook display all
+    # metrics without choking on pixel-count limits.
+    import io
```
amazeAmazing: nit: these should move to the top level

sagemaker-train/src/sagemaker/train/common_utils/metrics_visualizer.py
```diff
@@ -246,14 +257,31 @@ def plot_training_metrics(
...
+
+    # Render to PNG buffer for inline display
+    buf = io.BytesIO()
+    fig.savefig(buf, format="png", dpi=80, bbox_inches="tight")
+    plt.close(fig)
+    buf.seek(0)
+
+    # Embed as a scrollable HTML image in the notebook
+    b64 = base64.b64encode(buf.getvalue()).decode()
+    from IPython.display import HTML
```
amazeAmazing, line 277: Should this also be top level?

mujtaba1747: Agreed. Will move it.

amazeAmazing (approved):

zhaoqizqwang (approved):

Issue #6069, opened 2026-07-21
SageMakerClient CustomSession uses default values instead of passed user session


Alex-Mesmer:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**
SageMakerClient builds the sagemaker client from botocore.session.get_session(), ignoring the provided user session resulting in cross-account CreateTrainingJob failures.

In sagemaker-core 2.16.0, SageMakerClient.__init__ ([sagemaker/core/utils/utils.py, ~L395–406](https://github.com/aws/sagemaker-python-sdk/blob/v3.15.1/sagemaker-core/src/sagemaker/core/utils/utils.py#L380-L408)) creates self.sagemaker_client from a fresh default botocore session (botocore.session.get_session() → custom_session) rather than the session argument, under a # TODO: Remove post-launch custom service-model loader. The sagemaker-runtime/featurestore/metrics clients correctly use session. Because the class is a SingletonMeta singleton, the first (default-credential) client is reused process-wide ignoring passed in user sessions.

Impact: callers that pass an explicit boto3 session (e.g. ModelTrainer(sagemaker_session=…) → TrainingJob.create(session=…)) have the control-plane call issued under the ambient default AWS profile instead. When the execution role is in a different account than the default profile, CreateTrainingJob fails with `ValidationException: RoleArn: Cross-account pass role is not allowed`.

**To reproduce**
If a user has multiple AWS profiles (all for different accounts) set up in ~/.aws/config and the default in ~/.aws/credentials is NOT the same account as the intended profile that is used to create the SageMaker Session then CreateTrainingJob  will fail with `ValidationException: RoleArn: Cross-account pass role is not allowed.`

```python
from sagemaker.core.helper.session_helper import Session
from sagemaker.train import ModelTrainer
from sagemaker.train.configs import SourceCode, Compute, InputData
import boto3

sess = Session(boto_session=boto3.Session(profile_name="acct-A", region_name="us-east-2"))
ModelTrainer(
    sagemaker_session=sess,                                   # account A
    role="arn:aws:iam::<ACCOUNT_A>:role/<your-exec-role>",    # account A
    training_image="<ACCOUNT_A>.dkr.ecr.us-east-2.amazonaws.com/img:latest",
    compute=Compute(instance_type="ml.m5.xlarge", instance_count=1),
    source_code=SourceCode(source_dir="src", entry_script="train.py"),
[...]

PR #6107, opened 2026-07-27
fix(sagemaker-core): use caller session for sagemaker control-plane c…

lucasjia-aws:
…lient

SageMakerClient.__init__ built the "sagemaker" control-plane client from a fresh default botocore session (botocore.session.get_session()) instead of the caller-provided session. The custom data loader for the pre-GA Job APIs needs a botocore session to attach to, and the new session it created silently dropped the caller's credentials/profile.

Since SageMakerClient is a singleton, that default-credential client was reused process-wide, so control-plane calls went out under the ambient default AWS profile. When the execution role lived in a different account than the default profile, CreateTrainingJob failed with "RoleArn: Cross-account pass role is not allowed". This regressed V2 behavior and was introduced in the MTRL launch (#5919, first shipped in v3.13.0).

Register the custom data loader on the caller session's underlying botocore session (session._session) and build the sagemaker client from that session, so it keeps the caller's credentials while still loading the custom service model. The other clients already used the passed session and are unchanged.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-core/src/sagemaker/core/utils/utils.py
+++ b/sagemaker-core/src/sagemaker/core/utils/utils.py
@@ -383,8 +383,13 @@ class SageMakerClient(metaclass=SingletonMeta):
         # botocore release. Once these APIs ship in the official botocore
         # service-2.json, this custom loader is unnecessary and the standard
         # session.client("sagemaker", endpoint_url=...) path will work.
+        #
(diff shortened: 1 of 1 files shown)


lucasjia-aws:
related to issue #6069

aviruthen (approved):

Issue #6090, opened 2026-07-23
Transformer.transform() raises pydantic ValidationError ("tags: extra_forbidden") after submitting the job when tags are set


Alex-Mesmer:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**
Transformer.transform() raises a pydantic.ValidationError for TransformJob whenever the Transformer was created with non-empty tags. The error is raised after the transform job has already been submitted to SageMaker: transform() builds and submits the CreateTransformJob request successfully, then reconstructs a local sagemaker.core.resources.TransformJob from the request dict but that resource model declares no tags field and sets extra="forbid", so the round-trip fails when the request contains tags.

Another major issue is that there are no tags attached to the created transformation job.

Consequences: the transform job is actually created on SageMaker, but the SDK call raises and the caller never gets the job handle. The failure is independent of the tag value/shape, any non-empty tags triggers it, because the tags field itself is rejected on the resource model.

A related issue, the shape of the tags between `ModelTrainer` and `Transformer` are not the same.

**To reproduce**
This is exactly the construction transform() performs:
```python
from sagemaker.core.resources import TransformJob
TransformJob(transform_job_name="x", tags=[{"key": "team", "value": "ml"}])
```
Real-world trigger:
```python
from sagemaker.core.transformer import Transformer
t = Transformer(model_name="my-model", instance_count=1, instance_type="ml.m5.large",
                output_path="s3://bucket/out/", tags=[{"Key": "team", "Value": "ml"}])
t.transform(data="s3://bucket/in/", content_type="text/csv")   # job submits, then raises
```
Actual behavior:
```python
pydantic_core._pydantic_core.ValidationError: 1 validation error for TransformJob
tags
  Extra inputs are not permitted [type=extra_forbidden, input_value=[{'key': 'team', 'value': 'ml'}], input_type=list]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```
**Expected behavior**
Transformer.transform() succeeds with tags set and returns the job handle; tags are applied to the transform job.

Root Cause
In [sagemaker-core/src/sagemaker/core/transformer.py](https://github.com/aws/sagemaker-python-sdk/blob/v3.15.1/sagemaker-core/src/sagemaker/core/transformer.py), `Transformer.transform()`:

[...]

PR #6110, opened 2026-07-27
fix(sagemaker-core): drop tags before rebuilding TransformJob in tran…

lucasjia-aws:
…sform()

Transformer.transform() submitted the CreateTransformJob request successfully, then rebuilt a local TransformJob resource from the same request dict. That dict contains a "tags" key (Tags is a member of the CreateTransformJobRequest shape), but the TransformJob resource model has no tags field and sets extra="forbid", so the constructor raised a pydantic ValidationError after the job was already created on SageMaker, leaving the caller without a job handle.

Pop "tags" from the transformed dict before constructing TransformJob, mirroring the existing ProcessingJob fix (#5459). Add a regression test.

*Issue #, if available:*

*Description of changes:*


lucasjia-aws:
related to issue #6090

aviruthen (approved):

jam-jee (approved):

PR #6111, opened 2026-07-27
Update error message on ModelBuilder when deploying from S3 checkpoint

zhaoqizqwang:
Update error message on ModelBuilder when deploying from S3 checkpoint

- Logging training job name when calling `ModelTrainer.train()`
- Updating error message when missing required fields such as `image_uri`, `base_job_name` in ModelBuilder

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -167,6 +167,8 @@ from sagemaker.serve.model_reuse import (
     MODEL_SOURCE_TAG_KEY,
 )
 from sagemaker.core.training.utils import resolve_nova_checkpoint_uri
+from sagemaker.train.common_utils.model_aliases import normalize_model_name
+
 
 _LOWEST_MMS_VERSION = "1.2"
 SCRIPT_PARAM_NAME = "sagemaker_program"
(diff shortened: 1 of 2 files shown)


sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -2900,6 +2910,21 @@ def _build_single_modelbuilder(
                 base_model = model_package.inference_specification.containers[0].base_model
                 if base_model is not None:
                     self._fetch_and_cache_recipe_config()
+            else:
+                # No model package available (e.g. serverful SMTJ training job).
+                # Validate required fields that would normally be auto-resolved.
+                missing_fields = []
+                if not self.image_uri:
+                    missing_fields.append("image_uri")
```
mujtaba1747: do we need the image uri for OSS serverful SMTJ as well?

zhaoqizqwang: This would apply to both Nova and non-Nova models. I have updated the code in new commit to find image_uri from Hub based using `base_model_name`

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -1072,6 +1074,71 @@ def _fetch_and_cache_recipe_config(self):
...
+        if not base_model_name:
+            raise ValueError(
+                "base_model_name is required when deploying a model from an S3 checkpoint "
+                "(e.g. a serverful SMTJ training job) because no model package is available "
+                "to auto-resolve the inference container image. "
+                "Set trainer.base_model_name before calling build()."
+            )
+
+        # If user already provided image_uri, skip hub resolution
+        if self.image_uri:
```
amazeAmazing, line 1085: This should go before base_model_name. If the user already has an image uri, they do not need a base_model_name for resolution.

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -1267,7 +1334,13 @@ def _get_nova_hosting_config(self, instance_type=None):
             return hub_config
 
         model_package = self._fetch_model_package()
-        hub_content_name = model_package.inference_specification.containers[0].base_model.hub_content_name
+        if model_package:
+            hub_content_name = model_package.inference_specification.containers[0].base_model.hub_content_name
+        else:
+            # No model package (e.g. SMTJ trainer): resolve from base_model_name
+            from sagemaker.train.common_utils.model_aliases import normalize_model_name
```
amazeAmazing: This could be a top level import

zhaoqizqwang: Done

sagemaker-serve/src/sagemaker/serve/model_builder.py
zhaoqizqwang: base_model_name validation is still required. I just moved it out of this util function

amazeAmazing (approved):

Issue #6105, opened 2026-07-27
ModelBuilder silently drops `source_code` (no repack) for `image_uri` / `ModelTrainer` builds


Alex-Mesmer:
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**

`ModelBuilder` accepts a `source_code` argument, documented as *"Source code configuration for
custom inference code."* When building a model from an existing **script-mode container**
(`image_uri`) together with a **trained model artifact** (a model data S3 URI), I expect
`ModelBuilder.build()` to **repack** the provided `source_code` into the model artifact — i.e. bake
`code/` into `model.tar.gz` — exactly the way the classic
`sagemaker.model.Model(model_data=…, entry_point=…, source_dir=…)` did, and the way pipelines did
via `_RepackModelStep`.

A self-contained artifact is required for `register()`: a registered model package is persisted and
served on its own, so it cannot depend on an external `SAGEMAKER_SUBMIT_DIRECTORY` pointing at a
transient S3 source tarball.

Instead, for **any** `image_uri`-based build with no `model`/`inference_spec`, `ModelBuilder`
classifies the build as *passthrough* and **silently discards** `source_code`:
`_build_for_passthrough()` resets `self.source_dir = None` and `self.entry_point = None`, so
`is_repack()` returns `False` and no code is ever packaged. The resulting artifact contains the model
data but **none of the inference code**, and `register()` produces a model package that fails to
serve (no `model_fn`/`input_fn`/`predict_fn`/`output_fn`).

This is existing behavior in v2 of this SDK so this is a regression.

**To reproduce**

**Case 1 — `image_uri` + model URI + `source_code` (silent drop):**

```python
from sagemaker.serve import ModelBuilder
from sagemaker.core.training.configs import SourceCode

mb = ModelBuilder(
    image_uri="<script-mode framework or custom (BYO) container>",
    model_path="s3://my-bucket/.../model.tar.gz",   # trained artifact — the "model URI"
    source_code=SourceCode(                          # custom inference code
        source_dir="./code",
        entry_script="inference.py",                 # defines model_fn/input_fn/predict_fn/output_fn
    ),
    role_arn=role,
    sagemaker_session=session,
)
model = mb.build()

# BUG: the model.tar.gz backing `model` does NOT contain code/inference.py.
# `source_code` was dropped and no repack occurred. Registering `model` yields a
# package with no serving code.
```

[...]

PR #6112, opened 2026-07-27
fix(serve): repack source_code for image_uri/ModelTrainer builds

lucasjia-aws:
ModelBuilder classified any image_uri build with a model artifact and source_code (no model/inference_spec) as passthrough and dropped the inference code, so build() produced a model.tar.gz without code/ and register() yielded a package that could not serve. This regressed the classic v2 Model(model_data, entry_point, source_dir) repack behavior.

In _build_for_passthrough(), distinguish a pure image-only deployment from an "image + model artifact + source_code" build. For the latter, keep the source code and bridge model_path to s3_model_data_url so the existing _upload_code(repack=True) path repacks the code into the model artifact, producing a self-contained tarball. Pure-image (Nova/BYOC) passthrough behavior is unchanged. ModelTrainer builds are covered too since they are normalized to model_path before validation.

Also pass script_dependencies (a list derived from SourceCode) to repack_model instead of the deprecated self.dependencies auto-detect dict, which repack_model would otherwise iterate as filesystem paths.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -1851,7 +1851,6 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
             and not self.model
             and not self.inference_spec
             and not getattr(self, "_is_mlflow_model", False)
-            and not self.source_code
         ):
             self._passthrough = True
             return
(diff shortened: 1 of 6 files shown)


lucasjia-aws:
related to issue #6105

aviruthen (dismissed):

nargokul (dismissed):

aviruthen (approved):

PR #6113, opened 2026-07-27
Add create notifications helper method

mujtaba1747:
Summary
Improves SNS/EventBridge notification setup in sagemaker-train and makes permission failures clearer.

File changed: notifications.py

- Add create_notification_topic(topic_name, sagemaker_session): creates an SNS topic and applies an access policy that lets EventBridge in the same account publish to it. The returned ARN can be passed straight to enable_notifications, so callers no longer have to configure the topic policy by hand.

- Add _build_topic_policy(topic_arn, account_id) helper that assembles the policy: the default owner statement (standard SNS actions scoped to the owning account via AWS:SourceAccount) plus an explicit statement granting the EventBridge service principal (events.amazonaws.com) SNS:Publish, also scoped by AWS:SourceAccount.

- Tighten error handling in enable_notifications: replace the broad except Exception (which wrapped every failure in a generic ValueError) with a targeted except ClientError. On AccessDenied/AccessDeniedException it raises a concise PermissionError pointing at the missing events:PutTargets permission; all other errors propagate unchanged.

### DIFF
--- a/sagemaker-train/src/sagemaker/train/common_utils/notifications.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/notifications.py
@@ -166,10 +166,10 @@ def _build_topic_policy(topic_arn: str, account_id: str) -> str:
     """
     policy = {
         "Version": "2008-10-17",
-        "Id": "__default_policy_ID",
+        "Id": "SageMakerNotificationsTopicPolicy",
         "Statement": [
             {
-                "Sid": "__default_statement_ID",
+                "Sid": "SNSTopicAdministration",
                 "Effect": "Allow",
                 "Principal": {"AWS": "*"},
                 "Action": [
--- a/sagemaker-train/src/sagemaker/train/common_utils/notifications.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/notifications.py
@@ -150,6 +150,85 @@ def _validate_sns_topic(sns_client, topic_arn: str) -> None:
         raise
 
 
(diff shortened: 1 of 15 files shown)


sagemaker-train/src/sagemaker/train/common_utils/notifications.py
```diff
@@ -150,6 +150,85 @@ def _validate_sns_topic(sns_client, topic_arn: str) -> None:
...
+    Args:
+        topic_arn: The ARN of the SNS topic.
+        account_id: The AWS account ID that owns the topic.
+
+    Returns:
+        JSON string of the topic access policy.
+    """
+    policy = {
+        "Version": "2008-10-17",
+        "Id": "__default_policy_ID",
```
amazeAmazing: This should be more descriptive like "SageMakerNotificationsTopicPolicy", so if another user on the account is looking at this policy, they would know the source of it / get an sense of what the policy does from the name

sagemaker-train/src/sagemaker/train/common_utils/notifications.py
```diff
@@ -150,6 +150,85 @@ def _validate_sns_topic(sns_client, topic_arn: str) -> None:
...
+        "Statement": [
+            {
+                "Sid": "__default_statement_ID",
```
amazeAmazing: I recommend "SNSTopicAdministration"

amazeAmazing (approved):

PR #6114, opened 2026-07-28
fix(train): raise on expired credentials in show_metrics log fetch

skamal23:
When AWS credentials expire, show_metrics() reported "No CloudWatch logs found for job '<name>'. The job may still be starting, or logs may not be available yet", sending users to debug their training job instead of refreshing credentials.

_fetch_smtj_logs() and _fetch_smhp_logs() each wrapped their CloudWatch Logs call in a bare `except Exception`, logged a warning, and returned an empty list, so an ExpiredTokenException was indistinguishable from a job that genuinely has no logs yet. The empty list then became the misleading ValueError above.

Changes:
- cloudwatch_metrics.py: add _AUTH_ERROR_CODES and _raise_if_auth_error(); apply at describe_log_streams, get_log_events, and filter_log_events; narrow `except Exception` to `except ClientError`
- base_trainer.py: document PermissionError on show_metrics()
- test_cloudwatch_metrics.py: 20 unit tests covering all 7 auth codes on both platforms, ResourceNotFoundException degradation, mid-pagination failure, and a regression guard for the existing "no logs found" path

### DIFF
--- a/sagemaker-train/src/sagemaker/train/common_utils/cloudwatch_metrics.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/cloudwatch_metrics.py
@@ -15,6 +15,7 @@ import logging
 import re
 from datetime import datetime
 from typing import Any, Dict, List, Optional
+from botocore.exceptions import ClientError
 
 from sagemaker.core.training.configs import HyperPodCompute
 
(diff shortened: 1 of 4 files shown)


sagemaker-train/tests/unit/train/common_utils/test_cloudwatch_metrics.py
```diff
@@ -36,6 +37,24 @@
     {"message": "global_step: 2 train_rm_score: 0.72"},
 ]
 
+AUTH_ERROR_CODES = [
```
amazeAmazing: Can we have a common constants in sagemaker-train/src/sagemaker/train/common_utils/constants.py so that they're reusable in both test and src files ? Otherwise, we need to maintain two lists for test coverage

amazeAmazing (changes requested):

sagemaker-train/tests/unit/train/common_utils/test_cloudwatch_metrics.py
skamal23: Good call, moved it. Also rebased onto latest master-nova-follow-ups

amazeAmazing (approved):

sagemaker-train/src/sagemaker/train/common_utils/cloudwatch_metrics.py
```diff
@@ -43,6 +45,32 @@
...
+    Args:
+        error: The ``ClientError`` raised by the CloudWatch Logs client.
+        job_name: Training job name, included in the error message.
+        operation: CloudWatch Logs API that failed, e.g. "DescribeLogStreams".
+
+    Raises:
+        PermissionError: If the error code indicates expired/invalid credentials
+            or missing CloudWatch Logs permissions.
+    """
+    error_code = error.response.get("Error", {}).get("Code", "")
```
mujtaba1747, line 64: nice: https://docs.aws.amazon.com/boto3/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services

sagemaker-train/tests/unit/train/common_utils/test_cloudwatch_metrics.py
```diff
@@ -37,6 +39,14 @@
 ]
 
 
+def _client_error(code, operation):
```
mujtaba1747, line 42: Is _client_error only used by the tests, if yes, pls move this to test file.

mujtaba1747 (approved):

PR #6118, opened 2026-07-28
Instance type validation fix for serverful training jobs

mujtaba1747:
# Add instance_type validation for SMTJ training flow

## Description

Adds `instance_type` validation to the training flow so that unsupported instance types are rejected client-side, before a training job is submitted.

SMTJ (SageMaker Training Job) hub content does not include an `instance_type` enum in its override spec, but the SMHP (SageMaker HyperPod) recipe for the same model/technique does. This change reuses the SMHP override spec to source the allowed instance types and applies them to SMTJ recipe validation, closing a gap where invalid instance types would only fail later at job launch.

## Changes

- **`common_utils/finetune_utils.py`**: Add `_get_smhp_instance_type_enum(...)`, which fetches the `instance_type` enum from the SMHP override spec for a given model/technique/training type. Returns `None` (with a warning) when the enum is unavailable, so validation degrades gracefully instead of failing.
- **`base_trainer.py`**:
  - Add `_validate_instance_type(...)` that resolves the allowed enum and raises a `ValueError` listing the supported values when the provided `instance_type` is not allowed.
  - Wire instance-type validation into the recipe-preparation path and emit a warning when the SMHP recipe does not provide a valid enum (validation skipped).
  - Add a matching warning for the existing instance-count validation when no valid `replicas` enum is available.
- **Tests** (integ):
  - `test_sft_trainer_integration.py`: assert an unsupported instance type (`ml.t3.medium`) raises `ValueError` for a Llama LORA serverful job.
  - `test_sft_trainer_serverful_smtj.py`: assert an unsupported instance type raises `ValueError` for a Nova SMTJ job, and that an out-of-range instance count raises `ValueError` when a valid instance type is used.

## Behavior

- Invalid `instance_type` → `ValueError: Instance type '<type>' is not supported. Allowed values: [...]` raised during `train(dry_run=True)` before any job is submitted.
- When the SMHP recipe provides no enum, validation is skipped and a warning is logged (non-blocking).

## Testing

- Added integration tests covering invalid instance type (serverful + SMTJ) and invalid instance count (SMTJ), each asserting the job fails fast with a clear `ValueError`.

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -167,6 +167,8 @@ from sagemaker.serve.model_reuse import (
     MODEL_SOURCE_TAG_KEY,
 )
 from sagemaker.core.training.utils import resolve_nova_checkpoint_uri
+from sagemaker.train.common_utils.model_aliases import normalize_model_name
+
 
 _LOWEST_MMS_VERSION = "1.2"
 SCRIPT_PARAM_NAME = "sagemaker_program"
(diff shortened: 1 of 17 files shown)


amazeAmazing:
This PR has commits that already exist in the branch.

mujtaba1747:
Creating new pr to avoid stale commit history.

PR #6120, opened 2026-07-29
Fix/SageMaker client construction

lucasjia-aws:
## Description

This PR simplifies SageMaker client construction and includes a couple of small robustness improvements.

**Simplify control-plane client construction.** `SageMakerClient` previously loaded a custom botocore service model (from a bundled `sample/` directory) as a temporary measure to expose the generic Job APIs before they were available upstream. Those Job APIs (`CreateJob` / `DescribeJob` / `ListJobs`) now ship in the public botocore, so this temporary loader is no longer needed. The client now uses the standard `session.client("sagemaker", ...)` path, which is simpler and behaves consistently with the other clients in the constructor.

**Bump the boto3 floor to `>=1.43.20`.** Since the client now relies on the public botocore for the Job APIs, the minimum boto3 version is raised to the first release where those operations are available (`botocore>=1.43.20`), so the `Job.create()` / `Job.get()` / `Job.get_all()` paths resolve correctly on a clean install. This is compatible with the package's existing Python `>=3.10` requirement.

## Testing

- Added unit tests for the env-logging masking behavior.
- Verified the standard `session.client("sagemaker")` exposes `create_job` / `describe_job` / `list_jobs` at the pinned botocore floor.
- Existing `sagemaker-core` unit tests pass.


sagemaker-core/src/sagemaker/core/modules/train/container_drivers/scripts/environment.py
```diff
@@ -62,6 +63,19 @@
 SENSITIVE_KEYWORDS = ["SECRET", "PASSWORD", "KEY", "TOKEN", "PRIVATE", "CREDS", "CREDENTIALS"]
 HIDDEN_VALUE = "******"
 
+SENSITIVE_VALUE_PATTERNS = [
```
mujtaba1747: Env-log masking gap — the container-driver startup logger masks by key name only, so a credential-shaped value under an innocuous key name would be logged in clear text.

The PR is doing regex against some known patterns, but credentials may/may not match that regex.

I wonder what's the best practice to filter out credentials from logs. I would suggest to explore if there any official AWS guidance on this.

sagemaker-core/src/sagemaker/core/utils/utils.py
```diff
@@ -318,6 +318,9 @@ def __bool__(self):
 class SingletonMeta(type):
```
mujtaba1747, line 318: Singleton pinning — SageMakerClient is a process-wide singleton keyed by class only, so the first instance's session/region/endpoint is pinned for all later calls.



I believe this is by design. We don't need to fix it.

sagemaker-core/src/sagemaker/core/modules/train/container_drivers/scripts/environment.py
mujtaba1747: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/protect-sensitive-log-data-types.html

lucasjia-aws:
Re: Claude review

Fixed. `create_job`/`describe_job`/`list_jobs` first ship in botocore 1.43.20 (verified: 1.43.17/1.43.19 lack them, 1.43.20 has them; 1.43.0 and all of 1.42.x lack them). The old `boto3>=1.42.2` pin allowed botocore below that, so the create_job path could break at runtime. Bumped the pin to `boto3>=1.43.20` (requires botocore>=1.43.20). Python is fine — boto3 1.43.x needs >=3.10 and sagemaker-core already requires >=3.10.

Other notes acknowledged: dict-nested value masking is an intentional scope boundary; the CodeQL alert is a false positive (strict tuple-equality assertion, no URL sanitization).

sagemaker-core/pyproject.toml
```diff
@@ -12,7 +12,7 @@ authors = [
 readme = "README.rst" 
 dependencies = [
   # Add your dependencies here (Include lower and upper bounds as applicable)
-    "boto3>=1.42.2,<2.0.0",
+    "boto3>=1.43.20,<2.0.0",
```
mujtaba1747, line 15: Is this intentional?

aviruthen (dismissed):

nargokul (approved):

PR #6128, opened 2026-07-29
fix: update job key used for show_metrics/stream_logs in MTRL

zhaoqizqwang:
Updated logging level and fixed two erorrs:

MTRL's train() stores the job as  self._latest_job  (multi_turn_rl_trainer.py L336) but show_metrics() checks  self._latest_training_job  (base_trainer.py L400) so it raises "no training job found. Call .train() first" immediately after a successful train()

Secondly, MTRL never sets  _customization_technique  and has no entry in AVAILABLE_METRICS (only SFT/CPT/RLVR exist) so then it raises "Could not determine training technique.

### DIFF
--- a/sagemaker-train/src/sagemaker/train/multi_turn_rl_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/multi_turn_rl_trainer.py
@@ -173,6 +173,8 @@ class MultiTurnRLTrainer(BaseTrainer):
             and 'job_name_prefix'. If not specified, no notifications are sent.
     """
 
+    _customization_technique = "MTRL"
+
     def __init__(
         self,
         model: Union[str, ModelPackage],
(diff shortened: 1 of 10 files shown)


mujtaba1747 (approved):

sagemaker-train/src/sagemaker/train/multi_turn_rl_trainer.py
```diff
@@ -173,6 +173,8 @@ class MultiTurnRLTrainer(BaseTrainer):
             and 'job_name_prefix'. If not specified, no notifications are sent.
     """
 
+    _customization_technique = "MTRL"
```
mujtaba1747, line 176: What's the use of this variable?

zhaoqizqwang: It's for consistency. All the other trainers have it and will be called in BaseTrainer utils

PR #6135, opened 2026-07-31
Improve logging, error messages and minor bug fixes

zhaoqizqwang:
see details in commit messages:

- Cache sagemaker_session in BaseTrainer.__init__ to avoid creating
  duplicate sessions on every method call
- Remove redundant role validation (was validating 3x per train() call),
  now validates once in ModelTrainer.__init__
- Demote noisy INFO logs to DEBUG (role validated, stopping condition
  defaults, recipe paths, output compression)
- Add num_lines param to stream_logs() to limit output for long jobs
- Prefix CloudWatch log lines with [CloudWatch] and use print() to
  distinguish container output from SDK logging
- Improve show_metrics() error message when time range yields no logs
- Raise ValueError on AccessDenied in dry_run data path validation
  instead of silently warning
- Improve model_package_group error message to mention compute option
- Move local imports in _train_serverful_smtj to top-level
- Remove redundant get_role() call in ModelTrainer.from_recipe()


sagemaker-serve/src/sagemaker/serve/model_reuse.py
```diff
@@ -218,8 +218,124 @@ def _bedrock_resource_has_tag(bedrock_client, resource_arn: str, tag_value: str)
...
+            "sagemaker:endpoint").
+
+    Returns:
+        The resource ARN if found, empty string "" if no match (signals fast path
+        completed successfully with no results), or None if the tagging API is
+        unavailable (signals caller should fall back to the slow scan).
+    """
+    try:
+        # Extract region from the SageMaker client
+        import boto3 as _boto3
```
mujtaba1747: nit: can we move this to top level import?

mujtaba1747 (approved):

sagemaker-serve/src/sagemaker/serve/model_reuse.py
zhaoqizqwang: Thanks I'll remove this import

Issue #6130, opened 2026-07-30
JumpStartModel ignores tolerate_vulnerable_model and tolerate_deprecated_model: get_jumpstart_configs does not forward the flags


evakravi:
**PySDK Version**
- [x] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**

`get_jumpstart_configs` accepts no `tolerate_vulnerable_model` or `tolerate_deprecated_model` argument. It calls `verify_model_region_and_return_specs` without them, so the callee falls back to its `False` defaults and re-runs the model gate:

```python
# sagemaker/jumpstart/utils.py
def get_jumpstart_configs(
    region: str,
    model_id: str,
    model_version: str,
    config_names: Optional[List[str]] = None,
    sagemaker_session: Optional[Session] = constants.DEFAULT_JUMPSTART_SAGEMAKER_SESSION,
    scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
    model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
    hub_arn: Optional[str] = None,
) -> Dict[str, JumpStartMetadataConfig]:
    model_specs = verify_model_region_and_return_specs(
        region=region,
        model_id=model_id,
        version=model_version,
        sagemaker_session=sagemaker_session,
        scope=scope,
        model_type=model_type,
        hub_arn=hub_arn,
    )   # <-- tolerate_vulnerable_model / tolerate_deprecated_model not forwarded
```

`JumpStartModel.__init__` calls it as its final statement, after `super().__init__()` has already succeeded:

```python
# sagemaker/jumpstart/model.py
self._metadata_configs = get_jumpstart_configs(
    region=self.region,
    model_id=self.model_id,
    model_version=self.model_version,
    sagemaker_session=self.sagemaker_session,
    model_type=self.model_type,
    hub_arn=self.hub_arn,
)
```

So `JumpStartModel(..., tolerate_vulnerable_model=True)` raises `VulnerableJumpStartModelError` anyway. The constructor honors the flag through every other lookup and then discards it on the last line. The same applies to `tolerate_deprecated_model` and `DeprecatedJumpStartModelError`.

The flags are unusable for any flagged model: the object cannot be constructed, so `deploy` is unreachable. The failure is also non-obvious, because the traceback points at `utils.py` inside the gate rather than at anything the caller passed.

**To reproduce**

[...]

PR #6136, opened 2026-07-31
fix: forward tolerance flags from get_jumpstart_configs

evakravi:
*Issue #, if available:* #6130

*Description of changes:*

**Problem**

`tolerate_vulnerable_model=True` and `tolerate_deprecated_model=True` do not work on `JumpStartModel`. Passing them for a flagged model still raises, so the object cannot be constructed and `deploy` is unreachable. The flags are the documented escape hatch for exactly this case, and there is no supported way to reach a flagged model without them.

`JumpStartModel.__init__` honors both flags through every lookup and then discards them on its final statement. That statement calls `get_jumpstart_configs`, which accepted neither flag. It called `verify_model_region_and_return_specs` without them, so the callee fell back to its `False` defaults and re-ran the model gate that the caller had just asked to skip.

The failure is also hard to attribute. The traceback points into `utils.py` inside the gate rather than at anything the caller passed, and it fires after `super().__init__()` has already succeeded.

This blocks any release pipeline that deploys JumpStart models for validation. Such a pipeline has to deploy a model that is flagged, precisely because its job is to test that model.

**Solution**

Add `tolerate_vulnerable_model` and `tolerate_deprecated_model` to `get_jumpstart_configs` and forward both to `verify_model_region_and_return_specs`. Both default to `False`, so every existing caller keeps its current behavior and the gate still fires by default.

`JumpStartModel.__init__` then passes the two attributes it already set earlier in the same constructor, so no new plumbing is needed. A flagged model now resolves empty configs and construction completes, which is the same outcome the flags already produce elsewhere. The returned configs feed only `list_deployment_configs` and the benchmark metrics helpers, so no other path degrades.

`JumpStartEstimator.list_training_configs` had the identical gap and is fixed by the same signature change. Its docstring already documented raising `VulnerableJumpStartModelError`, and that behavior is unchanged for a caller that has not opted into tolerance.

The same gap exists on `master` in `sagemaker-core`, and is fixed by #6137. This PR is the 2.x half.

*Testing done:*

Five tests in `TestConfigs`. Three assert tolerance works for a vulnerable inference model, a deprecated model, and a vulnerable training model. Two assert the gate still raises by default, which is the regression guard for existing callers.

One more in `test_estimator.py` covers `JumpStartEstimator.list_training_configs`, which had no test at all. That gap left the estimator call site uncovered and showed up as a small project coverage drop on the first push.

The three tolerance tests fail before the change, on the dropped argument:

```
E  TypeError: get_jumpstart_configs() got an unexpected keyword argument 'tolerate_vulnerable_model'
E  TypeError: get_jumpstart_configs() got an unexpected keyword argument 'tolerate_deprecated_model'
E  TypeError: get_jumpstart_configs() got an unexpected keyword argument 'tolerate_vulnerable_model'
3 failed, 7 passed
```

All pass after:

```
$ python -m pytest tests/unit/sagemaker/jumpstart/test_utils.py::TestConfigs -q
10 passed
```

No regressions across the JumpStart suite. Baseline `master-v2` and this branch fail the same 4 tests, all unrelated to configs and failing before the change:

```
baseline:     4 failed, 366 passed
this branch:  4 failed, 371 passed
```

The 4 are `test_model_display_benchmark_metrics`, `test_jumpstart_local_metadata_override_specs_not_exist_both_directories`, `test_jumpstart_model_specs`, and `test_inference_configs_parsing`.

Lint on the touched files, using the versions pinned in `requirements/tox`:

```
$ black --check <touched files>     # 24.3.0: 4 files would be left unchanged
$ flake8 <touched files>            # exit 0, no output
$ pydocstyle <touched src files>    # exit 0, no output
$ pylint --rcfile=.pylintrc <touched src files>   # rated 9.98/10, gate is 9.9
```

Pylint reports two `unnecessary-dunder-call` findings in `set_deployment_config` and `set_training_config`. Both are pre-existing and outside this diff, which is additions only.

No integration test is included. Reproducing this requires a model whose published specs carry a vulnerability or deprecation flag, which is not something a test can provision, and the fix is a pure argument-forwarding change fully covered by unit tests.

#### General

#### Tests

### DIFF
--- a/src/sagemaker/jumpstart/estimator.py
+++ b/src/sagemaker/jumpstart/estimator.py
@@ -1206,6 +1206,8 @@ class JumpStartEstimator(Estimator):
             region=self.region,
             scope=JumpStartScriptScope.TRAINING,
             sagemaker_session=self.sagemaker_session,
+            tolerate_vulnerable_model=self.tolerate_vulnerable_model,
+            tolerate_deprecated_model=self.tolerate_deprecated_model,
         )
         return list(configs_dict.values())
 
--- a/src/sagemaker/jumpstart/model.py
+++ b/src/sagemaker/jumpstart/model.py
@@ -401,6 +401,8 @@ class JumpStartModel(Model):
             sagemaker_session=self.sagemaker_session,
             model_type=self.model_type,
             hub_arn=self.hub_arn,
+            tolerate_vulnerable_model=self.tolerate_vulnerable_model,
+            tolerate_deprecated_model=self.tolerate_deprecated_model,
         )
 
     def log_subscription_warning(self) -> None:
--- a/src/sagemaker/jumpstart/utils.py
+++ b/src/sagemaker/jumpstart/utils.py
@@ -1146,8 +1146,19 @@ def get_jumpstart_configs(
     scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
     model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
     hub_arn: Optional[str] = None,
+    tolerate_vulnerable_model: bool = False,
+    tolerate_deprecated_model: bool = False,
 ) -> Dict[str, JumpStartMetadataConfig]:
     """Returns metadata configs for the given model ID and region.
+
+    Args:
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model
+            specifications should be tolerated (exception not raised). If False, raises an
+            exception if the script used by this version of the model has dependencies with known
+            security vulnerabilities. (Default: False).
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated
+            (exception not raised). False if these models should raise an exception.
+            (Default: False).
 
     Raises:
         ValueError: If the script scope is not supported by JumpStart.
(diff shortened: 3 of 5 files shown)


evakravi:
**On the two red checks**

`codestyle-doc-tests` and `integ-tests` fail identically on unmodified `master-v2` and on every other open pull request against this branch. Neither is caused by this change. Reproduced locally with the tool versions pinned in `requirements/tox`:

```
black --check ./          10 files would be reformatted on BOTH this branch and master-v2 (identical file list, none of them mine)
flake8 src/sagemaker tests    8 findings on BOTH (all in tests/unit/test_expected_bucket_owner.py)
pylint --fail-under=9.9   9.93/10 on BOTH
pydocstyle src/sagemaker  0 findings
```

Restricted to the five files this pull request touches:

```
black --check <my 5 files>   5 files would be left unchanged
flake8 <my 5 files>          exit 0, no output
```

`integ-tests` also fails on open pull requests 5917, 5891, and 5890, which do not touch JumpStart.

Every unit-test job passes: py39, py310, py311, py312. `codecov/patch` and `codecov/project` both pass.

lhnealreilly (approved):

rsareddy0329 (approved):

PR #6137, opened 2026-07-31
fix: forward tolerance flags from get_jumpstart_configs

evakravi:
*Issue #, if available:* #6130

*Description of changes:*

**Problem**

`get_jumpstart_configs` accepted no `tolerate_vulnerable_model` or `tolerate_deprecated_model` argument. It called `verify_model_region_and_return_specs` without them, so the callee fell back to its `False` defaults and re-ran the model gate. A caller that had asked to tolerate a flagged model still got `VulnerableJumpStartModelError` or `DeprecatedJumpStartModelError`, so both flags were unusable for exactly the models they exist to allow.

In v3 the reachable caller is `ModelBuilder._ensure_metadata_configs`, which resolves the same configs lazily and had no way to opt out of the gate. A `ModelBuilder` carrying `tolerate_vulnerable_model=True` still failed on a flagged model, even though every other lookup in that class already forwards the flag.

This blocks any release pipeline that deploys JumpStart models for validation. Such a pipeline has to deploy a model that is flagged, precisely because its job is to test that model.

The same gap exists in 2.x and is fixed by #6136. This PR is the v3 half.

**Solution**

Add `tolerate_vulnerable_model` and `tolerate_deprecated_model` to `get_jumpstart_configs` and forward both to `verify_model_region_and_return_specs`. Both default to `False`, so every existing caller keeps its current behavior and the gate still fires by default.

Pass both from `_ensure_metadata_configs`, using the same `getattr` idiom the surrounding call sites already use for these two flags, normalized to a bool so an unset attribute means "do not tolerate". A flagged model then resolves empty configs and the caller proceeds, which is the same outcome the flags already produce elsewhere.

*Testing done:*

Six tests in `TestGetJumpstartConfigs`. Two assert the flags reach the spec lookup that runs the gate, on the inference and training scopes. Two exercise the gate end to end through `JumpStartModelsAccessor.get_model_specs`, confirming a vulnerable model and a deprecated model resolve configs instead of raising. Two are regression guards: the gate still raises for a vulnerable model by default, and the default forwarded value is `False`.

Two more in `TestEnsureMetadataConfigs` cover the `ModelBuilder` caller: tolerance reaches the lookup when set, and defaults to `False` when unset.

Five of the six new `sagemaker-core` tests fail before the change, on the dropped argument:

```
E  TypeError: get_jumpstart_configs() got an unexpected keyword argument 'tolerate_vulnerable_model'
E  TypeError: get_jumpstart_configs() got an unexpected keyword argument 'tolerate_deprecated_model'
E  KeyError: 'tolerate_vulnerable_model'
5 failed, 3 passed
```

Both new `sagemaker-serve` tests fail before the change:

```
E  KeyError: 'tolerate_vulnerable_model'
2 failed, 2 passed
```

All pass after, with no regressions in either package:

```
$ python -m pytest sagemaker-core/tests/unit/test_jumpstart_utils.py -q
174 passed, 1 skipped

$ python -m pytest sagemaker-core/tests/unit -k jumpstart -q
693 passed, 4 skipped, 2841 deselected

$ python -m pytest sagemaker-serve/tests/unit/test_model_builder_utils_additional_gaps.py -q
44 passed, 3 subtests passed
```

Lint introduces nothing new. `flake8` 7.1.2 reports the same 135 findings on the touched files before and after this change, all pre-existing and on lines this diff does not touch. Under `black` 26.3.1 the added lines are already formatted. All four touched files are reformatted by that version on unmodified `master`, so this diff deliberately leaves that pre-existing formatting alone rather than mixing an unrelated reformat into a bug fix.

No integration test is included. Reproducing this requires a model whose published specs carry a vulnerability or deprecation flag, which is not something a test can provision, and the fix is a pure argument-forwarding change fully covered by unit tests.

#### General

#### Tests

### DIFF
--- a/sagemaker-core/src/sagemaker/core/jumpstart/utils.py
+++ b/sagemaker-core/src/sagemaker/core/jumpstart/utils.py
@@ -1196,8 +1196,19 @@ def get_jumpstart_configs(
     scope: enums.JumpStartScriptScope = enums.JumpStartScriptScope.INFERENCE,
     model_type: enums.JumpStartModelType = enums.JumpStartModelType.OPEN_WEIGHTS,
     hub_arn: Optional[str] = None,
+    tolerate_vulnerable_model: bool = False,
+    tolerate_deprecated_model: bool = False,
 ) -> Dict[str, JumpStartMetadataConfig]:
     """Returns metadata configs for the given model ID and region.
+
+    Args:
+        tolerate_vulnerable_model (bool): True if vulnerable versions of model
+            specifications should be tolerated (exception not raised). If False, raises an
+            exception if the script used by this version of the model has dependencies with known
+            security vulnerabilities. (Default: False).
+        tolerate_deprecated_model (bool): True if deprecated models should be tolerated
+            (exception not raised). False if these models should raise an exception.
+            (Default: False).
 
     Raises:
         ValueError: If the script scope is not supported by JumpStart.
@@ -1210,6 +1221,8 @@ def get_jumpstart_configs(
         scope=scope,
         model_type=model_type,
         hub_arn=hub_arn,
+        tolerate_vulnerable_model=tolerate_vulnerable_model,
+        tolerate_deprecated_model=tolerate_deprecated_model,
     )
 
     if scope == enums.JumpStartScriptScope.INFERENCE:
--- a/sagemaker-core/tests/unit/test_jumpstart_utils.py
+++ b/sagemaker-core/tests/unit/test_jumpstart_utils.py
@@ -25,6 +25,7 @@ from sagemaker.core.jumpstart.types import (
     JumpStartBenchmarkStat,
     DeploymentConfigMetadata,
 )
+from sagemaker.core.jumpstart.exceptions import VulnerableJumpStartModelError
 from sagemaker.core.jumpstart.models import HubContentDocument
 from sagemaker.core.helper.pipeline_variable import PipelineVariable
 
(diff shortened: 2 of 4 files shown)


evakravi:
**On the red checks**

`codestyle-doc-tests` fails for all four packages, including `sagemaker-mlops` and `sagemaker-train`, which this change does not touch. Every failure reproduces on unmodified `master` with the tool versions pinned in `requirements/tox`:

```
pylint --rcfile=../.pylintrc src/sagemaker --fail-under=9.9
  sagemaker-core:   8.40/10 on BOTH this branch and master   <- below the 9.9 gate, pre-existing
  sagemaker-serve:  8.98/10 on BOTH                          <- below the 9.9 gate, pre-existing

black --check ./ (sagemaker-core)
  301 files would be reformatted on BOTH

flake8 (sagemaker-core)
  3758 findings on master, 3759 on this branch
```

That single flake8 delta is not a new finding. It is one pre-existing `F811` whose reported line number shifts because this change adds an import above it:

```
master:      tests/unit/test_jumpstart_utils.py F811 redefinition of unused 'TestRemoveEnvVar...' from line 393
this branch: tests/unit/test_jumpstart_utils.py F811 redefinition of unused 'TestRemoveEnvVar...' from line 394
```

`flake8` output for the touched source file is byte-identical on `master` and on this branch (the same three pre-existing findings at lines 24, 57, and 74).

`integ-tests (sagemaker-train)` also fails on open pull request 6131, which does not touch JumpStart.

All four unit-test jobs pass. `codecov/patch` and `codecov/project` both pass.

**Note on formatting.** All four touched files are already reformatted by the pinned `black` 26.3.1 on unmodified `master`. This change leaves that pre-existing formatting alone rather than mixing a repository-wide reformat into a bug fix. The added lines are themselves `black`-clean, verified by diffing the formatter's output before and after the edits.

lhnealreilly (approved):

rsareddy0329 (approved):

PR #6143, opened 2026-08-01
fix(serve): emit OPTION_SPECULATIVE_CONFIG so vLLM enables speculative decoding

ZealSV:
The speculative-decoding path set only OPTION_SPECULATIVE_DRAFT_MODEL, the env var the legacy lmi-dist rolling batcher read. The current LMI container runs the vLLM V1 engine, which reads OPTION_SPECULATIVE_CONFIG (a JSON blob) and ignores OPTION_SPECULATIVE_DRAFT_MODEL. As a result a model requested with speculative_decoding_config deploys and serves, but vLLM leaves speculative decoding off (speculative_config=None) -- the draft model is staged but never used.

Add _set_speculative_draft_model_env, which emits BOTH vars: the legacy path var (back-compat with older containers) and OPTION_SPECULATIVE_CONFIG {"method":"draft_model","model":<path>,"num_speculative_tokens":N}. Wire it into the custom, JumpStart, and deployment-config SD paths. num_speculative_tokens defaults to 5 and is overridable via NumSpeculativeTokens in the config.

Unit: 6 SD tests pass (3 new asserting OPTION_SPECULATIVE_CONFIG + override).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
@@ -138,6 +138,7 @@ from sagemaker.core.remote_function.core.serialization import _MetaData
 from sagemaker.serve.model_server.triton.config_template import CONFIG_TEMPLATE
 
 SPECULATIVE_DRAFT_MODEL = "/opt/ml/additional-model-data-sources"
+DEFAULT_NUM_SPECULATIVE_TOKENS = 5
 _DJL_MODEL_BUILDER_ENTRY_POINT = "inference.py"
 _NO_JS_MODEL_EX = "HuggingFace JumpStart Model ID not detected. Building for HuggingFace Model ID."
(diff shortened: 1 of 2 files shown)


sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
```diff
@@ -138,6 +138,7 @@ def build(self):
 from sagemaker.serve.model_server.triton.config_template import CONFIG_TEMPLATE
 
 SPECULATIVE_DRAFT_MODEL = "/opt/ml/additional-model-data-sources"
+DEFAULT_NUM_SPECULATIVE_TOKENS = 5
```
Lokiiiiii, line 141: This needs to come from the recommendation itself. I am not sure why a default is required here.

Lokiiiiii (changes requested):

PR #6147, opened 2026-08-03
fix(serve): create local model_path dir before using it as download dir

lucasjia-aws:
ModelBuilder.build() assigned the default model_path (/tmp/sagemaker/model-builder/<uuid>) to settings._local_download_dir without creating it on disk, so repack_model()'s _tmpdir() validation raised "Inputted directory ... does not exist" for source_code repack builds. Only use model_path as the local download dir when it is a local path, creating it first; skip s3:// URIs.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -3075,7 +3075,9 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
         self.sagemaker_session = (
             sagemaker_session or self.sagemaker_session or self._create_session_with_region()
         )
-        self.sagemaker_session.settings._local_download_dir = self.model_path
+        if isinstance(self.model_path, str) and not self.model_path.startswith("s3://"):
+            os.makedirs(self.model_path, exist_ok=True)
+            self.sagemaker_session.settings._local_download_dir = self.model_path
 
         client = self.sagemaker_session.sagemaker_client
         client._user_agent_creator.to_string = self._user_agent_decorator(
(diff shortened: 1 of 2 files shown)


lucasjia-aws:
**1. `s3://` behavior change** — Intentional and safe. `settings.local_download_dir` is only ever consumed as the parent dir for `tempfile.mkdtemp()` in `repack_model()` and `fw_utils` — nothing reads it as an S3 URI. The previous unconditional assignment was actually buggy for an S3 `model_path` (e.g. from a `TrainingJob`/`ModelTrainer`): `_tmpdir()` would call `os.path.exists("s3://...")` and raise the same `ValueError`. Leaving it unset for S3 paths falls back to the system temp dir, which is the correct behavior.

**2. `file://` prefix** — Not reachable in practice. `model_path` is always stored as a plain filesystem path; `file://` is only prepended when constructing `s3_upload_path`, so `os.makedirs("file://...")` can't occur here. Agreed it's a latent edge case, but out of scope for this fix.

**3. Regression test** — Added `TestModelBuilderLocalDownloadDir` in `test_model_builder_core.py` covering both branches: a local `model_path` is created on disk and wired into `settings.local_download_dir`, and an `s3://` `model_path` leaves it unset.

lucasjia-aws:
Both test failures are flaky and unrelated to this PR (which only touches `sagemaker-serve/model_builder.py`):

- **mlops** `test_feature_processor_transform_online_only_store_ingestion` — a `ConcurrentModificationException` in the shaded AWS SDK `AttributeMap` when the Feature Store Spark connector builds its client concurrently across executor threads.
- **us-east-1** `test_inspect_ai_upload_benchmarks` — the pipeline execution failed server-side with `Pipeline ... has been modified since your last read` (a concurrent-modification conflict under parallel test runs).

Both are concurrency races in infra/service code, not regressions from this change. Retrying should turn them green.

mujtaba1747 (approved):

PR #6148, opened 2026-08-03
feat(serve): list/filter jobs, deploy from recommendation row, compare benchmarks, DataFrame views

ZealSV:
Adds inference-recommender usability features to sagemaker.serve:

- list_benchmarks(endpoint=...) / list_recommendations(model=..., model_package=...): the ListAI* APIs cannot filter by endpoint/model server-side (those fields are on Describe, not the list summary), so filtering is client-side (list -> describe -> match) bounded by max_results. Chosen because it is the only option the boto APIs allow, and a bounded describe fan-out keeps a broad list from being unbounded.

- ModelBuilder.deploy(recommendation=<row>): accept a recommendation row (mb.recommendations.best or [i]) and resolve it to the existing spec-name/index selection. Chosen as an additive, back-compat param so callers deploy the best row without hand-copying a magic index; existing index/spec_name kwargs still work.

- compare_benchmarks(*results): N-way comparison, first result = baseline, one row per metric x one column per run + a signed delta% column oriented so + is always better. Chosen N-way (not strictly 2-way) because it is barely more code and the delta vs a baseline is the whole point of a comparison utility.

- to_dataframe() on every tabular surface (customer-requested DataFrame view): BenchmarkMetrics, BenchmarkResult, BenchmarkComparison, and both recommendation views (_RecommendationView, _RecommendationsView). Each returns a pandas DataFrame mirroring its printed table, but carrying the extra stats (min/max/p95/stddev) the width-limited text tables drop, and keeping numeric values native (deltas numeric, NaN where undefined) instead of preformatted strings. pandas is imported lazily via _require_pandas() so result.py stays dependency-light (its printed tables are stdlib-only); it is present transitively via sagemaker-core in any real install. __str__ and to_dataframe() share ordering/row-building helpers (_ordered_metric_pairs, _row_records, _delta_value) so the printed table and the frame never drift. BenchmarkResult.to_dataframe() raises on a search/sweep result (no single profile).

Unit: 45 new tests (11 listing, 5 deploy-row, 12 compare, 17 to_dataframe); full recommender suite 199 pass. Integ: test_ai_inference_recommender_enhancements_integration.py chains rec -> deploy(rows.best) InService -> 2 benchmarks -> compare_benchmarks; verified live (1 passed, 70 min) plus a no-GPU listing-plumbing test.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/_recommendation_view.py
+++ b/sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/_recommendation_view.py
@@ -21,9 +21,11 @@ from collections import defaultdict
 from typing import Any, Dict, List, Optional
 
 from sagemaker.serve.ai_inference_recommender.result import (
+    _coerce_numeric,
     _fmt_number,
     _format_table,
     _indent,
+    _require_pandas,
 )
 
 
@@ -63,9 +65,7 @@ class _ExpectedPerformanceMetric:
         return dict(self._stats)
 
     def __repr__(self) -> str:
-        parts = ", ".join(
-            f"{stat}={_fmt_number(v)}" for stat, v in self._stats.items()
-        )
+        parts = ", ".join(f"{stat}={_fmt_number(v)}" for stat, v in self._stats.items())
         unit = f" {self.unit}" if self.unit else ""
         return f"<{parts}{unit}>"
 
@@ -82,9 +82,7 @@ class _ExpectedPerformanceView:
     __slots__ = ("_by_metric",)
 
     def __init__(self, raw_rows: Optional[List[Any]]):
-        by_metric: Dict[str, Dict[str, Any]] = defaultdict(
-            lambda: {"unit": None, "stats": {}}
-        )
+        by_metric: Dict[str, Dict[str, Any]] = defaultdict(lambda: {"unit": None, "stats": {}})
         for row in raw_rows or []:
             metric = getattr(row, "metric", None)
             if not metric:
@@ -139,9 +137,9 @@ class _ExpectedPerformanceView:
         return len(self._by_metric)
 
     def __repr__(self) -> str:
-        return "{" + ", ".join(
-            f"{name}: {metric!r}" for name, metric in self._by_metric.items()
-        ) + "}"
+        return (
+            "{" + ", ".join(f"{name}: {metric!r}" for name, metric in self._by_metric.items()) + "}"
+        )
 
 
 def _to_float(value):
@@ -195,7 +193,6 @@ class _RecommendationView:
     def __str__(self) -> str:
         md = getattr(self._raw, "model_details", None)
         dc = getattr(self._raw, "deployment_configuration", None)
-        ep = getattr(self._raw, "expected_performance", None) or []
 
         config_lines = [
             f"instance_type:        {_safe_str(dc, 'instance_type')}",
@@ -217,14 +214,16 @@ class _RecommendationView:
                 env_lines = [f"  {k} = {v}" for k, v in items]
                 env_block = "\nenv vars ({0}):\n{1}".format(len(items), "\n".join(env_lines))
 
-        perf_rows = []
-        for m in ep:
-            perf_rows.append([
-                _safe_str(m, "metric"),
-                _safe_str(m, "stat"),
-                _fmt_number(_safe_float(m, "value")),
-                _safe_str(m, "unit"),
-            ])
+        # Same records to_dataframe() uses, formatted for display here.
+        perf_rows = [
+            [
+                rec["metric"] if rec["metric"] not in (None, "") else "-",
+                rec["stat"] if rec["stat"] not in (None, "") else "-",
+                _fmt_number(rec["value"]),
+                rec["unit"] if rec["unit"] not in (None, "") else "-",
+            ]
+            for rec in self._perf_records()
+        ]
         perf_table = _format_table(
             headers=["metric", "stat", "value", "unit"],
             rows=perf_rows,
@@ -248,6 +247,30 @@ class _RecommendationView:
     def _repr_pretty_(self, p, cycle):
         # Render the full table in notebooks (Jupyter uses this hook).
         p.text("..." if cycle else str(self))
+
+    def _perf_records(self) -> List[Dict[str, Any]]:
+        """(metric, stat, value, unit) records for this row's expected
+        performance — the rows of the printed table, one per (metric, stat)."""
+        ep = getattr(self._raw, "expected_performance", None) or []
+        return [
+            {
+                "metric": getattr(m, "metric", None),
+                "stat": getattr(m, "stat", None),
+                "value": _safe_float(m, "value"),
+                "unit": getattr(m, "unit", None),
+            }
+            for m in ep
+        ]
+
+    def to_dataframe(self):
+        """Return this recommendation's expected performance as a pandas
+        ``DataFrame`` — the same ``metric``/``stat``/``value``/``unit`` rows the
+        printed ``expected performance`` table shows, one row per (metric, stat).
+
+        Requires pandas.
+        """
+        pd = _require_pandas()
+        return pd.DataFrame(self._perf_records(), columns=["metric", "stat", "value", "unit"])
 
 
 def _safe_str(obj, attr) -> str:
@@ -291,47 +314,114 @@ class _RecommendationsView(list):
         # Render the full table in notebooks (Jupyter uses this hook).
         p.text("..." if cycle else str(self))
 
+    # Column labels for the comparative table / DataFrame, in display order.
+    _TABLE_COLUMNS = (
+        "idx",
+        "spec_name",
+        "instance_type",
+        "instances",
+        "copies/inst",
+        "container",
+        "req/s",
+        "tok/s",
+        "lat_p50",
+        "lat_p90",
+        "lat_p99",
+        "ttft_p50",
+        "itl_p50",
+    )
+    # Metric columns holding numbers; the rest are strings or the idx.
+    _NUMERIC_COLUMNS = (
+        "req/s",
+        "tok/s",
+        "lat_p50",
+        "lat_p90",
+        "lat_p99",
+        "ttft_p50",
+        "itl_p50",
+    )
+
+    def _row_records(self) -> List[Dict[str, Any]]:
+        """One record per row, keyed by ``_TABLE_COLUMNS``, with native values.
+
+        Shared by ``__str__`` and ``to_dataframe()`` so they cannot drift.
+        """
+        records = []
+        for view in self:
+            dc = getattr(view.raw, "deployment_configuration", None)
+            ep = view.expected_performance
+            records.append(
+                {
+                    "idx": view._index,
+                    "spec_name": view.recommendation_spec_name,
+                    "instance_type": getattr(dc, "instance_type", None) if dc else None,
+                    "instances": getattr(dc, "instance_count", None) if dc else None,
+                    "copies/inst": (getattr(dc, "copy_count_per_instance", None) if dc else None),
+                    # None when absent (not "-"), so the DataFrame keeps it as
+                    # missing; __str__ renders the dash.
+                    "container": (
+                        _short_container_tag(dc.image_uri)
+                        if dc is not None and getattr(dc, "image_uri", None)
+                        else None
+                    ),
+                    "req/s": _get_metric_stat(ep, "request_throughput", "avg"),
+                    "tok/s": _get_metric_stat(ep, "output_token_throughput", "avg"),
+                    "lat_p50": _get_metric_stat(ep, "request_latency", "p50"),
+                    "lat_p90": _get_metric_stat(ep, "request_latency", "p90"),
+                    "lat_p99": _get_metric_stat(ep, "request_latency", "p99"),
+                    "ttft_p50": _get_metric_stat(ep, "time_to_first_token", "p50"),
+                    "itl_p50": _get_metric_stat(ep, "inter_token_latency", "p50"),
+                }
+            )
+        return records
+
     def __str__(self) -> str:
         if not self:
             return "Recommendations[0]  (no rows)"
 
         rows = []
-        for view in self:
-            dc = getattr(view.raw, "deployment_configuration", None)
-            ep = view.expected_performance
-            rows.append([
-                f"[{view._index}]",
-                view.recommendation_spec_name or "-",
-                _safe_str(dc, "instance_type"),
-                _safe_str(dc, "instance_count"),
-                _safe_str(dc, "copy_count_per_instance"),
-                _short_container_tag(_safe_str(dc, "image_uri")),
-                _fmt_number(_get_metric_stat(ep, "request_throughput", "avg")),
-                _fmt_number(_get_metric_stat(ep, "output_token_throughput", "avg")),
-                _fmt_number(_get_metric_stat(ep, "request_latency", "p50")),
-                _fmt_number(_get_metric_stat(ep, "request_latency", "p90")),
-                _fmt_number(_get_metric_stat(ep, "request_latency", "p99")),
-                _fmt_number(_get_metric_stat(ep, "time_to_first_token", "p50")),
-                _fmt_number(_get_metric_stat(ep, "inter_token_latency", "p50")),
-            ])
+        for rec in self._row_records():
+            row = []
+            for col in self._TABLE_COLUMNS:
+                value = rec[col]
+                if col == "idx":
+                    row.append(f"[{value}]")
+                elif col in self._NUMERIC_COLUMNS:
+                    row.append(_fmt_number(value))
+                else:
+                    row.append("-" if value in (None, "") else str(value))
+            rows.append(row)
 
-        table = _format_table(
-            headers=[
-                "idx", "spec_name", "instance_type",
-                "instances", "copies/inst",
-                "container",
-                "req/s", "tok/s",
-                "lat_p50", "lat_p90", "lat_p99",
-                "ttft_p50", "itl_p50",
-            ],
-            rows=rows,
-        )
+        table = _format_table(headers=list(self._TABLE_COLUMNS), rows=rows)
 
         return (
             f"Recommendations[{len(self)}]  (.best = top row; index by [N] for full detail)\n"
             f"{table}\n"
             f"lat/ttft/itl in ms; req/s = requests/sec; tok/s = output tokens/sec"
         )
+
+    def to_dataframe(self):
+        """Return the recommendations as a pandas ``DataFrame`` — one row per
+        recommendation, columns matching the printed comparative table
+        (``instance_type``, ``instances``, ``req/s``, ``lat_p50``, ...), indexed
+        by the recommendation index ``idx``.
+
+        Latency columns are milliseconds; ``req/s`` is requests/sec and ``tok/s``
+        is output tokens/sec, same as the printed table's footnote. Numeric
+        columns hold real numbers (``NaN`` where a metric is absent), not
+        preformatted strings.
+
+        Requires pandas.
+        """
+        pd = _require_pandas()
+        columns = [c for c in self._TABLE_COLUMNS if c != "idx"]
+        records = self._row_records()
+        frame = pd.DataFrame(
+            [{c: rec[c] for c in columns} for rec in records],
+            columns=columns,
+            index=pd.Index([rec["idx"] for rec in records], name="idx"),
+        )
+        return _coerce_numeric(pd, frame, self._NUMERIC_COLUMNS)
 
 
 def _get_metric_stat(ep_view, metric_name: str, stat: str):
--- a/sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
+++ b/sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
@@ -1,0 +1,266 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+"""List and filter AI benchmark / recommendation jobs.
+
+The underlying ``ListAIBenchmarkJobs`` / ``ListAIRecommendationJobs`` APIs only
+filter by name substring, status, and creation-time window — the endpoint a
+benchmark targeted, or the model a recommendation ran on, live on the full
+Describe response, not the list summary. So ``endpoint`` / ``model`` /
+``model_package`` filtering is done client-side: the native filters narrow the
+list server-side, then each candidate is described (hydrated by sagemaker-core's
+resource iterator) and matched on the nested field. Two bounds keep the cost in
+hand: ``max_results`` caps the matches returned, and ``max_scan`` caps how many
+candidates are described — so a rarely-matching filter cannot fan out across the
+whole account.
+"""
+from __future__ import absolute_import
+
+import logging
+from typing import List, Optional, Union
+
+import boto3
+
+from sagemaker.core.helper.session_helper import Session
+from sagemaker.core.telemetry.constants import Feature
+from sagemaker.core.telemetry.telemetry_logging import _telemetry_emitter
+from sagemaker.serve.ai_inference_recommender.jobs import BenchmarkJob, RecommendationJob
+
+logger = logging.getLogger(__name__)
+
+# Default cap on how many matching jobs are returned.
+DEFAULT_MAX_RESULTS = 100
+# Default cap on how many candidates are described (each is one Describe, issued
+# by the resource iterator). Bounds the fan-out when a client-side filter
+# (endpoint / model / model_package) matches rarely.
+DEFAULT_MAX_SCAN = 1000
+# How many undescribable candidates to log individually before deferring the
+# rest to the end-of-scan summary (avoids one warning per candidate).
+_MAX_SKIP_WARNINGS = 5
+
+
+def _endpoint_matches(job, endpoint: str) -> bool:
+    """True if a benchmark job targeted ``endpoint`` (by name or ARN)."""
+    target = getattr(job, "benchmark_target", None)
+    ep = getattr(target, "endpoint", None) if target else None
+    identifier = getattr(ep, "identifier", None) if ep else None
+    if not identifier:
+        return False
+    # identifier may be a name or an ARN; match either the exact value or the
+    # name suffix of an ARN (endpoint/<name>).
+    return endpoint == identifier or identifier.endswith(f"/{endpoint}")
+
+
+def _model_matches(job, model: str) -> bool:
+    """True if a recommendation job ran on ``model`` (its source S3 URI)."""
+    source = getattr(job, "model_source", None)
+    s3 = getattr(source, "s3", None) if source else None
+    s3_uri = getattr(s3, "s3_uri", None) if s3 else None
+    if not s3_uri:
+        return False
+    return model == s3_uri or s3_uri.rstrip("/") == model.rstrip("/")
+
+
+def _model_package_matches(job, model_package: str) -> bool:
+    """True if a recommendation job is associated with ``model_package``.
+
+    Matches either the output model-package group the job registers into, or a
+    model-package ARN produced on one of the job's recommendation rows.
+    """
+    output = getattr(job, "output_config", None)
+    group = getattr(output, "model_package_group_identifier", None) if output else None
+    if group and (model_package == group or group.endswith(f"/{model_package}")):
+        return True
+    for row in getattr(job, "recommendations", None) or []:
+        details = getattr(row, "model_details", None)
+        arn = getattr(details, "model_package_arn", None) if details else None
+        if arn and (model_package == arn or arn.endswith(f"/{model_package}")):
+            return True
+    return False
+
+
+def _collect(iterator, predicate, max_results: int, max_scan: int, subclass) -> list:
+    """Keep candidates from ``iterator`` matching ``predicate``, re-typed to
+    ``subclass`` (so ``show_result`` is available).
+
+    The iterator hydrates each object as it yields it, so this loop does not
+    Describe again. ``max_results`` bounds matches returned; ``max_scan`` bounds
+    candidates examined. A candidate that fails to Describe is skipped, not
+    fatal. All three limits are logged when hit.
+    """
+    matches: list = []
+    scanned = 0
+    skipped = 0
+    hit_result_cap = False
+    it = iter(iterator)
+    while scanned < max_scan:
+        # next() drives the iterator's per-object Describe, so a failure for one
+        # candidate is caught here rather than aborting the listing.
+        try:
+            job = next(it)
+        except StopIteration:
+            break
+        except Exception as exc:  # noqa: BLE001 - per-job hydration is best-effort
+            skipped += 1
+            scanned += 1
+            # Log the first few individually for diagnosis; the rest are covered
+            # by the summary below, so a fully-denied role does not emit one
+            # warning per candidate up to max_scan.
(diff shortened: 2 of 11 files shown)


sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -5519,6 +5524,22 @@ def deploy(
...
+        # spec name), so callers never hand-copy a magic index.
+        if recommendation is not None:
+            if recommendation_spec_name is not None or recommendation_index:
+                raise ValueError(
+                    "Pass only one of `recommendation`, `recommendation_spec_name`, "
+                    "or `recommendation_index` to deploy()."
+                )
+            resolved_spec = getattr(recommendation, "recommendation_spec_name", None)
+            if resolved_spec is not None:
+                recommendation_spec_name = resolved_spec
```
Lokiiiiii: **F1 (blocker): this discards the row's index, so `deploy(recommendation=row)` can provision a different row than the one passed.**

The row's `_index` is read only in the `else` branch below — when the row has *no* spec name. Whenever it *does* have one, `recommendation_index` keeps its signature default of `0` and the row's true position is never forwarded. `_deploy_recommendation` then resolves a spec name by taking `matches[0]` (line 5215).

If two rows share an `inference_specification_name`, the wrong one is deployed. Reproduced with two rows differing in ModelPackage and instance type:

```
deploy(recommendation=view[1])    # caller wants pkg/2, ml.p4d.24xlarge x8
  WARNING recommendation_spec_name='dup-spec' matched 2 recommendations; deploying the first.
  SELECTED -> model-package/g/1   # row 0: ml.g5.2xlarge x1

deploy(recommendation_index=1)    # same intent, index path
  SELECTED -> model-package/g/2   # correct
```

`rec` drives ModelPackage, container, env vars, instance type, instance count and copy count (lines 5228-5363), so all of those come from the wrong row. The endpoint reaches `InService` on hardware the caller did not choose.

Duplicate spec names are not hypothetical — `AIRecommendationModelDetails` carries `instance_details` as a *list* of instance configurations per spec, so one spec fanning out to N rows is the normal shape. This PR's own `TestDeployRecommendationSpecNameMultiMatch` asserts the duplicate case, and `TestDeployRecommendationRowObject._make_builder` uses distinct spec names `"A"`/`"B"` — so both halves are tested and never composed, which is why all 199 tests pass.

Worth noting the warning at line 5209 advises "Use `recommendation_index` to pick a specific one", but `recommendation=` and `recommendation_index=` are mutually exclusive (line 5533), so the remedy it suggests is unreachable from the path that emits it.

Suggested direction: forward `recommendation_index = getattr(recommendation, "_index", 0)` unconditionally and let the index win when a row object was passed, using the spec name only as a fallback. The caller handed you an unambiguous identifier; keeping it makes the feature exact.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,207 @@
...
+    matches: list = []
+    scanned = 0
+    for job in iterator:
+        job.__class__ = subclass
+        if predicate is None:
+            matches.append(job)
+        else:
+            # The list summary lacks the nested field; hydrate before matching.
+            try:
+                job.refresh()
```
Lokiiiiii: **F2 (major): this is a second `Describe` per candidate — the iterator already refreshed the object.**

`sagemaker-core`'s `ResourceIterator.__next__` refreshes every object it yields, gated only on `hasattr`:

```python
# sagemaker/core/utils/utils.py:464-465
if hasattr(resource_object, "refresh"):
    resource_object.refresh()
```

`AIBenchmarkJob.refresh` and `AIRecommendationJob.refresh` both exist and neither caches, so objects arriving here are already hydrated. Measured with the real `get_all` and only the boto client patched, attributing each call by stack frame:

```
5 candidates -> 10 DescribeAIBenchmarkJob calls
by origin: {'ITERATOR.__next__': 5, '_collect': 5}
```

This also means the comment above ("The list summary lacks the nested field") no longer describes reality, and the `predicate is None` path still pays a full Describe per job inside the iterator — so the "hydrate only when filtering" optimisation the module docstring describes does not hold either.

Combined with F3 this doubles an already-unbounded fan-out. Suggested direction: drop this call and note in the docstring that hydration is the iterator's job, or if you want control over it, describe explicitly through the session's client instead (which would also address F5).

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,207 @@
...
+            except Exception as exc:  # pragma: no cover - best-effort hydration
+                logger.debug("Skipping %s; could not describe it: %s", job, exc)
+                continue
+            if predicate(job):
+                matches.append(job)
+        if len(matches) >= max_results:
```
Lokiiiiii, line 128: **F3 (major): `max_results` bounds the number of matches, not the Describe fan-out — contrary to three docstrings.**

This breaks on `len(matches)`, so when the predicate matches rarely the loop keeps describing. Measured against `_collect` directly:

```
stream of 500 candidates, max_results=2, only the last matches
  matches returned : 1
  refresh() calls  : 500
```

The module docstring states "`max_results` bounds that Describe fan-out" (line 22), and both public functions repeat it as "(and, when `endpoint` is set, on how many are described)" (lines 132-134, 168-170). None of that holds.

`scanned` (line 90, incremented at line 111) is assigned and never read anywhere — which reads like the scan budget this was meant to have, left unwired. It also sits after the `break`, so it would under-count by one even if something did read it.

Practical impact: `list_benchmarks(endpoint="typo-in-name")` on an account with 5k jobs issues 5k Describes with F2 doubling it to ~10k — minutes of latency and near-certain throttling, for a call that returns `[]`.

Suggested direction: bound the scan as well as the results — `if scanned >= max_scan: break` with `max_scan` either derived from `max_results` or its own parameter — and log the scanned count so the cost is visible. If instead the intent is that `max_results` caps results only, the three docstrings need to say so, and callers need some other way to bound the fan-out.

A unit test asserting `refresh.call_count` for a non-matching filter would pin this; the current `test_max_results_caps_output` uses filter-free jobs, so it exercises the one path where the cap and the fan-out coincide.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,207 @@
...
+    """Describe candidates from ``iterator`` and keep those matching ``predicate``.
+
+    Each candidate is re-typed to ``subclass`` (so ``show_result`` is available)
+    and, when a predicate is given, refreshed to hydrate the nested fields the
+    list summary omits. Stops once ``max_results`` matches are collected; logs
+    when the scan is truncated so a silent cap never reads as "all matches".
+    """
+    matches: list = []
+    scanned = 0
+    for job in iterator:
```
Lokiiiiii: **F4 (major): the first `Describe` happens on this line, outside the `try` below — so one inaccessible job aborts the whole listing.**

Because `ResourceIterator.__next__` refreshes each object as it yields it (`sagemaker/core/utils/utils.py:465`), the initial Describe for every job executes at this `for` statement, not at the guarded `job.refresh()` on line 98. Verified against the real iterator:

```
RAISED OUT of list_benchmarks: AccessDeniedException ...
  File ".../listing.py", line 91, in _collect      <- this line
  File ".../sagemaker/core/utils/utils.py", line 465, in __next__
  File ".../sagemaker/core/resources.py", line 383, in refresh
```

The remaining jobs are never examined. Control comparison with a plain `iter(list)` that does not refresh on yield: the bad job is skipped and the others are returned, i.e. the `try`/`except` works — but only for iterators that do not hydrate, which the real one does.

So the `except Exception` on line 99 is structurally unreachable for the failures most likely to occur, while reading as though it makes this function resilient to them. One job the caller lacks `DescribeAIBenchmarkJob` on anywhere in the account makes `list_benchmarks()` raise instead of returning the other N−1.

Suggested direction: wrap the iteration itself, e.g. drive it with `next()` inside the `try`, so a per-job Describe failure is handled in one place regardless of which layer issued the call.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,207 @@
...
+@_telemetry_emitter(
+    feature=Feature.INFERENCE_RECOMMENDER, func_name="ai_inference_recommender.list_benchmarks"
+)
+def list_benchmarks(
+    *,
+    endpoint: Optional[str] = None,
+    status: Optional[str] = None,
+    name_contains: Optional[str] = None,
+    max_results: int = DEFAULT_MAX_RESULTS,
+    sagemaker_session: Optional[Session] = None,
```
Lokiiiiii: **F5 (major): passing the type this parameter is annotated with raises a `ValidationError`.**

The annotation resolves to `sagemaker.core.helper.session_helper.Session` (imported line 28), but `get_all` is wrapped in pydantic `validate_call` and requires `boto3.session.Session`. These are unrelated types — `isinstance(Session(), boto3.session.Session)` is `False`. Verified:

```
>>> list_benchmarks(sagemaker_session=sagemaker.core.helper.session_helper.Session())
ValidationError: 1 validation error for AIBenchmarkJob.get_all
session
  Input should be an instance of Session [type=is_instance_of,
    input_value=<sagemaker.core.helper.session_helper.Session object ...>]

>>> list_benchmarks(sagemaker_session=boto3.session.Session(region_name="us-west-2"))
   accepted
```

The docstring says only "Optional session; a default is created if omitted", so a reader has nothing to warn them off the annotated type — and `sagemaker.Session()` is the idiomatic thing to reach for. The default `None` path works, so this only bites callers who use the parameter at all.

Same on `list_recommendations` (line 162).

Suggested direction: annotate `Optional[boto3.session.Session]` and say so in the docstring, or accept both and unwrap a `sagemaker` `Session` to its `boto_session` before handing it to `get_all`.

Related caveat worth documenting either way: `SageMakerClient` is a `SingletonMeta` keyed only on the class, so if the singleton was already built earlier in the process, the session passed here is ignored and the call targets the earlier one.

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -5519,6 +5524,22 @@ def deploy(
...
+            else:
+                recommendation_index = getattr(recommendation, "_index", 0)
```
Lokiiiiii: **F6 (major): with two `getattr` defaults and no type check, any non-row value silently deploys row 0.**

Line 5538's `getattr(..., None)` misses on anything that isn't a `_RecommendationView`, and this line's `getattr(..., 0)` then falls back to index 0. Measured across plausible inputs:

| `recommendation=` | resolves to | deploys |
|---|---|---|
| `"spec-b"` (a spec-name string) | `spec=None, index=0` | **row 0** |
| raw `AIRecommendation` shape for row 2 | `spec=None, index=0` | **row 0** |
| `2` (an int index) | `spec=None, index=0` | **row 0** |
| `{"recommendation_spec_name": "B"}` | `spec=None, index=0` | **row 0** |

The first two are easy mistakes. Passing the spec-name string is natural because the sibling parameter takes one; and the raw shape is what you get from the service object rather than the view, where the field is `model_details.inference_specification_name`, not `recommendation_spec_name`. Either way the caller gets a successful deploy of a configuration they did not pick, with no warning, and `recommendation: Optional[Any]` (line 5427) means no type checker flags it.

Suggested direction: guard on the shape you actually require — `hasattr(recommendation, "recommendation_spec_name") and hasattr(recommendation, "_index")`, or an `isinstance` check against the view class — and raise `TypeError` naming what was passed. Turning all four rows above into a clear error costs two lines.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/_recommendation_view.py
```diff
@@ -291,48 +315,102 @@ def _repr_pretty_(self, p, cycle):
...
 
+    def to_dataframe(self):
+        """Return the recommendations as a pandas ``DataFrame`` — one row per
+        recommendation, columns matching the printed comparative table
+        (``instance_type``, ``instances``, ``req/s``, ``lat_p50``, ...), indexed
+        by the recommendation index ``idx``.
+
+        Latency columns are milliseconds; ``req/s`` is requests/sec and ``tok/s``
+        is output tokens/sec, same as the printed table's footnote. Numeric
+        columns hold real numbers (``NaN`` where a metric is absent), not
```
Lokiiiiii, line 411: **F7 (major): when a metric is absent from every row, the column is `object` dtype holding `None`, not `NaN` — and sorting raises.**

pandas infers `object` from an all-`None` list, so the promise here does not hold for the sparse case. Verified on two rows with no `expected_performance`:

```
dtypes:  req/s object, tok/s object, lat_p50 object, lat_p90 object,
         lat_p99 object, ttft_p50 object, itl_p50 object
>>> df.nlargest(1, "req/s")
TypeError: Column 'req/s' has dtype object, cannot use method 'nlargest' with this dtype
```

`BenchmarkComparison.to_dataframe`'s docstring makes the same claim about deltas (`result.py:724-725`) and has the same behaviour when every delta is undefined.

This matters because sorting, `nlargest`, and `.mean()` are the reason a DataFrame was asked for, and the columns most likely to be entirely absent are exactly the ones a customer would rank on. `test_numeric_columns_are_native_and_missing_is_nan` asserts `pd.isna()` on a *partially* populated column, which pandas does coerce to `float64` — so it passes while this case is broken.

Suggested direction: coerce the numeric columns explicitly, e.g. `.astype({c: "float64" for c in numeric_cols})` or `pd.to_numeric(..., errors="coerce")` after construction, and add a case where a metric is missing from every row.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/result.py
```diff
@@ -524,7 +607,195 @@ def _format_table(headers, rows) -> str:
...
-        for row in str_rows
+        "  ".join(cell.ljust(widths[i]) for i, cell in enumerate(row)) for row in str_rows
     )
     return f"{header_line}\n{sep_line}\n{body}"
+
+
+# Metrics where a higher value is better (throughput); everything else in
+# _KEY_METRIC_FIELDS is a latency/duration where lower is better. Used to label
+# each run's delta vs. the baseline as an improvement or a regression.
+_HIGHER_IS_BETTER = frozenset(
```
Lokiiiiii, line 622: **F8 (major): this allowlist gives every metric outside it a confidently signed delta, including metrics with no meaningful direction.**

`_delta_value` treats anything not in this set as lower-is-better and flips the sign, but `_metric_names()` deliberately includes *every* metric present in any run — not just `_KEY_METRIC_FIELDS`. The comment above says "everything else in `_KEY_METRIC_FIELDS`", which is a narrower set than the code actually applies. Verified:

```
BenchmarkComparison
  baseline: base  |  stat: avg  (+Δ = better)
  metric                  unit  base   new  Δ% new
  output_sequence_length  -     128.0  64    +50.0%
  benchmark_duration      -     60     30    +50.0%
  http_foo                -     1      2    -100.0%
```

The first row is the problem: the model produced half as many output tokens and the table reports a 50% improvement, under a header that asserts `+Δ = better` unconditionally. `input_sequence_length`, `output_sequence_length` and any `http_*` counter have no better direction; presenting a signed delta for them is worse than presenting none, because it looks authoritative.

Suggested direction: make direction tri-state — higher-is-better, lower-is-better, unknown — and for unknown either leave the delta cell blank or show an unsigned raw change. Restricting the delta columns to metrics with a known direction would also work and is simpler.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/result.py
```diff
@@ -524,7 +607,195 @@ def _format_table(headers, rows) -> str:
...
+        pd = _require_pandas()
+        columns = ["unit", *self.names, *(f"Δ% {name}" for name in self.names[1:])]
+        metric_names = self._metric_names()
+        data = []
+        for metric_name in metric_names:
+            values = [self._value(r, metric_name) for r in self.results]
+            baseline = values[0]
+            record = {"unit": self._unit_for(metric_name)}
+            for name, value in zip(self.names, values):
+                record[name] = value
```
Lokiiiiii, line 758: **F9 (minor): keying the record dict on caller-supplied strings means duplicate or reserved `names` silently corrupt the frame.**

`compare_benchmarks` validates `len(names)` but not uniqueness, and not collision with the literal `"unit"` column. Two verified cases:

```
names=["a", "a"]
  printed table:  request_throughput  10   20   +100.0%     <- correct
  to_dataframe():                   20.0 20.0    100.0      <- run1's 10.0 gone
  columns: ['unit', 'a', 'a', 'Δ% a']

names=["unit", "b"]
  to_dataframe():  unit=10.0  unit=10.0  b=20.0             <- "req/s" gone
```

`__str__` builds its row positionally so it stays correct, which means the printed table and the frame disagree — the drift the commit message says the shared helpers prevent. The frame is also internally inconsistent in the first case: two identical values with a +100% delta between them. `names=["run1","run1"]` or names derived from repeating endpoint names are easy accidents.

Suggested direction: reject duplicate names and names colliding with `"unit"` in `compare_benchmarks`, alongside the existing length check. Building the frame positionally rather than through an intermediate dict would remove the failure mode entirely.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/_recommendation_view.py
```diff
@@ -291,48 +315,102 @@ def _repr_pretty_(self, p, cycle):
...
+            dc = getattr(view.raw, "deployment_configuration", None)
+            ep = view.expected_performance
+            records.append(
+                {
+                    "idx": view._index,
+                    "spec_name": view.recommendation_spec_name,
+                    "instance_type": getattr(dc, "instance_type", None) if dc else None,
+                    "instances": getattr(dc, "instance_count", None) if dc else None,
+                    "copies/inst": (getattr(dc, "copy_count_per_instance", None) if dc else None),
+                    "container": _short_container_tag(_safe_str(dc, "image_uri")),
```
Lokiiiiii: **F10 (minor): this is the one column that carries a display sentinel into the DataFrame.**

`_safe_str` maps `None`/`""` to `"-"`, and `_short_container_tag` passes `"-"` straight through, so the record gets the string `"-"` where every sibling column keeps `None`. That contradicts this method's own docstring two lines up ("numbers stay numbers, `None` stays `None`"). Verified with `image_uri=None`:

```
container repr    : ['-']       container isna    : [False]
spec_name repr    : [None]      instance_type isna: [True]
```

Impact is on the frame, not the table: `df[df.container.notna()]`, `groupby("container")` and `value_counts()` all treat "no container" as a container named `-`.

Suggested direction: pass the raw value — `_short_container_tag(getattr(dc, "image_uri", None) if dc else None)` — and let `__str__`'s existing `"-" if value in (None, "")` branch handle display. That branch already exists and already covers this.

sagemaker-serve/tests/unit/test_ai_inference_recommender/test_listing.py
```diff
@@ -0,0 +1,162 @@
...
+)
+from sagemaker.serve.ai_inference_recommender import (
+    list_benchmarks,
+    list_recommendations,
+)
+from sagemaker.serve.ai_inference_recommender.jobs import BenchmarkJob, RecommendationJob
+
+
+# Real BenchmarkJob / RecommendationJob instances (not SimpleNamespace): list_*
+# reassigns each returned job's __class__ to the subclass — a no-op on a real
```
Lokiiiiii: **F11 (minor): this comment is describing why the test cannot exercise the line it is about.**

Because every stand-in is constructed as `BenchmarkJob`/`RecommendationJob` already, `job.__class__ = subclass` in `_collect` is a genuine no-op in all 11 listing tests — deleting that line keeps the suite green. Yet it is the reason `_collect` takes a `subclass` parameter at all, and the reason the `Returns:` docstrings can promise "each with `show_result`".

The behaviour does work: feeding base `AIBenchmarkJob` instances through `list_benchmarks` does yield `BenchmarkJob` objects with `show_result`. Nothing guards it.

Suggested direction: build the stand-ins as base `AIBenchmarkJob` / `AIRecommendationJob` and assert `isinstance(out[0], BenchmarkJob)`. That gives the retype its first real assertion and matches what `get_all` actually yields, since it hardcodes `resource_cls=AIBenchmarkJob` rather than `cls`.

While you're here: the autouse `refresh` stub on lines 45-51 uses a `lambda`. A `MagicMock` would cost nothing and would make the Describe call count assertable, which is what F3 needs.

sagemaker-serve/tests/integ/test_ai_inference_recommender_enhancements_integration.py
```diff
@@ -0,0 +1,296 @@
...
+        tokenizer=WORKLOAD_TOKENIZER,
+        concurrency=1,
+        request_count=10,
+        prompt_input_tokens_mean=32,
+        output_tokens_mean=32,
+        streaming=True,
+    )
+
+
+def test_list_benchmarks_and_recommendations_plumbing():
```
Lokiiiiii, line 63: **F12 (minor): this is the only integ test that runs on PR checks, and it triggers the F3 fan-out against a live account.**

The `slow_test` / `gpu_intensive` markers on line 89-90 belong to the next function, so this one collects into any integ run. `sagemaker-serve/tox.ini:66` says `gpu_intensive` "runs on scheduled CI, not PR checks" — so the e2e test that actually covers deploy-from-row and `compare_benchmarks` is excluded, while this one is not.

Lines 78-84 then call `list_benchmarks(endpoint="no-such-endpoint-<uuid>")` and `list_recommendations(model="s3://no-such-bucket-<uuid>/model/")`. Both are guaranteed zero-match by construction, which per F3 means a full-account Describe sweep of both job types — doubled by F2. Under `pytest -n auto` that is a reliable way to throttle the whole suite for everyone sharing the account.

The intent is good and the test is cheap in principle; it is the interaction with F3 that makes it expensive. Fixing F3 largely fixes this too.

Separately, the assertion style in the e2e test is worth a look: `for job in found: assert ...` passes vacuously when `found` is empty, so a filter regressing to always-False would not be caught. Asserting `found` is non-empty first — or accepting the raciness and asserting on a job you created — would make it load-bearing.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,207 @@
...
+    for job in iterator:
+        job.__class__ = subclass
+        if predicate is None:
+            matches.append(job)
+        else:
+            # The list summary lacks the nested field; hydrate before matching.
+            try:
+                job.refresh()
+            except Exception as exc:  # pragma: no cover - best-effort hydration
+                logger.debug("Skipping %s; could not describe it: %s", job, exc)
```
Lokiiiiii: **F13 (minor): a permissions or throttling failure silently shortens the result list, at DEBUG.**

`except Exception` plus `logger.debug` plus `continue` means a job the caller cannot describe is dropped with no visible trace at default log level. If the role lacks `sagemaker:DescribeAIBenchmarkJob`, or the API throttles under the F3 fan-out, *every* candidate is skipped and `list_benchmarks(endpoint=...)` returns `[]`. The customer concludes no benchmark ever targeted their endpoint; the on-call has nothing to correlate. Under throttling it presents as "the filter intermittently finds nothing".

This is the same hazard the `max_results` log on line 105 was added to prevent — the docstring says it "logs when the scan is truncated so a silent cap never reads as 'all matches'" — and the two cases deserve the same treatment.

Suggested direction: `logger.warning`, and count the skips so the caller can be told "N jobs were skipped because they could not be described". Throttling arguably should not be swallowed at all, since retrying is the right response rather than silently returning a short list.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/result.py
```diff
@@ -96,6 +101,23 @@ def _repr_pretty_(self, p, cycle):
...
+        ``avg``/``min``/``max``/``p50``/``p90``/``p95``/``p99``/``stddev``).
+        Rows are ordered exactly as the printed table: non-HTTP metrics
+        alphabetically, then ``http_*`` transport metrics last. Unlike the text
+        table, the frame keeps ``min``/``max``/``p95``/``stddev``.
+
+        Requires pandas (an optional dependency); raises ``ImportError`` with
+        install guidance if it is missing.
+        """
+        pd = _require_pandas()
+        ordered_names = [n for n in sorted(self.all_metrics) if not n.startswith("http_")]
```
Lokiiiiii: **F15 (minor): this re-derives the ordering rule that `__str__` already implements, so "never drift" is not enforced here.**

`__str__` (lines 91-94) buckets on `name.startswith("http_")` in a single pass; these two comprehensions derive the same partition independently. They agree today — verified `alpha, zeta, http_a, http_x` from both — but by coincidence rather than by construction, and the docstring below asserts "Rows are ordered exactly as the printed table".

`BenchmarkResult` does this properly 160 lines down: `_ordered_metric_pairs()` is called by both. Same gap in `_recommendation_view.py`, where `_perf_records()` was added as "the rows of the printed table" but `__str__` still builds `perf_rows` inline — and there the values already differ (`None`/`""` in the frame vs `"-"` in the table).

So the commit message's claim that the shared helpers keep table and frame in step holds for 2 of the 5 surfaces. The `"http_"` literal is now a live predicate in four places (91, 117, 118, 292).

Suggested direction: give `BenchmarkMetrics` an `_ordered_metric_pairs()` equivalent and have both callers use it, and refactor `_RecommendationView.__str__` onto `_perf_records()`. A single parametrised test comparing `__str__` row order against `to_dataframe().index` across all five surfaces would turn the invariant from convention into something enforced.

Lokiiiiii (changes requested):

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -5519,6 +5524,16 @@ def deploy(
...
+        # A recommendation row (mb.recommendations.best or [i]) may be passed
+        # directly. Resolve it to its positional index, which names the exact
+        # row even when spec names repeat across rows.
+        if recommendation is not None:
+            if recommendation_spec_name is not None or recommendation_index:
+                raise ValueError(
+                    "Pass only one of `recommendation`, `recommendation_spec_name`, "
+                    "or `recommendation_index` to deploy()."
+                )
+            recommendation_index = getattr(recommendation, "_index", 0)
```
Lokiiiiii: **F16 (major): the index is a proxy for the row, and it goes stale — so this can still deploy a different row than the one passed.**

This correctly fixes the duplicate-spec-name case I raised (verified: two rows sharing `"dup"`, passing `mb.recommendations[1]` now provisions row 1's ModelPackage and `ml.p4d.24xlarge` x4). But the row object is discarded here and only an integer survives, and that integer is minted from a different list than the one it indexes into:

- minted at `model_builder.py:5087` — `rows = list(job.recommendations or [])` snapshots the rows when `.recommendations` is accessed
- read at `model_builder.py:5177-5180` — `rows` is re-read at deploy time, and `:5179` may call `refresh()`, replacing the list wholesale from a later service response

Nothing ties the two together. A verifier asked to disprove the fix reproduced five silent misroutes:

| input | deploys |
|---|---|
| row from a **different** `ModelBuilder` | that builder's row N — unrelated ModelPackage and instance type |
| stale row across a `refresh()` that **reordered** rows | wrong row |
| stale `_index` after a re-run returning a different same-length list | wrong row |
| captured `.best` after the job re-ran and row 0 changed | wrong row |
| `row.raw` (a **public** property), or a string / int / dict | row 0 |

The range guard at `:5218` only fires when the list *shrinks* past the index; whenever the new list is long enough, a stale or foreign index is accepted with no error and no warning. Picking a different instance type on a `p4d`-class row is a real cost and capacity event, and the user has no signal it happened.

Suggested direction — either:

1. keep the index but verify it: after `rec = rows[recommendation_index]` (`:5224`) assert `rec is recommendation.raw` and raise otherwise. Two lines, and it closes every row in the table above including F6; or
2. stop using a proxy: pass the row's `raw` shape straight through and skip the `rows` lookup when a row object was supplied, which removes the two-reads coupling entirely.

Either way `getattr(..., "_index", 0)`'s `0` default should go — silently defaulting a mistyped argument to the top-ranked row is the root of the last table row.

On the test: `test_duplicate_spec_deploys_the_exact_row_passed` mocks `_deploy_recommendation` and asserts on its kwargs, so it proves the index is forwarded but not that the index selects the intended row. `TestDeployRecommendationRowSelection` already exercises selection for real against `describe_model_package` — the duplicate-spec case would be worth adding there so the fix is actually pinned.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,249 @@
...
+        # next() drives the iterator's per-object Describe, so a failure for one
+        # candidate is caught here rather than aborting the listing.
+        try:
+            job = next(it)
+        except StopIteration:
+            break
+        except Exception as exc:  # noqa: BLE001 - per-job hydration is best-effort
+            skipped += 1
+            scanned += 1
+            logger.warning("Skipping a %s that could not be described: %s", subclass.__name__, exc)
```
Lokiiiiii: **F17 (nit): the F13 fix overshot — this emits one WARNING per undescribable candidate, up to `max_scan`.**

Raising this from DEBUG to WARNING was the right move, and the summary line at `:126` carries the operator-facing information. But this per-candidate line fires on every skip, so a role that cannot `DescribeAIBenchmarkJob` at all produces up to `DEFAULT_MAX_SCAN` (1000) warnings for a single `list_benchmarks()` call. Measured with 250 denied candidates: 250 per-candidate warnings plus the one summary.

Suggested direction: log the first skip (or the first few) at WARNING for diagnosability and leave the rest to the summary, or log per-candidate at DEBUG and keep WARNING for the summary only. The summary already reports the count.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/listing.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def _collect(iterator, predicate, max_results: int, max_scan: int, subclass) -> list:
+    """Keep candidates from ``iterator`` matching ``predicate``, re-typed to
+    ``subclass`` (so ``show_result`` is available).
+
+    The iterator hydrates each object as it yields it, so this loop does not
+    Describe again. ``max_results`` bounds matches returned; ``max_scan`` bounds
+    candidates examined. A candidate that fails to Describe is skipped, not
+    fatal. All three limits are logged when hit.
```
Lokiiiiii, line 98: **F18 (nit): "All three limits are logged when hit" — `max_results` isn't.**

Measured against the three cases: hitting `max_scan` warns (`:120`), skipped candidates warn (`:126`), and hitting `max_results` (`:118`) logs nothing. Natural exhaustion correctly stays silent — the `while/else` is right, I checked that specifically.

Revision 1 did log this one, and its docstring gave the reason: "logs when the scan is truncated so a silent cap never reads as 'all matches'." That rationale still applies to `max_results` — a caller who gets exactly 100 rows back has no way to tell whether that is all the matches or just the cap. The log went away with the rewrite.

Suggested direction: either restore a log on the `max_results` break, or narrow the docstring claim to the two limits that are logged.

Lokiiiiii (changes requested):

sagemaker-serve/src/sagemaker/serve/model_builder.py
Lokiiiiii: This still needs to be addressed.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/_recommendation_view.py
Lokiiiiii: This still needs to be addressed.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/result.py
Lokiiiiii: This still needs to be addressed.

sagemaker-serve/src/sagemaker/serve/ai_inference_recommender/result.py
Lokiiiiii: This still needs to be addressed.

Lokiiiiii (approved):

zhaoqizqwang (approved):

PR #6154, opened 2026-08-04
fix(train): set input_data_config to None

danb27:
*Issue #, if available:* N/A - small fix, can create if required. 

*Description of changes:* When no input channels are provided, input_data_config must be None rather than an empty list to avoid SageMaker API validation errors. This appears to be a regression caused by this [recent fix](https://github.com/aws/sagemaker-python-sdk/commit/4171658b738e622daf60197a3506db5d7726f29a). Before, `input_data_config=[]` would be dropped from the config silently, avoiding the validation issue in the API. Now, however, it is not dropped, which breaks the validation. 


danb27:
@papriwal sorry to tag you directly, but I saw are active on this repo. I was hoping to get this PR looked at. It has been open for some time. Please let me know if I missed any required process.

PR #6159, opened 2026-08-05
feat(telemetry): Add tracking for dry_run, notifications, reuse

zhaoqizqwang:
feat(telemetry): Add tracking for dry_run, notifications, reuse


sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -627,6 +637,14 @@ def list_notification_rules(
             event_bus_arn=event_bus_arn,
         )
 
+    @_telemetry_emitter(
+        feature=Feature.MODEL_CUSTOMIZATION,
+        func_name="BaseTrainer.stream_logs",
+        telemetry_params=[
+            ("_model_name", TelemetryParamType.ATTR_VALUE),
```
mujtaba1747: Do we have approval to collect this in telemetry?

zhaoqizqwang: Yes we are reusing the decorator `_telemetry_emitter`. It has been used in other places

mujtaba1747:
Few more spots where we can add telemetry:
evaluators dry run
stream logs / show metrics for evaluators (not sure if we implemented this for Evaluators)
model reuse for bedrock modelbuilder
Ideally if the model reuse telemetry can also show if model is_nova or not, that would be great, but it's a nice to have.

sagemaker-serve/src/sagemaker/serve/bedrock_model_builder.py
```diff
@@ -258,7 +258,9 @@ def _is_nova_model_for_telemetry(self) -> bool:
         func_name="BedrockModelBuilder.deploy",
         telemetry_params=[
             ("model_package", TelemetryParamType.ATTR_EXISTS),
+            ("_is_nova_model_for_telemetry", TelemetryParamType.ATTR_CALL),
```
mujtaba1747, line 261: how does this work? What if bedrock model builder doesn't call _is_nova_model_for_telemetry() and calls _is_nova_model instead?

zhaoqizqwang: so every time BedrockModelBuilder.deploy() is called, the decorator runs `_is_nova_model_for_telemetry` before the actual deploy() body executes. The actual method called inside deploy doesn't matter

mujtaba1747: Thanks for explaining.

PR #6162, opened 2026-08-06
fix: Add pre-validation for Bedrock and MLflow role permissions

rsareddy0329:
Add pre-validation for Bedrock and MLflow role permissions

Validate execution role permissions before submitting LLM-as-Judge evaluation jobs to fail fast with actionable guidance instead of failing during execution.

ref doc used: https://docs.aws.amazon.com/sagemaker/latest/dg/model-customize-open-weight-prereq.html

*Issue #, if available:*

*Description of changes:*


rsareddy0329:
> **Claude finished @rsareddy0329's task in 1m 12s** —— [View job](https://github.com/aws/sagemaker-python-sdk/actions/runs/31670257594)
> 
> ### Code Review: Pre-validation for Bedrock/MLflow role permissions
> * [x]  Read the PR diff
> * [x]  Inspect `iam_role_resolver` / `iam_policies` integration
> * [x]  Verify per-backend role-type scoping & trust design
> * [x]  Check tests & CodeQL findings
> 
> This revision is in good shape. The blocker that recurred through the earlier rounds — unconditional Bedrock gating on **all** evaluators — is resolved, and I confirmed the following against the checked-out source:
> 
> * ✅ **Per-backend scoping is correct.** `_get_aws_execution_context()` takes `role_type` defaulting to `"training"` (`base_evaluator.py:741`), and only `LLMAsJudgeEvaluator.evaluate` passes `"model_eval"` (`llm_as_judge_evaluator.py`). Benchmark/CustomScorer/InspectAI/MultiTurnRL keep the `"training"` default, so they are not gated on Bedrock permissions they don't use.
> * ✅ **Trust policy** for `model_eval` requires only `sagemaker.amazonaws.com` — correct for the SMTJ/serverless path where the container calls Bedrock as the execution role (Bedrock never assumes the role). Verified `_expected_trust_services` (`iam_role_resolver.py:392`) returns `{"sagemaker.amazonaws.com"}`.
> * ✅ **No resource-scoped false denials.** `_get_smoke_test_actions` (`iam_role_resolver.py:115`) gates only on `Resource: "*"` statements, so the MLflow (`mlflow-app/*`), SageMaker (`training-job/*`), S3, and KMS statements are correctly excluded. The latest commit scoping `bedrock:InvokeModel*` to `foundation-model/*` means only `bedrock:CreateEvaluationJob`/`GetEvaluationJob` are gated — locked in by `test_evaluation_smoke_actions_exclude_bedrock_invoke`.
> * ✅ **Existing tests updated** and the new `test_get_aws_execution_context_role_type_override` locks in the `"model_eval"` forwarding.
> 
> ### 🟡 Medium (non-blocking) — confirm `bedrock:CreateEvaluationJob`/`GetEvaluationJob` are actually needed
> These two are the only gated Bedrock actions. Per the design comment in `base_evaluator.py`, the eval container runs **as** the execution role and calls Bedrock with the role's own credentials — that's `bedrock:InvokeModel*` (now correctly _un_-gated). `CreateEvaluationJob`/`GetEvaluationJob` are the _native Bedrock evaluation-job_ orchestration APIs. If the SMTJ/serverless LLM-as-Judge backend does **not** call those directly (i.e. it drives evaluation itself rather than delegating to a Bedrock-managed job), a role that legitimately has `InvokeModel` but not the job-orchestration actions would be hard-failed for callers who can self-simulate. Please confirm they're genuinely required on this path; if not, drop them (or move them to a non-gated, resource-scoped statement).
> 
> ### 🟢 Minor (non-blocking)
> * **Naming drift:** the config key/role type is `model_eval`, but the new test file is `test_bedrock_role_validation.py`, its class is `TestEvaluationRoleType`, and several docstrings still say _"the 'evaluation' role type"_ (`test_bedrock_role_validation.py:1,12,15`). The tests pass (they assert `"model_eval"`), but aligning on one term would help future readers.
> * **Regression-test gap:** no test asserts a _non_-LLM-as-Judge evaluator resolves with `role_type="training"`. A one-line test (e.g. Benchmark's `_get_aws_execution_context()` calls `resolve_and_validate_role` with `role_type="training"`) would explicitly lock in the exact regression this PR guards against.
> * **CodeQL** ("incomplete URL substring sanitization" on `test_bedrock_role_validation.py`): `_expected_trust_services` returns a `Set[str]`, so these are set-membership checks and semantically fine — a false positive. `test_evaluation_trust_includes_sagemaker` already uses the tighter `assert expected == {...}` form; applying the same equality style to the other assertions would silence the scanner.
> 
> **Summary:** the per-backend role-type scoping, trust-policy design, smoke-test gating, and test updates are all correct. Nothing blocking. The one item worth a definitive answer before merge is whether `bedrock:CreateEvaluationJob`/`GetEvaluationJob` are truly exercised by this backend; the rest are cosmetic. LGTM otherwise.

Are CreateEvaluationJob/GetEvaluationJob needed?
  
Yes -- they're the specific errors that APIs fail at,  From the error tracker:
  
```
"Failed to create Bedrock evaluation job: An error occurred (ValidationException) when calling the CreateEvaluationJob operation: Your account is not authorized to invoke this API operation"
```
  
The SMTJ/serverless LLM-as-Judge requires CreateEvaluationJob — it delegates the actual judge inference to a Bedrock-managed evaluation job rather than calling InvokeModel directly. That's why customers without this permission fail 30 minutes in. This is the core customer pain point the PR solves.

jam-jee:
verified fortress reports are low sev (explicit usage of aws:arn**)

jam-jee (approved):

PR #6175, opened 2026-08-10
fix(serve): pre-deploy JumpStart benchmark data + public HuggingFace download helper

ZealSV:
Fixes two issues with ModelBuilder benchmark APIs for pre-deploy JumpStart models, plus adds a public HuggingFace-Hub download helper:

- benchmark_metrics: display_benchmark_metrics() read self.benchmark_metrics, a property that existed in v2 but was dropped in the v3 port, so it raised AttributeError. Restore the property (a pandas DataFrame built from the model's published deployment-config benchmark data).

- list_deployment_configs(): with no instance_type it called _get_deployment_configs() without first loading the JumpStart metadata configs, so a pre-deploy builder (_metadata_configs=None) got []. Added _ensure_metadata_configs() into _get_deployment_configs so every caller (listing and benchmark data) lazily loads them.

- download_huggingface_model(): new public helper in serve.utils.hf_utils, exported from sagemaker.serve. Downloads a Hub snapshot and optionally uploads it to S3, so notebooks no longer hand-roll snapshot_download + S3Uploader. The previous downloader was a method on a private class and not importable.

Unit: 11 new tests (6 benchmark/config, 5 HF helper).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sagemaker-core/src/sagemaker/core/jumpstart/utils.py
+++ b/sagemaker-core/src/sagemaker/core/jumpstart/utils.py
@@ -1442,8 +1442,19 @@ def get_metrics_from_deployment_configs(
     if not deployment_configs:
         return {}
 
-    data = {"Instance Type": [], "Config Name": [], "Concurrent Users": []}
-    instance_rate_data = {}
+    # Build per-row records, then pivot to column-oriented at the end, padding
+    # columns a row didn't set with None. Keeps every column the same length and
+    # each value on its own row; appending columns independently would crash
+    # pd.DataFrame on ragged lengths and misalign the sparse (pricing) column.
+    rows: List[Dict[str, Any]] = []
+    column_order: List[str] = ["Instance Type", "Config Name", "Concurrent Users"]
+    seen_columns = set(column_order)
+
+    def _register_column(name: str) -> None:
+        if name not in seen_columns:
+            seen_columns.add(name)
+            column_order.append(name)
+
     for index, deployment_config in enumerate(deployment_configs):
         benchmark_metrics = deployment_config.benchmark_metrics
         if not deployment_config.deployment_args or not benchmark_metrics:
@@ -1465,25 +1476,31 @@ def get_metrics_from_deployment_configs(
                     else current_instance_type
                 )
 
-                data["Config Name"].append(deployment_config.deployment_config_name)
-                data["Instance Type"].append(instance_type_to_display)
-                data["Concurrent Users"].append(concurrent_user)
+                row: Dict[str, Any] = {
+                    "Instance Type": instance_type_to_display,
+                    "Config Name": deployment_config.deployment_config_name,
+                    "Concurrent Users": concurrent_user,
+                }
+
+                for metric in metrics:
+                    column_name = _normalize_benchmark_metric_column_name(metric.name, metric.unit)
+                    _register_column(column_name)
+                    row[column_name] = metric.value
 
                 if instance_type_rate:
                     instance_rate_column_name = (
                         f"{instance_type_rate.name} ({instance_type_rate.unit})"
                     )
-                    instance_rate_data[instance_rate_column_name] = instance_rate_data.get(
-                        instance_rate_column_name, []
-                    )
-                    instance_rate_data[instance_rate_column_name].append(instance_type_rate.value)
+                    _register_column(instance_rate_column_name)
+                    row[instance_rate_column_name] = instance_type_rate.value
 
-                for metric in metrics:
-                    column_name = _normalize_benchmark_metric_column_name(metric.name, metric.unit)
-                    data[column_name] = data.get(column_name, [])
-                    data[column_name].append(metric.value)
+                rows.append(row)
 
-    data = {**data, **instance_rate_data}
+    # Pivot to column-oriented, padding unset cells with None.
+    data: Dict[str, List[Any]] = {column: [] for column in column_order}
+    for row in rows:
+        for column in column_order:
+            data[column].append(row.get(column))
     return data
 
 
--- a/sagemaker-serve/src/sagemaker/serve/__init__.py
+++ b/sagemaker-serve/src/sagemaker/serve/__init__.py
@@ -41,6 +41,7 @@ from sagemaker.serve.ai_inference_recommender import (
     WorkloadValidationError,
     start_benchmark,
 )
+from sagemaker.serve.utils.hf_utils import download_huggingface_model
 
 __all__ = [
     "InferenceSpec",
@@ -56,4 +57,5 @@ __all__ = [
     "FeatureGatedError",
     "WorkloadValidationError",
     "start_benchmark",
+    "download_huggingface_model",
 ]
--- a/sagemaker-serve/src/sagemaker/serve/model_builder.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder.py
@@ -4782,6 +4782,20 @@ class ModelBuilder(_InferenceRecommenderMixin, _ModelBuilderServers, _ModelBuild
             sagemaker_session=self.sagemaker_session,
         )
 
+    @property
+    def benchmark_metrics(self):
+        """Benchmark metrics for the model's JumpStart deployment configs.
+
+        Returns a pandas ``DataFrame`` (one row per config/instance) built from
+        the model's published benchmark data. Available for JumpStart models
+        (or HuggingFace models with a JumpStart equivalent) before deploy.
+        """
+        import pandas as pd
+
+        df = pd.DataFrame(self._get_deployment_configs_benchmarks_data())
+        df.index = [""] * len(df)
+        return df
+
     @_telemetry_emitter(
         feature=Feature.MODEL_CUSTOMIZATION, func_name="model_builder.display_benchmark_metrics"
     )
--- a/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
+++ b/sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
@@ -2709,6 +2709,11 @@ class _ModelBuilderUtils:
             selected_config_name (Optional[str]): The name of the selected deployment config.
             selected_instance_type (Optional[str]): The selected instance type.
         """
+        # Lazily load the JumpStart metadata configs. Without this a pre-deploy
+        # builder (model set, build()/deploy() not yet called) has
+        # _metadata_configs=None, so both list_deployment_configs() and the
+        # benchmark-metrics data would come back empty.
+        self._ensure_metadata_configs()
         deployment_configs = []
         if not self._metadata_configs:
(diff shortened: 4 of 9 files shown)


sagemaker-serve/src/sagemaker/serve/utils/hf_utils.py
```diff
@@ -14,14 +14,82 @@
...
+
+    try:
+        from huggingface_hub import snapshot_download
+    except ImportError as exc:
+        raise ImportError(
+            "download_huggingface_model requires huggingface_hub, which is not "
+            "installed. Install it with `pip install huggingface_hub`."
+        ) from exc
+
+    download_dir = local_dir or tempfile.mkdtemp(prefix="hf-model-")
```
Lokiiiiii: **major** — When `s3_uri` is given without `local_dir`, this temporary directory is never removed.

The function returns the S3 URI at the end, so the caller never receives `download_dir` and has no handle to clean it up — the path only appears in the INFO log on the next line. `S3Uploader.upload` doesn't delete its source either: it delegates to `Session.upload_data`, which walks the tree and calls `upload_file` per object. I also searched the repo for `shutil.rmtree`, `TemporaryDirectory`, `atexit`, and `weakref.finalize` and found nothing that could reclaim a `hf-model-*` directory.

So every call in this mode leaves a full model snapshot on local disk — tens to hundreds of GB for an LLM, which will exhaust the volume in a Studio or notebook environment. The docstring describes this directory as a staging area for the upload, which is exactly the right intent; it just never gets cleaned up.

Worth noting that no test reaches this branch: all five cases in `TestDownloadHuggingfaceModel` either omit `s3_uri` or pass `local_dir` alongside it, so `mkdtemp` never executes.

Suggestion: wrap the download and upload in `tempfile.TemporaryDirectory()`, or `shutil.rmtree(download_dir)` in a `finally` guarded on `local_dir is None` so a caller-supplied directory is left alone. `serve/utils/uploader.py` cleans up its own temp payload, for precedent.

sagemaker-core/src/sagemaker/core/jumpstart/utils.py
```diff
@@ -1465,25 +1476,31 @@ def get_metrics_from_deployment_configs(
...
 
                 if instance_type_rate:
                     instance_rate_column_name = (
                         f"{instance_type_rate.name} ({instance_type_rate.unit})"
                     )
-                    instance_rate_data[instance_rate_column_name] = instance_rate_data.get(
-                        instance_rate_column_name, []
-                    )
-                    instance_rate_data[instance_rate_column_name].append(instance_type_rate.value)
+                    _register_column(instance_rate_column_name)
```
Lokiiiiii, line 1494: **minor** — Registering the pricing column here, before the metric loop just below, changes the resulting column order. Previously the rate column was merged in last (`data = {**data, **instance_rate_data}`), so it landed at the end.

I ran the same non-ragged input through both versions; lengths are identical, so ordering is the only difference:

- before: `['Instance Type', 'Config Name', 'Concurrent Users', 'Latency, TTFT (P50 in ms)', 'Instance Rate (USD/Hr)']`
- after: `['Instance Type', 'Config Name', 'Concurrent Users', 'Instance Rate (USD/Hr)', 'Latency, TTFT (P50 in ms)']`

That's user-visible in `display_benchmark_metrics()` output, and it isn't mentioned in the PR description, which reads as a pure bug fix. No test asserts column order, so nothing pins it either way.

If it's unintended, registering the metric columns before the rate column restores the previous layout. If it's intended, it's worth a line in the description.

sagemaker-serve/src/sagemaker/serve/__init__.py
```diff
@@ -41,6 +41,7 @@
     WorkloadValidationError,
     start_benchmark,
 )
+from sagemaker.serve.utils.hf_utils import download_huggingface_model
```
Lokiiiiii, line 44: **minor** — This adds a new public export, but `download_huggingface_model` doesn't appear in `migration.md`, `docs/`, `llms.txt`, or `v3-examples/`. `AGENTS.md` asks for this explicitly:

> **Public API changes:** update `migration.md` and any affected docstrings/examples so v3 guidance stays consistent across the repo.

The restored `ModelBuilder.benchmark_metrics` property is also absent from `migration.md`, even though a v2 property returning in v3 is squarely a migration note.

This matters more than usual here: the helper's stated purpose is that notebooks stop hand-rolling `snapshot_download` + `S3Uploader`, and that depends entirely on someone being able to find it.

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -4782,6 +4782,29 @@ def transformer(
...
+        the model's published benchmark data. Available for JumpStart models
+        (or HuggingFace models with a JumpStart equivalent) before deploy.
+        """
+        import pandas as pd
+
+        data = self._get_deployment_configs_benchmarks_data() or {}
+        # Pad short columns so a ragged source renders instead of raising
+        # "All arrays must be of the same length".
+        if data:
+            max_len = max(len(values) for values in data.values())
```
Lokiiiiii: **minor** — This padding looks redundant now. `_get_deployment_configs_benchmarks_data()` calls `get_metrics_from_deployment_configs()`, which after the `sagemaker-core` change in this PR always returns equal-length columns — so the ragged input this guards against can no longer come from the production path.

Two consequences: the same invariant is now enforced in two places and will drift, and `test_benchmark_metrics_property_pads_ragged_columns` constructs an input the pipeline can't actually produce.

Either drop the padding and rely on the core fix, or keep it and note in a comment that it's defensive for callers passing a raw dict — so the next reader knows which layer is authoritative.

sagemaker-serve/src/sagemaker/serve/model_builder_utils.py
```diff
@@ -2709,6 +2709,11 @@ def _get_deployment_configs(
             selected_config_name (Optional[str]): The name of the selected deployment config.
             selected_instance_type (Optional[str]): The selected instance type.
         """
+        # Lazily load the JumpStart metadata configs. Without this a pre-deploy
+        # builder (model set, build()/deploy() not yet called) has
+        # _metadata_configs=None, so both list_deployment_configs() and the
+        # benchmark-metrics data would come back empty.
+        self._ensure_metadata_configs()
```
Lokiiiiii, line 2716: **minor** — The description says this `_ensure_metadata_configs()` call was *moved* into `_get_deployment_configs`, but it's an addition (+5/−0 in this file) and the previous call sites all remain: `model_builder_utils.py:2298`, `model_builder.py:4904`, and `model_builder.py:5056`.

The net effect is what you describe, and it's harmless because the method is idempotent (it early-returns unless `_metadata_configs is None`). But `model_builder.py:5050` now calls `_get_deployment_configs(...)` and then `:5056` calls `_ensure_metadata_configs()`, which is a guaranteed no-op — easy to drop while it's fresh, and "also call" would be more accurate than "move" in the description.

Separately, the description says 9 new tests; I count 11 — 1 in `test_jumpstart_utils.py`, 5 in `TestPreDeployBenchmarkData`, 5 in `TestDownloadHuggingfaceModel`.

sagemaker-serve/tests/unit/utils/test_hf_utils.py
```diff
@@ -216,5 +220,58 @@ def _urlopen_side_effect(request):
...
+    def test_requires_local_dir_or_s3_uri(self):
+        """Neither destination given is a caller error, before any download."""
+        with self.assertRaises(ValueError) as context:
+            download_huggingface_model("gpt2")
+        self.assertIn("local_dir, s3_uri", str(context.exception))
+        self._hf_stub.snapshot_download.assert_not_called()
+
+    def test_downloads_to_local_dir_and_returns_it(self):
+        """With only local_dir, it downloads there and returns the path."""
+        result = download_huggingface_model("gpt2", local_dir="/tmp/gpt2")
```
Lokiiiiii: **minor** — These tests pass a hard-coded `/tmp/gpt2`, and the function under test calls the real `os.makedirs(download_dir, exist_ok=True)`, so running the suite creates that directory on the host. I confirmed it: `/tmp/gpt2` was absent before the run and present afterwards.

Same pattern at :254, :264, and :272. On a shared host, or where `/tmp/gpt2` already exists owned by another user, this can collide or fail for reasons unrelated to the code under test.

Using `tempfile.mkdtemp()` / `TemporaryDirectory` in the test, or patching `os.makedirs`, keeps the suite side-effect free.

sagemaker-serve/src/sagemaker/serve/utils/hf_utils.py
```diff
@@ -14,14 +14,82 @@
...
+    os.makedirs(download_dir, exist_ok=True)
+    logger.info("Downloading model %s from Hugging Face Hub to %s", model_id, download_dir)
+    snapshot_download(repo_id=model_id, local_dir=download_dir, token=hf_hub_token)
```
Lokiiiiii: **nit** — For a helper aimed at notebooks, two `snapshot_download` passthroughs would add a lot of value:

- `revision` — without it every download tracks the latest `main`, so results aren't reproducible and the weights can change under a pinned SDK version.
- `allow_patterns` / `ignore_patterns` — repos commonly ship both `.bin` and `.safetensors` copies of the weights, so the default pulls roughly twice the bytes and twice the disk. That compounds the temp-directory cleanup issue at :75.

sagemaker-serve/tests/unit/utils/test_hf_utils.py
```diff
@@ -216,5 +220,58 @@ def _urlopen_side_effect(request):
...
+        download_huggingface_model("gpt2", local_dir="/tmp/gpt2", hf_hub_token="hf_tok")
+        self.assertEqual(self._hf_stub.snapshot_download.call_args.kwargs["token"], "hf_tok")
+
+    def test_missing_huggingface_hub_raises_import_error(self):
+        """A clear ImportError is raised when huggingface_hub is unavailable."""
+        with patch.dict(sys.modules, {"huggingface_hub": None}), self.assertRaises(
+            ImportError
+        ) as context:
+            download_huggingface_model("gpt2", local_dir="/tmp/gpt2")
+        self.assertIn("huggingface_hub", str(context.exception))
```
Lokiiiiii: **nit** — This assertion can't distinguish the helper's message from Python's own. Setting `sys.modules["huggingface_hub"] = None` makes the import machinery itself raise `ImportError: import of huggingface_hub halted; None in sys.modules`, which also contains `huggingface_hub` — so the test passes whether or not the custom `raise ImportError(...)` wrapper survives.

Asserting on something distinctive, such as `pip install huggingface_hub`, would pin the behaviour you actually added.

Lokiiiiii (changes requested):

Lokiiiiii (approved):

PR #6176, opened 2026-08-12
change(train): gate deep integ tests behind gpu_intensive, add shallow submit-then-stop suite

jam-jee:
## What this changes

Adds a fast server-side acceptance suite for `sagemaker-train` to the PR gate, and marks the deep tests that submit a job and wait for it so the existing gate stops running them.

**The existing `integ-tests` CodeBuild job is unchanged and still runs.** This PR adds a second job next to it rather than replacing it.

`CreateTrainingJob` returns a `TrainingJobArn` only after the request has cleared every synchronous server-side gate: public-model shape validation, SigV4, `sagemaker:CreateTrainingJob` authorization (including condition keys), `iam:PassRole` on the execution role, the training backend's synchronous request validators, its role-assuming validators (which make real S3/ECR/FSx calls as the customer), post-validator business logic (training-plan capacity, routing, recipe filtering), and the final conditional write that rejects duplicate job names.

So **"the ARN came back" proves the SDK-shaped payload was accepted as sent, and the caller held the permissions needed to submit it** — without paying for a training run. The job is stopped immediately.

## New: `tests/integ/train/shallow` (100 tests, 84 on the PR gate)

One file per trainer, mirroring the existing deep-suite layout so the shallow counterpart of any deep test is obvious:

| Shallow file | Deep counterpart |
|---|---|
| `test_model_trainer.py` | `test_model_trainer.py` |
| `test_sft_trainer.py` | `test_sft_trainer_integration.py` |
| `test_dpo_trainer.py` | `test_dpo_trainer_integration.py` |
| `test_rlvr_trainer.py` | `test_rlvr_trainer_integration.py` |
| `test_rlaif_trainer.py` | `test_rlaif_trainer_integration.py` |
| `test_cpt_trainer.py` | `test_cpt_hyperpod.py` |
| `test_multi_turn_rl_trainer.py` | `test_multi_turn_rl_trainer_integration.py` |
| `test_tuner.py` | `test_tuner_distributed.py` |
| `test_nova_data_mixing.py` | `test_sft_trainer_data_mixing_integration.py` |
| `test_nova_trainers.py` | `::test_sft_trainer_nova_workflow`, `::test_rlvr_trainer_nova_workflow`, `test_sft_trainer_serverful_smtj.py` |

`harness.py` provides `submitted()` / `assert_submitted()` / `assert_rejected()`: forces `wait=False`, resolves the submitted job across the four attribute names trainers use for it, and stops the job in a `finally` so a failed assertion still cleans up.

`recipe_cases.py` holds the cases every recipe trainer shares. Each per-trainer class subclasses `RecipeTrainerCases` and sets `TRAINER`, so adding a trainer is a two-line file and per-trainer deviations are declared rather than duplicated — `EXTRA_KWARGS` (RLAIF's reward model), `SUPPORTS_SERVERFUL = False` (RLAIF takes no `compute`), `SUPPORTS_TRAINING_TYPE = False` (CPT has no LoRA/full split).

Negative tests are included deliberately: without them the suite would stay green even if the SDK started sending a permissive-but-wrong payload.

Not every trainer creates a `TrainingJob` — `HyperparameterTuner` creates a `HyperParameterTuningJob` and `MultiTurnRLTrainer` creates an AgentRFT `Job` — so `assert_submitted` takes the expected ARN resource segment.

## Coverage of every `gpu_intensive` test

The rule: **a deep test belongs off the PR gate only if this suite covers the same code path.** There are 46 `gpu_intensive` tests in `tests/integ/train`. All are accounted for:

**Covered by this suite** — `test_model_trainer.py` (8: tar source, py/sh entry, MPI, torchrun, HP json/yaml, custom driver), SFT (complete workflow incl. MLflow, validation dataset, sequence length, Nova, serverful SMTJ), DPO (both), RLAIF (complete workflow, reward-prompt ARN, continued fine-tuning), RLVR (complete workflow, all three reward-function forms, recipe+overrides, sequence length, Nova), tuner (`sm_drivers` channel), MTRL (3, needs prerequisites), CPT HyperPod (needs a cluster), Nova data mixing.

The full test-by-test mapping is in [`tests/integ/train/shallow/README.md`](https://github.com/aws/sagemaker-python-sdk/blob/shallow-pr-checks-sagemaker-train/sagemaker-train/tests/integ/train/shallow/README.md).

MLflow is worth calling out: every `*_complete_workflow` deep test configures it, so `RecipeTrainerCases` covers both forms — experiment/run names (always runs) and `mlflow_resource_arn` (skips when the account has no app).

**Not covered, and why:**

- **11 evaluator tests** (`test_benchmark_evaluator.py`, `test_custom_scorer_evaluator.py`, `test_mtrl_evaluator_3p_agent.py`, `test_mtrl_trainer_integration.py`) — `evaluate()` is a different API surface returning pipeline executions rather than jobs, so it needs its own harness support. **Already `gpu_intensive` on master, so this PR loses no coverage there.** Clearest follow-up, and per "What bounded the `integ-tests` job" below it is also what would let the six newly marked evaluator tests be covered rather than merely deselected.
- **3 HyperPod tests** — submit to a pre-provisioned cluster, not `CreateTrainingJob`. `test_cpt_trainer.py` is written in the shallow style and activates when `SHALLOW_HYPERPOD_CLUSTER` is set.

**This PR newly marks 16 tests**, in two groups:

- **10 with shallow counterparts** — the 8 in `test_model_trainer.py`, `test_sft_trainer_lora_with_sequence_length`, and `test_tuner_includes_sm_drivers_channel`. These satisfy the rule above outright.
- **6 pipeline-waiting evaluator tests** — see "What bounded the `integ-tests` job" below. These do *not* have a shallow counterpart yet; the reasoning for marking them anyway, and what coverage that costs, is spelled out there and in the suite README rather than glossed over.

Everything else listed above was already marked on master.

Deliberately **not** marked: `test_notifications.py` (EventBridge/SNS side effects) and `test_local_model_trainer.py` (no service call, but it does run real containers).

The rule is documented in the suite README so a future change cannot silently erode the gate.

`tox.ini`: widened the `gpu_intensive` description. Despite the name it gates anything consuming real training capacity, including serverless and CPU-instance jobs.

### Fixtures that look up rather than create

`mlflow_arn`, `reward_lambda_arn` and `reward_evaluator` only look their resources up and skip when absent. The deep suite's equivalents create them — IAM roles, Lambdas, MLflow apps, registry entries — which is a durable side effect a fast PR-gate suite should not have.

## Workflow change

Two things, and the first is smaller than it sounds:

**1. `integ-tests` is untouched.** The job definition is byte-identical to master — `sagemaker-train` is still in its matrix and still runs the full CodeBuild suite. What changes is what that suite *selects*: its buildspec already filters `-m "not gpu_intensive and not us_east_1"`, so the marks this PR adds are what take the expensive tests off the gate. No workflow edit was needed for that, and no client-side coverage is lost — the CodeBuild job runs the whole `tests/integ` tree, so the ~170 tests that make no service call keep running exactly as before.

**2. New `fast-integ-tests` job**, scoped to `tests/integ/train/shallow` only, running directly on the runner:

```
python -m pytest tests/integ/train/shallow -m "not gpu_intensive and not us_east_1" -n 8 --dist loadfile
```

84 of the suite's 100 tests; the 16 deselected are the 5 `us_east_1` Nova cases (this job holds us-west-2 credentials only — they run in `integ-tests-us-east-1`) and the 11 `gpu_intensive` CPT/MTRL cases, which are written in the shallow style but need a HyperPod cluster and an agent runtime respectively.

Deliberately **not** widened to the whole `tests/integ/train` tree: the CodeBuild job already covers the client-side tests, so widening would duplicate them and double the training jobs this suite creates. Separate job rather than folded into the buildspec because the buildspec is CDK-managed outside this repo, while this selection is reviewable in the PR that changes it — and because a shallow failure then reports as its own check, distinguishable at a glance from a deep-suite failure.

## Concurrency cap (`SHALLOW_MAX_CONCURRENT_JOBS`)

Creating a job per test puts the suite against the training-job concurrency quota — specifically *"Maximum number of concurrent model customization serverless jobs per Region"* (**20** in the test account), which counts serverless jobs; the default recipe-trainer path (`compute=None`) submits serverless. Serverful jobs (`ModelTrainer`, the tuner, `test_explicit_compute_is_accepted`) are instead bounded by the per-instance-type quota, counted in instances. A slot is one concurrent job, plus one per additional instance for a serverful job, so a single cap (default **10**) is safe against both.

**The subtle part — and I got it wrong at first.** The service counts a job against the quota from `CreateTrainingJob` until it reaches `Completed`/`Failed`/`Stopped`, **not** until `StopTrainingJob` returns. Those are far apart: `stop()` returns in a few seconds, but the job takes **~1–3 min** to actually drain (measured; the reservation is torn down without ever becoming billable). My first cut released the slot at `stop()`, which bounded nothing — with cap 10 and 8 workers each slot recycled ~20× inside one job's counted lifetime, the suite peaked at **~37** concurrent jobs and tripped `ResourceLimitExceeded` at *utilization 21 against the limit of 20*.

The fix holds the slot until the job is terminal (`_wait_until_terminal`, polling the job's status per type). That makes the cap bound what the service actually counts — this is the "batches of 10" behaviour: **at most 10 jobs counted at once**. Cost is runtime: the floor becomes roughly `(#jobs × drain) / cap`, so ~13 min at cap 10 versus ~2 min for the (quota-breaching) early-release version. Both waits degrade to a logged warning rather than a failure, since the cap is a courtesy to the quota, not an assertion about the SDK. Enforced in the harness, so a new test is capped by default. `SHALLOW_MAX_CONCURRENT_JOBS=0` disables it for single-worker debugging.

## Verification

Run against a real account. The shallow suite alone (us-west-2):

```
83 passed, 1 skipped, 0 failed in 13m30s
```

Zero `ResourceLimitExceeded`, and live counted concurrency held at **4–5** against the cap of 10 for the whole run (account fully drained afterward). The 13:30 is the cost of holding each slot until its job is terminal — see the concurrency-cap section; the earlier ~3m33s figure was the early-release version that breached the quota. The `SimulatePrincipalPolicy` memoization described below still applies underneath.

The skip is RLAIF's serverful case, which reports its own reason: `RLAIFTrainer takes no compute argument`.

The five `us_east_1` tests, run separately with us-east-1 credentials:

```
5 passed in 47s
```

Those five had **never executed before this PR** — see the bug list below.

Also run: the whole `tests/integ/train` tree under the same selection the CodeBuild job uses (`-m "not gpu_intensive and not us_east_1"`, `-n 8 --dist loadfile`) — **198 passed, 9 skipped, 0 failed**. Skips all report their own prerequisite: Docker absent locally (5), RLAIF's serverful case, and three pre-existing `skip`/prerequisite cases. That run predates the marker changes below, so it includes the five us-west-2 evaluator tests now deselected; it is the "before" measurement, and the runtime caveat in the next section is what came out of it.

Marker selection after this PR, verified by collection: **266 of 342 collected, 76 deselected**, none of the six newly marked evaluator tests among them, and no `PytestUnknownMarkWarning` remaining.

## What bounded the `integ-tests` job (pre-existing, now fixed here)

Not the shallow suite, and not the new job — `fast-integ-tests` is the 5-minute one. This is about the CodeBuild job, and it is worth stating plainly because I got it wrong earlier in this PR's own description.

**CI proved it on this branch.** From the gate's own run, the serial pass durations:

```
2783.83s  test_llm_as_judge_base_model_fix.py::...::test_base_model_evaluation_uses_correct_weights
2504.59s  test_llm_as_judge_base_model_fix.py::...::test_base_model_false_still_works
  91.44s  the next-slowest test in that pass
```

**88 minutes for two tests, against the project's 180-minute build timeout**, and they were the entire tail — everything else in that pass finished in under 92s. Six tests block on `execution.wait(..., timeout=14400)`, a 4-hour ceiling each, and none of them was marked:

| Test | Was | Now |
|---|---|---|
| `test_llm_as_judge_base_model_fix.py::test_base_model_evaluation_uses_correct_weights` | `serial` | `gpu_intensive` (class) |
| `test_llm_as_judge_base_model_fix.py::test_base_model_false_still_works` | `serial` | `gpu_intensive` (class) |
| `test_benchmark_evaluator.py::test_benchmark_evaluation_full_flow` | none | `gpu_intensive` |
| `test_custom_scorer_evaluator.py::test_custom_scorer_evaluation_full_flow` | `xdist_group` | `gpu_intensive` |
| `test_llm_as_judge_evaluator.py::test_llm_as_judge_evaluation_full_flow` | none | `gpu_intensive` |
| `test_llmaj_custom_model.py::TestLLMAJCustomModelIntegration` | `@pytest.mark.slow` — **unregistered, silent no-op** | `gpu_intensive` |

The last row was a genuine mismarking: the registered name is `slow_test`, so `@pytest.mark.slow` did nothing and only raised a `PytestUnknownMarkWarning`. `us_east_1` already kept it off the us-west-2 gate, so marking it changes nothing there; it now stays off the us-east-1 job too.

**An earlier revision of this description said I deliberately would not mark these, on the grounds that they have no shallow counterpart. The 88-minute measurement changed my mind, and it is a real trade rather than a free win, so here is what it costs.** Three of the files are marked per-test and keep their constructor/validation tests on the gate — that is where SDK-side regressions surface. The two class-level ones leave nothing behind, and what the gate stops checking there is that a submitted evaluation pipeline is *accepted and succeeds*. Their already-marked siblings (`test_benchmark_evaluation_base_model_only`, `test_custom_scorer_base_model_only`) show this was already the established call for pipeline-waiting tests — these six were unmarked by omission, not by decision. Shallow `evaluate()` coverage is the follow-up that closes the gap properly.

`fast-integ-tests` existing as its own check is the other half of the answer: it returns in minutes regardless of what the deep suite is doing.

### Also fixed: `SimulatePrincipalPolicy` throttling

The same CI run failed four shallow tests, and they were not test defects:

```
FAILED tests/integ/train/shallow/test_model_trainer.py::TestSourceCodePackaging::test_shell_entry_script
FAILED tests/integ/train/shallow/test_rlaif_trainer.py::TestRLAIFTrainerSubmission::test_mlflow_resource_arn
FAILED tests/integ/train/shallow/test_rlvr_trainer.py::TestRLVRTrainerSubmission::test_with_validation_dataset
FAILED tests/integ/train/shallow/test_rlvr_trainer.py::TestRLVRTrainerSubmission::test_dataset_passed_to_train_overrides_constructor
ClientError: (Throttling) ... SimulatePrincipalPolicy (reached max retries: 9): Rate exceeded
```

All four pass locally in isolation and under a local `-n 36` run. Every `ModelTrainer` construction calls `TrainDefaults.get_role` → `resolve_and_validate_role`, which paginates `SimulatePrincipalPolicy` over ~20 action names against a low, **account-wide** TPS limit. CodeBuild runs the whole tree under `-n auto` (~36 workers on a 2XLARGE) — 188 trainer constructions, enough to exhaust even the adaptive 10-attempt budget the existing retry fixture grants. The cause is volume, not burstiness, so more retries would not have helped.

Fixed with a `_memoize_role_validation` autouse session fixture: each distinct `(role, role_type, region)` is validated once per xdist worker instead of once per test. Measured with an instrumented `botocore` `_make_api_call` — 3 trainers → 3 calls unpatched, 10 trainers → 1 call memoized. Exceptions are cached alongside successes so a bad role still fails (`test_unassumable_role_is_rejected` still passes), and teardown restores any caller now holding the memoized function rather than only the ones the fixture explicitly patched, since a module imported after the patch binds it at its own import time.

**Cost model measured, not assumed.** Across ~110 training jobs created by these runs, in both us-west-2 and us-east-1, every one ended `Stopped` and every `BillableTimeInSeconds` was `null` — jobs are torn down while still in `Starting`/`Pending`, before instances become billable.

Real bugs the AWS runs found in these tests (all wrong assumptions on my part, not service problems), each now fixed with the evidence recorded in a comment:

- the public botocore model has no `ServerlessJobConfig.SequenceLength`, so those requests failed client-side before reaching the service — fixed with an `AWS_DATA_PATH` fixture mirroring the one in `test_recipe_override_integration.py`
- tuning job names cap at **32** characters, not the 63 allowed for training jobs
- the tuner derives its own job name from the training image plus a second-granularity timestamp and ignores `base_job_name`, so two tuner tests in the same second collided with `ResourceInUse`
- `ModelTrainer.__init__` validates the execution role, so a bad role is rejected at construction and never reaches `CreateTrainingJob`
- `RLAIFTrainer` takes no `compute`; `CPTTrainer` takes no `training_type` and requires HyperPod
- for this model the recipe catalogue offers only `4K` sequence length and no serverless recipe for `TrainingType.FULL`
- RLVR reward functions are pre-validated by *invoking* them over sample records, so they need GSM8k-shaped data — this suite's generic chat-format fixture fails with `GSM8k scoring failed`
- `list_mlflow_apps` is not a paginatable operation
- **the five `us_east_1` tests had never actually run.** They referenced a bucket hardcoded to one test account (`s3://sagemaker-us-east-1-784379639078/...`), which other accounts cannot read — `AccessDenied` on `ListObjectsV2` from 729646638167. They now derive every path from `default_bucket()` and resolve the reward function from the caller's own hub, the way `test_sft_trainer_serverful_smtj.py::training_resources` already does. All five pass.
- a **model package group must be in the job's region** — passing the us-west-2 ARN the other trainers use is rejected with `Model package group ARN region 'us-west-2' does not match expected region 'us-east-1'`, so the Nova files use a bare name that resolves per-session
- `gpu_intensive` and `us_east_1` were declared only in `tox.ini`, but pytest reads its config from `pyproject.toml` here, so **both markers were unregistered at runtime**. That matters when the gate selects with `-m "not gpu_intensive and not us_east_1"`: a typo'd marker name would silently put an expensive deep test back on the gate instead of warning. Now registered where pytest reads them.

## What this deliberately does NOT cover

Training *behaviour*: no model artifacts, no metrics, no container logs, no convergence.

**A regression that breaks training itself — a bad entry script, a broken container command, a distributed-launch bug — will pass this gate** and be caught by the scheduled suites instead. That is the accepted trade for the runtime and cost reduction, and it is stated plainly in `tests/integ/train/shallow/README.md`.

Also out of scope for this pattern: HyperPod (submits to a pre-provisioned cluster, not `CreateTrainingJob`) and local container mode (`test_local_model_trainer.py` — no service call, but it runs real containers, and it stays on the CodeBuild job which already has Docker).

Known remaining gap: evaluator `evaluate()` submissions. Same pattern applies — assert the pipeline execution ARN comes back without waiting for the pipeline to finish — but it is a distinct API surface, so it is left for a follow-up. Until then the gate checks that evaluators construct and validate correctly, but not that a submitted pipeline is accepted.

## Note for reviewers

Because this workflow triggers on `pull_request_target`, CI runs the workflow definition from the **base branch**, not this PR's. So the `fast-integ-tests` job added here does not appear in this PR's own checks; it starts running once merged. The test selection was verified locally (numbers above); the job definition itself is unproven in CI.

`integ-tests (sagemaker-train)` does run on this PR, as it will after merge, and it runs *this branch's* test code — which is how the 88-minute evaluator tail and the `SimulatePrincipalPolicy` throttling were both found. What it runs against is master's *workflow definition*, so the new job's absence from these checks is expected; the marker changes, being in the test files themselves, are exercised on the next run of this PR.


.github/workflows/pr-checks-master.yml
```diff
@@ -243,6 +243,118 @@ jobs:
           project-name: ${{ github.event.repository.name }}-ci-${{ matrix.submodule }}-integ-tests
           source-version-override: 'refs/pull/${{ github.event.pull_request.number }}/head^{${{ github.event.pull_request.head.sha }}}'
 
+  # Additive: runs the shallow (submit-then-stop) suite for sagemaker-train
+  # alongside the existing integ-tests job above, which is unchanged.
+  #
+  # Why a separate job rather than folding this into the CodeBuild suite: this
```
jam-jee, line 249: We will move this job to codebuild after initial POC and outcome.

sagemaker-train/tests/integ/train/shallow/test_tuner.py
```diff
@@ -0,0 +1,192 @@
...
+        objective_type="Minimize",
+        max_jobs=1,
+        max_parallel_jobs=1,
+    )
+    kwargs.update(overrides)
+    return HyperparameterTuner(**kwargs)
+
+
+@contextmanager
+def _tuning(tuner, job_name):
```
papriwal, line 117: Tuner path escapes the concurrency cap. `_tuning()` calls `tuner.tune()` directly (no `job_slots()`, no `_wait_until_terminal()`), yet the `README`, the `DEFAULT_MAX_CONCURRENT_JOBS` note, and `_requested_slots` all describe the tuner as inside the cap. Peak overshoot is small (both tuners are max_jobs=1), but it's the release-before-terminal
  pattern the cap note blames for the prior ~37-concurrent breach. Wrap in with `job_slots(...)` and drain after `stop_tuning_job()`, or document the exemption.

sagemaker-train/pyproject.toml
```diff
@@ -84,6 +84,15 @@ addopts = ["-vv"]
...
     "serial: marks tests that must run serially (not in parallel)",
+    # gpu_intensive and us_east_1 are declared in tox.ini too, but pytest reads
+    # its config from this file (it is the first of the candidates present), so
+    # markers listed only there are unregistered at runtime and raise
+    # PytestUnknownMarkWarning. Registering them here matters because the PR gate
+    # selects with -m "not gpu_intensive and not us_east_1": a typo'd marker name
+    # would otherwise silently put an expensive deep test back on the gate instead
+    # of warning.
+    "gpu_intensive: marks a test that consumes real training capacity (scheduled CI, not PR checks); see tests/integ/train/shallow",
+    "us_east_1: marks a test that must run in us-east-1 (Nova); runs in the us-east-1 integ job",
```
papriwal, line 95: Instead of using `us_east_1` as the test marker. We should use a more contextual name talking about the tests being related to `nova`?

sagemaker-train/tests/integ/train/shallow/recipe_cases.py
```diff
@@ -0,0 +1,282 @@
...
+# The Nova files (us-east-1) cannot reuse the ARN above. Verified against AWS: the
+# model package group must be in the same region as the job, and passing the
+# us-west-2 ARN is rejected with
+#
+#     Model package group ARN region 'us-west-2' does not match expected region
+#     'us-east-1'
+#
+# A bare name resolves in whichever region the session is in. Shared here rather
+# than duplicated per Nova file so the two cannot drift apart.
+NOVA_MODEL_PACKAGE_GROUP = "sdk-test-finetuned-models"
```
papriwal: `recipe_cases.py `pins `MODEL_PACKAGE_GROUP` to an account ARN while the Nova path uses the bare `NOVA_MODEL_PACKAGE_GROUP`; prefer the bare name in both.

sagemaker-train/tests/integ/train/shallow/conftest.py
```diff
@@ -0,0 +1,360 @@
...
+    try:
+        return client.describe_hub_content(
+            HubName=hub, HubContentType="JsonDoc", HubContentName=name
+        )["HubContentArn"]
+    except Exception as e:
+        pytest.skip(f"Reward function {name!r} not in hub {hub!r}: {e}")
+
+
+@pytest.fixture(scope="module")
+def reward_scored_data_uri():
```
papriwal, line 239: Should be named `reward_scored_dataset_uri`?

sagemaker-train/tests/integ/train/shallow/harness.py
```diff
@@ -0,0 +1,633 @@
...
+# A small CPU instance is sufficient: acceptance of the request does not depend
+# on the instance type being an accelerator, and asking for GPU capacity we
+# immediately discard is both slower and antisocial in a shared test account.
+DEFAULT_INSTANCE_TYPE = "ml.m5.large"
+DEFAULT_INSTANCE_COUNT = 1
+
+# Public DLC, present in every commercial region we test in. Using a real image
+# matters: the backend's role-assuming validators resolve the training image
+# against ECR, so a bogus URI would fail for the wrong reason.
+CPU_IMAGE = "763104351884.dkr.ecr.us-west-2.amazonaws.com/pytorch-training:2.0.0-cpu-py310"
```
papriwal: If this `CPU_IMAGE` is present in every commercial region would be beneficial to be region agnostic here. Reduces the blast radius of mishaps.

jam-jee:
Agreed, and done in `8666594a` — the two constants are now a single bare-name `MODEL_PACKAGE_GROUP` used by both regions.

Checking the SDK, either form works: `_resolve_model_package_group_arn()` returns an ARN unchanged, and otherwise resolves a name via `ModelPackageGroup.get()` against the **session's** region. So a name is region- and account-portable where an ARN pins both — and pinning the region is what forced the split originally, since passing the us-west-2 ARN to a us-east-1 Nova job is rejected with `Model package group ARN region 'us-west-2' does not match expected region 'us-east-1'`. It also drops a hardcoded account ID from a public repo.

Verified: the bare name resolves to the same ARN via `DescribeModelPackageGroup` in us-west-2, and the us-west-2 recipe path still submits — SFT, DPO and RLVR minimal-request tests pass against the service (3 passed). Collection unchanged at 100 tests.

sagemaker-train/tests/integ/train/shallow/harness.py
jam-jee: Good point, will update it.

sagemaker-train/tests/integ/train/shallow/recipe_cases.py
jam-jee: nice,sounds good, updated.

sagemaker-train/pyproject.toml
jam-jee: yes, these are existing markers, need to reword these. We will be removing most of the markers and merging them together.

papriwal (approved):
LGTM!

PR #6187, opened 2026-08-17
feat(train): Add inherited list_supported_models to BaseTrainer

jam-jee:
## Problem

`MultiTurnRLTrainer` exposes `list_supported_models()` so users can discover which
hub models have a matching recipe, but the other fine-tuning trainers
(`SFTTrainer`, `RLVRTrainer`, `RLAIFTrainer`, `DPOTrainer`, `CPTTrainer`) have no
equivalent — users have no programmatic way to find which models support a given
technique.

## Why it matters

Without this, users must consult external docs or trial-and-error a model name into
`train()` to find out whether a technique is supported for it. A discovery method
per trainer makes the supported-model set self-describing from the SDK.

## Fix (symptom → root cause → change)

The listing logic is already generic: `MultiTurnRLTrainer.list_supported_models`
just delegates to `recipe_utils._list_hub_models_by_recipe(recipe_type, technique,
session)`, and the hub keyword filter is keyed off the trainer's
`_customization_technique` value (`@recipe:finetuning_{technique}_...`). Rather than
copy the method into each trainer, add a single inherited `classmethod` on
`BaseTrainer` that resolves `cls._customization_technique` and delegates to that
same primitive. Every fine-tuning trainer inherits it for free:

- `SFTTrainer.list_supported_models()` → technique `SFT`
- `RLVRTrainer.list_supported_models()` → technique `RLVR`
- `RLAIFTrainer.list_supported_models()` → technique `RLAIF`
- `DPOTrainer.list_supported_models()` → technique `DPO`
- `CPTTrainer.list_supported_models()` → technique `CPT`

`RLAIFTrainer` previously referenced its technique inline and lacked the
`_customization_technique` class attribute, so it is added there for the inherited
method to resolve. `MultiTurnRLTrainer` keeps its own override (it is a CreateJob
flow whose technique is not a `CustomizationTechnique` enum value and delegates to
`_list_all_mtrl_models`). If a subclass defines no technique, the method raises
`NotImplementedError` with a clear message.

## Tests

- `test_base_trainer_compute.py`: base classmethod delegates with the class's
  technique; raises `NotImplementedError` when the technique is missing.
- `test_sft_trainer.py` / `test_rlvr_trainer.py` / `test_dpo_trainer.py` /
  `test_rlaif_trainer.py`: each subclass resolves and passes its own technique
  (`SFT`/`RLVR`/`DPO`/`RLAIF`) through the inherited method, including session
  passthrough.

All 250 unit tests across the touched trainer suites (including the existing MTRL
suite, which confirms the base method does not clash with MTRL's override) pass.

## Manual verification

N/A — unit coverage is sufficient; the change is a thin delegation to the already
integration-tested `_list_hub_models_by_recipe` helper.

---
X-AI-Prompt: Add list_supported_models to SFT/RLVR/RLAIF/DPO trainers like MTRL; refactored to a shared BaseTrainer method
X-AI-Tool: Kiro

### DIFF
--- a/sagemaker-train/src/sagemaker/train/base_trainer.py
+++ b/sagemaker-train/src/sagemaker/train/base_trainer.py
@@ -102,6 +102,37 @@ class BaseTrainer(ABC):
     training_image: Optional[str] = None
     latest_training_job: Optional[TrainingJob] = None
 
+    @classmethod
+    @_telemetry_emitter(
+        feature=Feature.MODEL_CUSTOMIZATION,
+        func_name="BaseTrainer.list_supported_models",
+    )
+    def list_supported_models(cls, session=None) -> List[str]:
+        """Return the models that support this trainer's fine-tuning technique.
+
+        Queries SageMakerPublicHub for all models whose ``RecipeCollection``
+        contains a FineTuning recipe for this trainer's customization technique
+        (``cls._customization_technique``, e.g. ``"SFT"``, ``"DPO"``,
+        ``"RLVR"``, ``"RLAIF"``, ``"CPT"``).
+
+        Args:
+            session: Optional boto3 session.
+
+        Returns:
+            Sorted list of hub content model names supporting the technique.
+        """
+        from sagemaker.train.common_utils.recipe_utils import _list_hub_models_by_recipe
+
+        technique = getattr(cls, "_customization_technique", None)
+        if not technique:
+            raise NotImplementedError(
+                f"{cls.__name__} does not define a customization technique and "
+                "cannot list supported models."
+            )
+        return _list_hub_models_by_recipe(
+            recipe_type="FineTuning", technique=technique, session=session
+        )
+
     def __init__(
         self,
         sagemaker_session: Optional[Session] = None,
--- a/sagemaker-train/src/sagemaker/train/common_utils/recipe_utils.py
+++ b/sagemaker-train/src/sagemaker/train/common_utils/recipe_utils.py
@@ -506,20 +506,25 @@ def resolve_recipe(
 
 
 def _build_recipe_keyword(recipe_type: str, technique: str) -> str:
-    """Build the ``@recipe:`` search keyword for a recipe type and technique.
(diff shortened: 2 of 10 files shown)


jam-jee:
Verified fortress reported issues are LOW severity.

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -102,6 +102,37 @@ class BaseTrainer(ABC):
...
+        (``cls._customization_technique``, e.g. ``"SFT"``, ``"DPO"``,
+        ``"RLVR"``, ``"RLAIF"``, ``"CPT"``).
+
+        Args:
+            session: Optional boto3 session.
+
+        Returns:
+            Sorted list of hub content model names supporting the technique.
+        """
+        from sagemaker.train.common_utils.recipe_utils import _list_hub_models_by_recipe
```
rsareddy0329, line 124: Nit[non-blocking]: We can move the import to above

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -102,6 +102,37 @@ class BaseTrainer(ABC):
     training_image: Optional[str] = None
     latest_training_job: Optional[TrainingJob] = None
 
+    @classmethod
+    @_telemetry_emitter(
+        feature=Feature.MODEL_CUSTOMIZATION,
+        func_name="BaseTrainer.list_supported_models",
```
rsareddy0329, line 108: non blocking but a callout: we might not be able to trace which Trainer is used to list supported models.

rsareddy0329:
As this is public discoverable method, worth adding docs for the same.

rsareddy0329 (approved):

sagemaker-train/src/sagemaker/train/base_trainer.py
jam-jee: good point. But I dont think we need granular list_supported_models metrics ATM.

PR #6192, opened 2026-08-20
fix(rlaif): accept preset reward_prompt template names

rsareddy0329:
*Issue #, if available:*  
`reward_prompt` did not accept preset template names via the Python SDK. Passing "summarize.jinja", "summarize", or any plain string (including the example from the SDK docs, "Rate the helpfulness of this response on a scale of 1-10") caused the SDK to treat the value as a HubContent name and call DescribeHubContent, failing with a validation error.
  
Only strings with an exact "Builtin" prefix were routed to preset resolution; everything else fell through to a HubContent lookup.

*Description of changes:*
  Resolve preset template names directly against the recipe's judge_prompt_template enum (which is already loaded in memory as hyperparameters._specs — no API
  call), matching RLVR's preset behavior. New resolution order:
  
  1. Preset template name → resolved locally against the enum. Accepts "Builtin.Summarize", "summarize", and "summarize.jinja".
  2. Evaluator ARN (arn:aws:sagemaker:...) → validated/assigned as-is.
  3. Otherwise → HubContent name lookup (registered custom prompt), raising a clear error if not found.
  
  Details:
  
  - Added _normalize_template_name — strips an optional Builtin. prefix, takes the basename, strips an optional .jinja suffix, lowercases. So "Builtin.Summarize",
  "summarize", "summarize.jinja", and enum values like /opt/ml/code/verl/summarize.jinja all normalize to summarize.
  - Added _is_preset_reward_prompt and _get_judge_prompt_template_enum helpers; reordered routing to preset → ARN → HubContent.
  - Rewrote _update_judge_prompt_template_direct to use normalized matching (no longer assumes the Builtin. prefix) with a clearer error listing valid presets.
  - Preset detection/resolution is local; the HubContent API call only happens for genuine custom-prompt names.
  - Updated the SDK docstring examples from the invalid raw string to reward_prompt="summarize".


sagemaker-train/src/sagemaker/train/rlaif_trainer.py
```diff
@@ -60,7 +60,7 @@ class RLAIFTrainer(BaseTrainer):
             model="meta-llama/Llama-2-7b-hf",
             model_package_group="my-rlaif-models",
             reward_model_id="reward-model-id",
-            reward_prompt="Rate the helpfulness of this response on a scale of 1-10"
```
mujtaba1747, line 63: Do we still accept inline prompt?

mujtaba1747 (approved):

sagemaker-train/src/sagemaker/train/rlaif_trainer.py
rsareddy0329: No, we do not accept inline prompt currently as well, this input raises an error.
Correcting this incorrect example previously given in doc strings.

jam-jee (approved):

PR #6196, opened 2026-08-20
fix: forwarded role to Evaluator

zhaoqizqwang:
RLVRTrainer does not forward the user-provided role to its internal Evaluator.create() call, causing role resolution failures for users whose caller identity cannot auto-resolve, and the required Lambda response format (aggregate_reward_score) is undocumented in the training interface.


rsareddy0329 (approved):
Can we include unit tests as well.

jam-jee:
> No test coverage for the changed path (testing)

Can we add relevant tests.

PR #6197, opened 2026-08-21
fix(core,mlops): honor caller region in feature_store ingest_dataframe and stop telemetry from blocking SDK calls

nargokul:
## Issue

`sagemaker.mlops.feature_store.feature_utils.ingest_dataframe()` had no way for a caller to say which AWS region the FeatureGroup lives in. Both the `DescribeFeatureGroup` call and the record writes were made with no region, so the region was whatever boto3 happened to resolve for the process. Callers who work with a FeatureGroup outside their resolved default region had no way to redirect the call.

Note that `list_records()` in the same module already accepts `region`, so this also makes the module consistent.

## Description of changes

Adds an optional `region` argument to `ingest_dataframe()` and threads it through the whole ingestion path:

- `feature_utils.ingest_dataframe(..., region=None)`
  - `CoreFeatureGroup.get(feature_group_name=..., region=region)`
  - `IngestionManagerPandas(..., region=region)`
- `IngestionManagerPandas` gains a `region` field, forwarded to the FeatureStore runtime calls in every ingestion path:
  - single process / single thread → `put_record(region=...)`
  - `_ingest_single_batch` → `put_record(region=...)`
  - `_run_multi_threaded` (passed positionally through the process pool args) → both batch paths
  - `_ingest_batch_write` → `batch_write_record(region=...)`

`region` is appended as the last parameter and defaults to `None`, so existing positional and keyword calls keep working and behave exactly as before.

```python
from sagemaker.mlops.feature_store import ingest_dataframe

manager = ingest_dataframe(
    feature_group_name="my-fg",
    data_frame=df,
    region="eu-west-1",
)
```

Also documents the new argument in the feature store `MIGRATION_GUIDE.md`.

### Known limitation (not addressed here)

`sagemaker-core`'s `SageMakerClient` is a process-wide singleton keyed only on the class (`sagemaker-core/src/sagemaker/core/utils/utils.py`), so the first region used in a process wins and later `region=` arguments are ignored:

```python
Base.get_sagemaker_client(region_name="us-east-1")  # -> us-east-1
Base.get_sagemaker_client(region_name="eu-west-1")  # -> us-east-1 (silently reused)
```

So `region` here is honored when `ingest_dataframe` is the first thing in the process to build a client (the common case), and is subject to that cache otherwise. This affects every `region`/`session` argument in sagemaker-core equally, not just this function, so it is called out in the docstring and left for a separate fix in sagemaker-core.

## Testing

New unit tests (14), all passing:

`tests/unit/sagemaker/mlops/feature_store/test_feature_utils.py::TestIngestDataframeRegion`
- region reaches both `CoreFeatureGroup.get` and `IngestionManagerPandas`
- defaults to `None` when not supplied
- works together with `use_batch_write_record=True`
- signature check that `region` is appended last and does not shift existing positional args

`tests/unit/sagemaker/mlops/feature_store/test_ingestion_manager_pandas.py::TestIngestionManagerRegion`
- `region` stored on the manager, defaults to `None`
- `_ingest_row` forwards region to `put_record` (and passes `None` when unset)
- single-threaded `run()` forwards region to every `put_record`
- `_ingest_single_batch` forwards region to every `put_record`
- `_ingest_batch_write` and a `use_batch_write_record=True` `run()` forward region to `batch_write_record`
- `_run_multi_threaded` forwards region to each thread's batch
- `_run_multi_process` includes region in the process pool args

Full existing suite for the area still green: 375 passed in `sagemaker-mlops/tests/unit/sagemaker/mlops/feature_store/` (the `feature_processor` subdirectory is excluded — it fails collection on `ModuleNotFoundError: No module named 'pyspark'` on master too, unrelated to this change).

---

# Second change: telemetry no longer blocks SDK calls

## Issue

A customer reported `ingest_dataframe` taking over 45 minutes from a private VPC, while the same write from boto3 was fast. The service-side worklog shows the underlying `PutRecord` completed in **348 ms**, about two seconds into a notebook cell that ran for ~47 minutes. All of the remaining time was client-side, spent in the telemetry emissions that run *after* the wrapped function returns.

There are exactly two telemetry-decorated calls in that flow (`ingest_dataframe` and `IngestionManagerPandas.run`), which matches ~23.5 minutes per emission: two identical unbounded waits.

Three separate defects combine to produce it.

**1. The telemetry GET had no timeout at all.**

```python
response = requests.get(url, timeout)   # binds to params=, not timeout=
```

`requests.get(url, params=None, **kwargs)` takes `params` second, so the `2` was appended to the query string (the prepared URL ends in `&2`) and the request was left with no timeout. From a VPC with no route to the endpoint the call hangs until some network device drops the flow.

**2. The fallback session hardcoded `us-west-2`.**

`ingest_dataframe` is a module-level function with no session, so the decorator synthesizes one via `_get_default_sagemaker_session()`, which was `boto3.Session(region_name=DEFAULT_AWS_REGION)`. That pointed both the STS `get_caller_identity` call and the telemetry GET at public us-west-2 endpoints, regardless of the caller's actual region (ca-central-1 here). This is also why the `region` argument added above does **not** fix the hang on its own: telemetry ignored the caller's region entirely.

**3. Emission was synchronous.** Both network legs sat directly in the caller's critical path.

The team's successful repro without a VPC is consistent with this rather than ruling the SDK out: without egress restrictions both legs complete in milliseconds, which is exactly how a client-side call to an unreachable endpoint behaves.

## Description of changes

`sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py`:

- `_requests_helper` passes `timeout=` as a keyword, with a comment explaining the `params` trap so it does not regress. The timeout value is now the named `TELEMETRY_REQUEST_TIMEOUT` constant.
- `_get_default_sagemaker_session` lets boto3 resolve the region from the caller's environment (`AWS_REGION`, `AWS_DEFAULT_REGION`, or the active profile), falling back to `DEFAULT_AWS_REGION` only when boto3 resolves nothing, since `Session` requires a region.
- `_send_telemetry_request` now dispatches to a daemon thread and returns immediately; the existing body moved unchanged to `_send_telemetry_request_sync`. This covers both slow legs, since the STS call lives inside that body. Daemon threads are killed at interpreter exit, so a pending send cannot delay shutdown either. Nothing can escape the thread. **No event is dropped** - one thread per event, no in-flight cap.

  Worth calling out why the send has to move off the caller's thread rather than just having its timeout tightened: Feature Store ingestion is decorated at more than one level (`ingest_dataframe` and `IngestionManagerPandas.run`), so a single user call emits several events. Even a bounded 2s per event adds up on the caller's thread, and unbounded it is what produced the 45+ minute cell in this ticket.
- Request failures log at debug instead of emitting a full traceback at error level; a best-effort metric should not surface as an error.

`sagemaker-serve/src/sagemaker/serve/utils/telemetry_logger.py` had the identical positional-timeout defect in its own `_requests_helper`; fixed the same way.

### Trade-off

Telemetry events still in flight when the process exits may be lost, because the threads are daemons. That is the one case where an event can go missing, and it is intentional: telemetry must not be able to hold up a customer's call or their interpreter shutdown. Events are never dropped by the SDK itself.

### Still outstanding (not in this PR)

`sagemaker-serve`'s `_send_telemetry` remains synchronous. It always uses the caller's real session and region, so it does not have the cross-region problem, and with the timeout fix its GET is now bounded at 2s. Its `_get_accountId` STS call still runs with default botocore timeouts, though. Happy to make that path async too if preferred.

## Testing

10 new unit tests in `sagemaker-core/tests/unit/telemetry/test_telemetry_logging.py`:

`TestRequestsHelperTimeout`
- timeout is passed as a keyword and `params` is not set
- the URL does not end in `&2` and the only kwarg is `timeout`

`TestTelemetryIsNonBlocking`
- `_send_telemetry_request` returns to the caller while the send is still blocked
- the send runs on a daemon thread
- all arguments are forwarded to `_send_telemetry_request_sync`
- an exception in the thread does not escape
- no event is dropped when many are in flight (25 concurrent events, all 25 reach the sender)
- a decorated function returns without waiting on a blocked telemetry send (the regression guard for this issue)

`TestDefaultSessionRegion`
- the session uses the region boto3 resolves, not a hardcoded one
- falls back to `DEFAULT_AWS_REGION` only when no region resolves

Existing tests that asserted the old behaviour were updated: the six that exercise URL construction now call `_send_telemetry_request_sync` directly, and the `requests.get` assertions expect `timeout=` as a keyword.

Also verified at the unit level against a black-holed address (`10.255.255.1`), which hangs rather than refusing:

```
_requests_helper against a black hole returned in 2.01s   (was unbounded)
_send_telemetry_request returned to the caller in 0.0005s
background send finished at 2.00s
```

### End-to-end reproduction of the ticket

Ran the customer's notebook as a script against a real Feature Group in a dev account: create feature group (offline store only) -> poll until `Created` -> `ingest_dataframe(data_frame=df, max_workers=1, wait=True)` with a single row, exactly as in the ticket. The customer's no-egress VPC was imitated by pointing DNS for `sm-pysdk-t-*` and `sts.*` at the non-routable `10.255.255.1`, so those hostnames resolve but never connect.

| telemetry code | network | `ingest_dataframe` |
|---|---|---|
| before this PR | black-holed (customer's VPC) | **770 s (12.8 min)** |
| before this PR | open (the non-VPC retry) | 1.94 s |
| after this PR | black-holed | **1.40 s** |
| after this PR | open | 1.22 s |

Two things this settles:

1. The hang is reproducible and is entirely client-side telemetry. Same SDK code, same feature group, same one-row frame; only reachability of the telemetry and STS endpoints differs.
2. The team's "it works without a VPC" retry does not exonerate the SDK. Row 2 is that retry, on the unfixed code, and it is fast for the same reason: the endpoints were reachable, so the unbounded wait never triggered.

The local run measures 12.8 min rather than the reported 45+ because TCP connect-retry behaviour and proxy configuration differ between macOS here and the customer's environment; the failure mode is the same, only the length of the stall varies.

Suites run:
- `sagemaker-core/tests/unit`: 3550 passed, 22 skipped
- `sagemaker-core/tests/unit/telemetry` + both sagemaker-serve telemetry test files: 111 passed, 2 skipped
- `sagemaker-mlops/tests/unit/sagemaker/mlops/feature_store` (excluding `feature_processor`, which fails collection on missing `pyspark` on master too): still green

## Merge Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that my tests are not configured for a specific region or account
- [x] The changes are backward compatible

### DIFF
--- a/sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
+++ b/sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
@@ -70,16 +70,19 @@ _telemetry_msg_shown = False
 # Seconds to wait for the telemetry endpoint before giving up on an event.
 TELEMETRY_REQUEST_TIMEOUT = 2
 
-# Telemetry is best-effort, so it must never sit in the caller's critical path.
-# Every event is sent from a daemon thread: a slow or unreachable telemetry
-# endpoint (for example from inside a VPC with no route to it) can no longer
-# stall the SDK call the customer actually made, and because the threads are
-# daemons a pending send cannot delay interpreter shutdown either. If more than
-# MAX_IN_FLIGHT_TELEMETRY_REQUESTS sends are already outstanding we drop the
-# event rather than grow threads without bound.
-MAX_IN_FLIGHT_TELEMETRY_REQUESTS = 8
-_in_flight_telemetry_requests = 0
-_in_flight_telemetry_lock = threading.Lock()
+# Telemetry must never sit in the caller's critical path, so every event is sent
+# from a daemon thread. A slow or unreachable telemetry endpoint (for example from
+# inside a VPC with no route to it) can no longer stall the SDK call the customer
+# actually made, and because the threads are daemons a pending send cannot delay
+# interpreter shutdown either. No event is ever dropped: one thread is started per
+# event.
+#
+# Special case worth calling out: Feature Store ingestion is decorated at more than
+# one level (``ingest_dataframe`` and ``IngestionManagerPandas.run``), so a single
+# user call can emit several events. Sending them serially on the caller's thread is
+# what turned an ingest that the service finished in under a second into a
+# multi-minute wait, which is why the send is moved off that thread here rather than
+# only having its timeout tightened.
 
 FEATURE_TO_CODE = {
     str(Feature.SDK_DEFAULTS): 11,
@@ -440,23 +443,15 @@ def _send_telemetry_request(
 ) -> threading.Thread:
     """Schedule a telemetry event to be sent on a background daemon thread.
 
-    This returns as soon as the thread is started so that telemetry never adds
-    latency to the SDK call that triggered it. The event is dropped (and None
-    returned) if too many sends are already outstanding.
+    Every event is still sent; this only moves the send off the caller's thread so
+    that telemetry never adds latency to the SDK call that triggered it.
 
     Returns:
-        threading.Thread: The thread doing the send, or None if the event was
-            dropped. Callers generally ignore this; tests can join on it.
+        threading.Thread: The thread doing the send. Callers generally ignore this;
+            tests can join on it.
     """
-    global _in_flight_telemetry_requests  # pylint: disable=W0603
-    with _in_flight_telemetry_lock:
-        if _in_flight_telemetry_requests >= MAX_IN_FLIGHT_TELEMETRY_REQUESTS:
-            logger.debug("Too many telemetry requests in flight. Dropping this one.")
-            return None
-        _in_flight_telemetry_requests += 1
 
     def _run():
-        global _in_flight_telemetry_requests  # pylint: disable=W0603
         try:
             _send_telemetry_request_sync(
                 status, feature_list, session, failure_reason, failure_type, extra_info
@@ -467,9 +462,6 @@ def _send_telemetry_request(
             # _send_telemetry_request_sync can fail once the interpreter starts
             # tearing down. Telemetry is best-effort, so drop the event silently.
             pass
-        finally:
-            with _in_flight_telemetry_lock:
-                _in_flight_telemetry_requests -= 1
 
     thread = threading.Thread(target=_run, name="sagemaker-telemetry", daemon=True)
     thread.start()
--- a/sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
+++ b/sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
@@ -16,6 +16,7 @@ import logging
 import os
 import platform
 import sys
+import threading
 from time import perf_counter
 from typing import List
 import functools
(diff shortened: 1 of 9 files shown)


sagemaker-core/src/sagemaker/core/telemetry/telemetry_logging.py
```diff
@@ -66,6 +67,23 @@
 )
 _telemetry_msg_shown = False
 
+# Seconds to wait for the telemetry endpoint before giving up on an event.
+TELEMETRY_REQUEST_TIMEOUT = 2
+
+# Telemetry must never sit in the caller's critical path, so every event is sent
+# from a daemon thread. A slow or unreachable telemetry endpoint (for example from
+# inside a VPC with no route to it) can no longer stall the SDK call the customer
+# actually made, and because the threads are daemons a pending send cannot delay
```
aviruthen, line 76: Minor nit on whether you want to include all these comments here!

aviruthen (approved):

PR #6201, opened 2026-08-21
fix: resolve private hub Models and aliased references for modelTrainer

tanvikab4:
Description of changes
  
  **Summary**
  
  ModelTrainer.from_jumpstart_config(...) could not resolve models that live in a private hub — only public
  JumpStart models and (by a fragile assumption) plain private-hub references worked. This change fixes hub-content
  resolution so ModelTrainer reaches parity with ModelBuilder, supporting:
  
  - Public JumpStart models (unchanged)
  - Model References in a private hub (pointer to a public model)
  - Aliased references — filed under a hub content name that differs from the public model_id
  - Privately-owned Models authored directly into a private hub

***Background: v2 → v3 parity gap***
  
  ModelTrainer is the v3 PySDK equivalent of v2's JumpStartEstimator (sagemaker.jumpstart.estimator) — the JumpStart training-job abstraction. The v2 JumpStartEstimator already supported training from privately-owned Models in a private hub, not just Model References. This capability was lost in the initial v2→v3 migration of the PySDK, so the scope of this task was to restore it in v3 ModelTrainer.
  
  How v2 did it (verified against v2.245.0): when resolving hub content, v2 did not guess the content type from the hub name — it probed the service, trying ModelReference first and falling back to Model, in sagemaker/jumpstart/accessors.py::get_model_specs:
  
```
 # Users only input model id, not contentType, so first try ModelReference, then Model
  if hub_arn:
      try:
          model_specs = JumpStartModelsAccessor._cache.get_hub_model_reference(...)
          model_specs.set_hub_content_type(HubContentType.MODEL_REFERENCE)
          return model_specs
      except Exception:
          # Failed as ModelReference, try as a privately-owned Model
          model_specs = JumpStartModelsAccessor._cache.get_hub_model(...)
          model_specs.set_hub_content_type(HubContentType.MODEL)   # ← owned private-hub Model
          return model_specs
  
```
  The v3 rewrite replaced this probe with a name-based guess ("ModelReference" if private hub else "Model"), which broke the owned-Model path that v2 supported. This change restores v2's probe-with-fallback behavior in v3's  document.py::get_hub_content_and_document(), bringing ModelTrainer back to parity with the v2 JumpStartEstimator for private-hub training.
  

  
  **Root cause**
  
  sagemaker.core.jumpstart.document.get_hub_content_and_document() guessed the hub content type from the hub name:
  
  hub_content_type = "Model" if hub_name == SAGEMAKER_PUBLIC_HUB else "ModelReference"
  
  A private hub can hold either a Model or a ModelReference. This guess meant:
  
  - Privately-owned Models were looked up as ModelReference → ResourceNotFound → resolution failed.
  - The lookup used model_id and ignored hub_content_name, so aliased references were never found.
  
  Fix (sagemaker-core/src/sagemaker/core/jumpstart/document.py)
  
  - Replace the guess with a probe: for a private hub, try ModelReference first, then fall back to Model; the
  public hub uses Model only. This mirrors ModelBuilder's resolution in accessors.py.
  - Honor hub_content_name (falling back to model_id) so aliased references resolve.
  - On miss, raise a combined error naming both content types attempted.
  
  No changes were needed elsewhere: defaults.py already attaches HubAccessConfig based on hub_content_type, and
  model_trainer.py / JumpStartConfig already support hub_name/hub_content_name — they become correct automatically
  once the content type is resolved honestly.
  
  **Testing**
  
  Unit (sagemaker-core/tests/unit/jumpstart/test_document.py) — 5 new tests, all passing:
  
  - public hub resolves as Model (single lookup, no probe)
  - private-hub reference resolves on the first probe
  - private-hub Model resolves via the fallback (asserts probe order ["ModelReference", "Model"])
  - hub_content_name alias is used for lookup
  - neither type present → raises after attempting both
  
  Integration (sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py) — 3 new tests, each creates a
  temporary private hub, runs a real training job, and tears down (skips gracefully without hub permissions). All
  verified passing end-to-end against AWS:
  
  - test_jumpstart_train_from_private_hub_reference
  - test_jumpstart_train_from_aliased_reference
  - test_jumpstart_train_from_private_owned_model

### DIFF
--- a/sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
+++ b/sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
@@ -10,14 +10,231 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-"""This module contains the Integ Tests for JumpStart Training."""
+"""This module contains the Integ Tests for JumpStart Training.
+
+Coverage:
+  * Public JumpStart models (model_id only).
+  * Private-hub ModelReference (a pointer to a public model), including an
+    aliased reference whose hub content name differs from the public model_id.
+  * A privately-owned Model authored directly into a private hub.
+
+The private-hub / private-model tests each create their own temporary hub,
+populate it, run training, and tear the hub down. They skip gracefully if the
+environment lacks permissions to create hubs or import content.
+"""
+
 from __future__ import absolute_import
 
+import time
+import uuid
+import logging
+
 import pytest
+from botocore.exceptions import ClientError
 
 from sagemaker.core.jumpstart import JumpStartConfig
+from sagemaker.core.helper.session_helper import Session
 from sagemaker.train import ModelTrainer
 from sagemaker.train.configs import Compute
+
+logger = logging.getLogger(__name__)
+
+# A trainable classical-ML model keeps these tests fast/cheap on CPU.
+TRAINABLE_MODEL_ID = "catboost-regression-model"
+# Gated, trainable model reused from the v2 private-hub parity tests. Exercises
+# the accept_eula / ModelAccessConfig path for a gated ModelReference. Gated
+# models resolve to GPU and require real EULA acceptance, so the test that uses
+# it runs a real training job and is marked slow_test + gpu_intensive (scheduled
+# CI, not PR checks). The instance type is intentionally left to the SDK:
+# from_jumpstart_config validates a supplied instance_type against the model's
+# SupportedTrainingInstanceTypes and raises if it is not in the list, so
+# resolving the model's own default is safer than hardcoding one here.
+GATED_TRAINABLE_MODEL_ID = "meta-textgeneration-llama-3-2-1b"
+HUB_NAME_PREFIX = "sdk-integ-train-hub"
+ALIASED_REFERENCE_NAME = "sdk-integ-aliased-catboost"
+PRIVATE_MODEL_NAME = "sdk-integ-private-catboost"
+
+# Only these error codes represent "this restricted account is not allowed to
+# set up the fixture" and warrant a graceful skip. Any other ClientError is a
+# real failure (e.g. a service-side regression in create_hub_content_reference
+# or import_hub_content) and must fail loudly so the test does not silently
+# stop guarding the fix while CI stays green.
+_SKIPPABLE_SETUP_ERROR_CODES = frozenset(
+    {
+        "AccessDeniedException",
+        "AccessForbiddenException",
+        "UnauthorizedOperation",
+    }
+)
+
+
+def _skip_if_unauthorized(e, message):
+    """Skip only on an expected authorization error; re-raise everything else."""
+    if e.response.get("Error", {}).get("Code") in _SKIPPABLE_SETUP_ERROR_CODES:
+        pytest.skip(f"{message}: {e}")
+    raise
+
+
+def _assert_reference_channels(model_trainer):
+    """Assert the SDK resolved a ModelReference into hub-aware training channels.
+
+    Resolving a ModelReference must produce a container image and attach a
+    HubAccessConfig(hub_content_arn=...) to the model channel (defaults.py,
+    hub_content_type == "ModelReference" branch). Asserting this on the
+    SDK-resolved channels — rather than passing an explicit training channel to
+    train() — is what actually guards the fix; a non-gated model would train
+    fine even if this plumbing regressed.
+    """
+    assert model_trainer.training_image
+    model_channels = [
+        c for c in model_trainer.input_data_config if getattr(c, "channel_name", None) == "model"
+    ]
+    assert len(model_channels) == 1
+    hub_access_config = model_channels[0].data_source.s3_data_source.hub_access_config
+    # A resolved reference must carry a real HubAccessConfig. Use a truthy check
+    # (not `is not None`): the field's unset default is the Unassigned() sentinel,
+    # so `is not None` would wrongly pass if the plumbing regressed and left it
+    # unset. Unassigned() is falsy, a real HubAccessConfig is truthy.
+    assert hub_access_config
+    assert hub_access_config.hub_content_arn
+
+
+def _assert_gated_reference_channels(model_trainer):
+    """Assert accept_eula flowed into the model channel's ModelAccessConfig.
+
+    Verified against defaults.py get_model_artifact_input: the resolved "model"
+    channel always carries
+    data_source.s3_data_source.model_access_config = ModelAccessConfig(
+        accept_eula=jumpstart_config.accept_eula). Using a gated model_id is what
+    makes accept_eula=True meaningful (a gated model is unusable without it); the
+    assertion itself is the same plumbing every JumpStart model uses.
+
+    Asserted before .train() so the gated ModelAccessConfig plumbing is pinned
+    even if the training job itself later fails for an unrelated capacity/quota
+    reason.
+    """
+    assert model_trainer.training_image
+    model_channels = [
+        c for c in model_trainer.input_data_config if getattr(c, "channel_name", None) == "model"
+    ]
+    assert len(model_channels) == 1
+    model_access_config = model_channels[0].data_source.s3_data_source.model_access_config
+    # Truthy check rather than `is not None`: the unset default is Unassigned()
+    # (falsy), which `is not None` would let through; a real ModelAccessConfig is
+    # truthy.
+    assert model_access_config, "gated reference resolved without a ModelAccessConfig"
+    assert model_access_config.accept_eula is True
+
+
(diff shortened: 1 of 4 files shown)


sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -10,14 +10,145 @@
...
+
+def _public_model_arn(region, model_id):
+    return f"arn:aws:sagemaker:{region}:aws:hub-content/" f"SageMakerPublicHub/Model/{model_id}"
+
+
+def _wait_for_content(sm, hub_name, name, content_type, timeout=300, poll=10):
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        try:
+            resp = sm.list_hub_contents(HubName=hub_name, HubContentType=content_type)
```
Narrohag: can we use describe_hub_content directly?

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -10,14 +10,145 @@
...
+                logger.warning("Failed to delete hub content %s: %s", c, e)
+    try:
+        sm.delete_hub(HubName=hub_name)
+    except ClientError as e:
+        logger.warning("Failed to delete hub %s: %s", hub_name, e)
+
+
+def _default_training_dataset(region, model_id):
+    """Resolve the model's default training dataset S3 URI from JS metadata."""
+    from sagemaker.core.jumpstart.accessors import JumpStartModelsAccessor
```
Narrohag: let's avoid the lazy imports, claude loves to add them for some reason lol

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -54,3 +185,125 @@ def test_jumpstart_train(test_case):
...
+        accept_eula=True,
+    )
+    model_trainer = ModelTrainer.from_jumpstart_config(
+        jumpstart,
+        role=_execution_role(sagemaker_session),
+        base_job_name="sdk-integ-train-alias",
+        compute=Compute(instance_type="ml.m5.xlarge"),
+        sagemaker_session=sagemaker_session,
+    )
+    model_trainer.train(input_data_config=[InputData(channel_name="training", data_source=dataset)])
```
Narrohag: is there anything to assert or just validating nothing is thrown?

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -54,3 +185,125 @@ def test_jumpstart_train(test_case):
         compute=test_case.get("compute"),
     )
     model_trainer.train()
+
+
+def test_jumpstart_train_from_private_hub_reference(private_hub, sagemaker_session):
```
Narrohag, line 276: can you add a unit test for training a model reference to a gated model? there should have been a similar test in v2 so you can use that same model. There's some ModelAccessConfig/accept_eula stuff that we should verify works

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -10,14 +10,145 @@
...
+    """Create a temporary private hub; tear it (and its contents) down after."""
+    sm = _sm_client(sagemaker_session)
+    hub_name = f"{HUB_NAME_PREFIX}-{uuid.uuid4().hex[:8]}"
+    try:
+        sm.create_hub(
+            HubName=hub_name,
+            HubDescription="SDK integ test JumpStart training private hub",
+        )
+    except ClientError as e:
+        pytest.skip(f"Cannot create private hub (missing permissions?): {e}")
```
sahilper: Skipping gracefully without hub permissions is the right intent, but the implementation is wider than that: every setup step here is a bare `except ClientError` → skip (`create_hub`, `create_hub_content_reference`, `import_hub_content`, `describe_hub_content`, and both `_wait_for_content` timeouts).

The assertion itself fails loudly — `from_jumpstart_config` isn't wrapped. But if `import_hub_content` for a private-hub `Model` breaks service-side, `test_jumpstart_train_from_private_owned_model` skips and CI stays green, so the test for the core case in this fix silently stops running.

Suggest matching only the specific authorization codes you expect in a restricted account and letting everything else fail; `_wait_for_content` returning `False` should be a `pytest.fail` once the hub was creatable.

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
```diff
@@ -54,3 +185,125 @@ def test_jumpstart_train(test_case):
...
+
+    jumpstart = JumpStartConfig(model_id=TRAINABLE_MODEL_ID, hub_name=private_hub, accept_eula=True)
+    model_trainer = ModelTrainer.from_jumpstart_config(
+        jumpstart,
+        role=_execution_role(sagemaker_session),
+        base_job_name="sdk-integ-train-ref",
+        compute=Compute(instance_type="ml.m5.xlarge"),
+        sagemaker_session=sagemaker_session,
+    )
+    model_trainer.train(input_data_config=[InputData(channel_name="training", data_source=dataset)])
```
sahilper: This explicit `training` channel bypasses the hub-aware channel construction the fix enables. `_create_training_job_args` merges with "method parameter taking precedence" (`existing_channels[new_input.channel_name] = new_input`), so this bare `InputData` replaces the channel built by `JumpStartTrainDefaults.get_training_dataset_input` — the only place `ModelAccessConfig` and `HubAccessConfig(hub_content_arn=...)` are attached (`defaults.py`, `hub_content_type == "ModelReference"` branch).

So the reference and alias tests never assert that a `HubAccessConfig` was derived from the resolved hub content, and with a non-gated model like `catboost-regression-model` they'd pass even if that plumbing were wrong.

The existing `test_jumpstart_train` calls `train()` with no arguments and relies on SDK-resolved channels — suggest the same here and dropping `_default_training_dataset`, which also removes the hardcoded `jumpstart-cache-prod-{region}` bucket.

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
tanvikab4: switched _wait_for_content to use describe_hub_content directly

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
tanvikab4: This test's purpose is to verify the aliased-reference resolution path so when hub_content_name differs from the public model_id, from_jumpstart_config resolves the reference by its alias. I added these assertions to strengthen the test:
  1. _jumpstart_config.hub_content_name == ALIASED_REFERENCE_NAME (the alias was threaded through resolution)
  2. training_image is set 
  3. the model channel's S3 source carries a HubAccessConfig with a hub_content_arn

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
tanvikab4: Added a unit test class TestJumpStartTrainDefaultsGatedModelReferenceEula in test_defaults.py using the same gated model as v2 (mocks the resolver seam and verifies the
  ModelAccessConfig/accept_eula behavior)

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
tanvikab4: Added a _skip_if_unauthorized helper that skips only on an authorization allowlist (AccessDeniedException/AccessForbiddenException/UnauthorizedOperation) and re-raises everything else

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
tanvikab4: Dropped explicit channel and _default_training_dataset. Train() now uses SDK-resolved channels like test_jumpstart_train. Also added assertions on the resolved channels so the plumbing is guarded

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
Narrohag: Sorry, meant integ test. My bad

sahilper (approved):

sagemaker-train/tests/integ/jumpstart/test_jumpstart_train.py
Narrohag: might have missed it but do we have this as an integ test?

tanvikab4: I looked into the v2 tests and there exists a gated private hub test on the ModelBuilder/deploy side, but not one for the training path. I'll add a corresponding integ test for ModelTrainer rn

Narrohag (approved):

rsareddy0329 (approved):

PR #6217, opened 2026-08-27
evaluator_model was checked against a hardcoded _ALLOWED_EVALUATOR_MO…

mohamedzeidan2021:
…DELS

LLMAsJudgeEvaluator.evaluator_model was validated against a hardcoded _ALLOWED_EVALUATOR_MODELS dict (model → regions) in sagemaker/train/constants.py. This PR replaces that with two-step validation against authoritative, service-maintained sources, and removes the dict.

Problem

1. Maintenance toil. Every Bedrock judge-model add/deprecation required hand-editing the dict and cutting an SDK release — and the same list is triplicated across the SDK, Studio UI, and SageMaker Agent Skills.
2. Stale list → deep runtime failures. When a judge model reaches end of life the dict still lists it, so it passes client-side validation, the eval job spins up, and only the in-container Bedrock CreateEvaluationJob call fails ("model version has reached end of life"). Compute is wasted and the error surfaces deep inside a running job instead of failing fast.

This is real today: the supported list still advertises claude-3-5-sonnet-20240620, claude-3-5-haiku, claude-3-5-sonnet-v2, and claude-3-7-sonnet, all of which return ResourceNotFoundException from Bedrock in us-west-2.

Solution — two-step validation

Step 1 — construction (is it a judge-capable model?): fetch the service-maintained list at s3://jumpstart-cache-prod-<region>/fmhMetadata/supported-llmaj-judge-models.json (source of truth for supported judge models) and fail fast if evaluator_model isn't in it.

Step 2 — evaluate() (is it still in service?): the supported list is a superset that can still include end-of-life models, so we call bedrock:GetFoundationModel and fail fast if the model is unavailable in the region or past its endOfLifeTime. The lookup is gated on the caller's IAM permission via a new non-raising caller_can_perform() helper that reuses the existing iam:SimulatePrincipalPolicy caller-check pattern (verify_evaluation_caller_permissions).

Graceful degradation everywhere: we never block on "can't tell." If a source can't be read (missing permission, unreadable list, transient error), the SDK logs an actionable warning with a link to the supported models and continues.

User-facing messaging

Fail-fast errors (block construction / submission):

- Not a supported judge model (Step 1):
▎ evaluator_model '<model>' is not a supported LLM-as-Judge model in region '<region>'. Choose one of the supported judge models. See https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
- Past end-of-life (Step 2):
▎ evaluator_model '<model>' has reached end of life in region '<region>' (end-of-life <timestamp>) and can no longer be used as a judge. Choose a judge model that is in service. See https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
- Unavailable in region (Step 2):
▎ evaluator_model '<model>' is not available in region '<region>'. It may be unsupported in this region or have reached end of life. Choose a judge model that is in service in this region. See https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported

Warnings (degrade, do not block):

- Caller's role is missing the permission (Step 2):
▎ Your IAM role does not include the bedrock:GetFoundationModel permission, so the SDK can't check whether the evaluator model '<model>' is still in service or has reached end of life. The evaluation will still run, but it may fail if this model has been retired. Add bedrock:GetFoundationModel to your role to enable this check. See the list of supported judge models: https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
- Permission can't be confirmed — identity isn't a role, or iam:SimulatePrincipalPolicy unavailable (Step 2):
▎ The SDK couldn't confirm your IAM role includes the bedrock:GetFoundationModel permission, so it can't check whether the evaluator model '<model>' is still in service or has reached end of life. The evaluation will still run, but it may fail if this model has been retired. See the list of supported judge models: https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
- Transient Bedrock error while checking lifecycle (Step 2):
▎ The SDK couldn't verify whether the evaluator model '<model>' is still in service right now (a temporary error occurred). The evaluation will still run, but it may fail if this model has been retired. See the list of supported judge models: https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported
- Supported-models list unreadable (Step 1):
▎ The SDK couldn't retrieve the list of supported judge models, so it can't confirm '<model>' is a valid judge model. The evaluation will still run, but it may fail if the model isn't supported. See the list of supported judge models: https://docs.aws.amazon.com/bedrock/latest/userguide/evaluation-judge.html#evaluation-judge-supported

### DIFF
--- a/sagemaker-core/src/sagemaker/core/helper/iam_role_resolver.py
+++ b/sagemaker-core/src/sagemaker/core/helper/iam_role_resolver.py
@@ -758,59 +758,6 @@ def verify_evaluation_caller_permissions(
     return True
 
 
-def caller_can_perform(
-    actions: List[str], sagemaker_session=None
-) -> Optional[bool]:
-    """Return whether the caller identity is allowed to perform ALL of ``actions``.
-
-    Read-only, non-raising sibling of :func:`verify_evaluation_caller_permissions`:
-    it resolves the caller's backing IAM role and simulates ``actions`` against it
-    with ``iam:SimulatePrincipalPolicy``. Use it to gate an optional client-side
-    AWS call on whether the caller is actually permitted to make it, degrading
-    gracefully (rather than raising) when that cannot be determined.
-
-    ``actions`` should be account-level / wildcard-resource actions — they are
-    simulated without ``ResourceArns``, so resource-scoped actions can come back
-    ``implicitDeny`` even for a caller who holds them.
-
-    Args:
-        actions: IAM action names to check (e.g. ``["bedrock:GetFoundationModel"]``).
-        sagemaker_session: SageMaker session (used to get the boto session).
-
-    Returns:
-        True  — every action is allowed.
-        False — at least one action is denied.
-        None  — could not be determined (caller is not a role, cannot call
-                ``sts:GetCallerIdentity``, or lacks ``iam:SimulatePrincipalPolicy``).
-    """
-    boto_session = _get_boto_session(sagemaker_session)
-    sts_client = boto_session.client("sts")
-    iam_client = boto_session.client("iam")
-
-    try:
-        caller_identity = sts_client.get_caller_identity()
-    except ClientError:
-        return None
-
-    caller_arn = caller_identity["Arn"]
-    account_id = caller_identity["Account"]
-    partition = _partition_from_arn(caller_arn)
-
-    caller_role_arn = _resolve_caller_role_arn(iam_client, caller_arn, account_id, partition)
-    if not caller_role_arn:
-        return None
-
-    try:
-        denied = _simulate_denied_actions(iam_client, caller_role_arn, list(actions))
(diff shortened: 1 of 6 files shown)


jam-jee:
Overall changes LGTM , thanks.
- Can we pls also add integ tests (positive and negative) for validating a model against JS hub. (No need to add separate integ tests for E2E eval jobs.)
- Positive and negative integ tests for bedrock role.

papriwal (commented):
LGTM, thanks!

papriwal (approved):

jam-jee (approved):

PR #6224, opened 2026-08-28
feat(pipeline): Add inference and lineage step types

Rishabh0255:
Add 4 pipeline step classes:

    EndpointConfigStep, EndpointStep (SageMaker inference deployment)
    InferenceComponentStep (multi-model endpoint support)
    LineageStep (ML governance tracking)

Design: each step takes step_args captured under a PipelineSession, following the convention used by TrainingStep and ModelStep. Session.endpoint_from_production_variants, Session.create_endpoint and Session.create_inference_component route through _intercept_create_request -- a pass-through under a plain Session, capturing the request without a service call under a PipelineSession. LineageStep records one entity per step, with step_args from Action/Artifact/Context/Association.create(); an association can reference an entity created by another step via Steps.<name>.ActionArns['<entity>']. Each step validates step_args provenance with validate_step_args_input, rejecting a wrong producer or a raw dict.

Chaining note: the resource created by an EndpointConfig step carries an execution-unique name suffix, so a downstream step must reference it as config_step.properties.EndpointConfigName rather than the requested name. The Endpoint step does not suffix.

Retryability: only EndpointConfigStep is retryable. Cacheability: EndpointConfigStep and EndpointStep are structurally cacheable via cache_config.

Includes 29 unit tests and two integration tests covering all four step types end to end against the service (2 passed in 473.64s).

Supersedes #6140 -- same feature, reworked to step_args per review feedback.

Issue #, if available:

Description of changes:


sagemaker-mlops/src/sagemaker/mlops/workflow/_argument_validation.py
```diff
@@ -0,0 +1,124 @@
...
+                "Installed botocore does not know %s.%s; skipping "
+                "client-side argument shape validation for this step.",
+                service_name,
+                operation_name,
+            )
+            _SHAPE_CACHE[cache_key] = None
+    return _SHAPE_CACHE[cache_key]
+
+
+def validate_step_arguments(
```
rohangujarathi: why do we need all of this additional validation here? Other steps do not have this explicit validation. How are these steps different from other steps?

sagemaker-mlops/src/sagemaker/mlops/workflow/endpoint_step.py
```diff
@@ -0,0 +1,223 @@
...
+    ``ExplainerConfig``, which are rejected by the pipeline service.
+
+    Per the pipeline service's step contract, ``EndpointConfig`` is structurally
+    cacheable (``cache_config``) and retryable (``retry_policies``).
+    """
+
+    def __init__(
+        self,
+        name: str,
+        arguments: Dict[str, Any],
```
rohangujarathi: This is not the right implementation for any of these steps. It will be very difficult to construct these arguments manually. We need to use the existing pysdk constructs and pass them as arguments. Please see how Training/Model steps are implemented and follow that pattern here. You must use step_args from PipelineSession instead of raw arguments: dict

SDK primitives exists for all four steps, and it eliminates the entire _argument_validation.py machinery

sagemaker-mlops/tests/integ/workflow/test_lineage_step.py
```diff
@@ -0,0 +1,145 @@
...
+def pipeline_session():
+    return PipelineSession()
+
+
+@pytest.fixture
+def role():
+    return get_execution_role()
+
+
+def test_lineage_step_execute_end_to_end(sagemaker_session, pipeline_session, role):
```
rohangujarathi, line 59: please add integ tests for other steps as well

sagemaker-mlops/src/sagemaker/mlops/workflow/lineage_step.py
```diff
@@ -0,0 +1,166 @@
...
+            root.__dict__[field] = _EntityArnMap(step_name=name, path=field)
+        root.__dict__["Associations"] = Properties(step_name=name, path="Associations")
+        self._properties = root
+
+    @property
+    def arguments(self) -> RequestType:
+        """The ``Arguments`` block describing the lineage entity to create."""
+        key = _CALLER_TO_ARGUMENTS_KEY[self.step_args.caller_name]
+        entity = self.step_args.args
+        if key == "Associations":
```
rohangujarathi, line 152: Cross-step Source.Arn = {"Get": "Steps.X.ActionArns[...]"} won't resolve. Service only accepts a literal/param ARN or an intra-step Name+Type lookup (populated only from entities created in the same step). The service models a batch of entities+associations per step; this does one entity per step and chains across steps, which the runtime doesn't support. Reference siblings by Name+Type within one step, and let one step carry multiple entities.

sagemaker-core/src/sagemaker/core/helper/session_helper.py
```diff
@@ -1439,6 +1464,19 @@ def _intercept_create_request(
         """
         return create(request)
 
+    def _is_pipeline_context(self) -> bool:
```
rohangujarathi, line 1467: create_model doesn't do this. It routes through _intercept_create_request and returns the name; the pipeline branch lives in PipelineSession. Can you keep base Session pipeline-agnostic and drop this helper?

sagemaker-mlops/tests/integ/workflow/test_lineage_step.py
```diff
@@ -0,0 +1,137 @@
...
```
rohangujarathi, line 59: Only covers a single Action. Please test with all the entities

sagemaker-mlops/src/sagemaker/mlops/workflow/endpoint_step.py
```diff
@@ -0,0 +1,203 @@
...
+            name=name,
+            step_type=StepTypeEnum.ENDPOINT_CONFIG,
+            display_name=display_name,
+            description=description,
+            depends_on=depends_on,
+            retry_policies=retry_policies,
+        )
+        validate_step_args_input(
+            step_args=step_args,
+            expected_caller={"endpoint_from_production_variants"},
```
rohangujarathi, line 103: can you check if this is the only helper class that creates endpointConfig? also this one seems to be creating both endpoint config as well are endpoint. are we handling this?

rohangujarathi (commented):
Inference steps look right and validation follows the existing pattern. Two blockers:
(1) capture is done three different ways — please route all four steps through PipelineSession._intercept_create_request (the seam Model.create/fit/run already use) instead of adding _is_pipeline_context() in session_helper.py and the _invoke_api hook. 
(2) LineageStep bug needs to be fixed

sagemaker-mlops/src/sagemaker/mlops/workflow/_argument_validation.py
Rishabh0255: Removed. It was only there because the steps took raw dicts.

sagemaker-mlops/src/sagemaker/mlops/workflow/endpoint_step.py
Rishabh0255: Done, all four steps take step_args now and validate the producer, same as
TrainingStep. That removed _argument_validation.py entirely.

sagemaker-mlops/tests/integ/workflow/test_lineage_step.py
Rishabh0255: Added tests/integ/workflow/test_deployment_steps.py, one pipeline chaining EndpointConfig → Endpoint → InferenceComponent, with cleanup.

sagemaker-mlops/src/sagemaker/mlops/workflow/lineage_step.py
Rishabh0255: Fixed. A step now creates multiple entities and associations reference them by name and type, or by ARN for anything created elsewhere.

Associations are declared rather than captured: Association.create() only takes ARNs, and a sibling has no ARN until the step runs. A reference to an entity the step doesn't create is rejected up front.

sagemaker-core/src/sagemaker/core/helper/session_helper.py
Rishabh0255: Dropped. PipelineSession._intercept_create_request returns the captured context now so producers just return its result and base Session has no pipeline branch.

sagemaker-mlops/tests/integ/workflow/test_lineage_step.py
Rishabh0255: Now creates an Action, an Artifact and a Context in one step with associations between them, and asserts an ARN for each.

sagemaker-mlops/src/sagemaker/mlops/workflow/endpoint_step.py
Rishabh0255: No, EndpointConfig.create() in resources.py also does and it creates only the config,but it takes a boto3 session and calls the client directly so it can't be a producer without a codegen change.

endpoint_from_production_variants creates both so I added Session.create_endpoint_config which creates only the config and pointed the step at that.

sagemaker-mlops/src/sagemaker/mlops/workflow/lineage_step.py
```diff
@@ -0,0 +1,332 @@
...
+    consume. Note these cannot be used as an association endpoint in another
+    ``LineageStep``: the service resolves name and type only against the
+    entities created by the same step.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        step_args: Optional[Union[_JobStepArguments, List[_JobStepArguments]]] = None,
+        associations: Optional[List[LineageAssociation]] = None,
```
rohangujarathi, line 211: Let's define this step in the following way

```
LineageStep(
    name="lineage_step_name",
    actions=[action1_args, action2_args],       # captured from Action.create()
    artifacts=[artifact_args],                  # captured from Artifact.create()
    contexts=[context_args],                    # captured from Context.create()
    associations=[LineageAssociation(...)],
    ....                                        # rest of parameters
)
```

Rishabh0255: Done, actions artifacts and contexts are separate arguments now. Each is validated
against its own producer so a misplaced entity fails right away.

rohangujarathi (dismissed):

jam-jee (approved):

PR #6235, opened 2026-09-02
fix(train): complete PipelineSession support for V3 trainers (SFT/DPO…

### LINKED ISSUE #6163: does SageMaker Pipeline in SageMaker Python SDK v3 support fine-tuning (such as SFTTrainer, DPOTrainer, RLAIFTrainer)?
**PySDK Version**
- [ ] PySDK V2 (2.x)
- [X] PySDK V3 (3.x)

**Describe the bug**
Hi team, I have a quick question, does SageMaker Pipeline in SageMaker Python SDK v3 support fine-tuning (such as SFTTrainer, DPOTrainer, RLAIFTrainer)?

I have this kind of codes. I created SFTTrainer() with PipelineSession in the sagemaker_session. However, the train() is not defered, but automatially launches training job for fine-tuning.

**To reproduce**
```
session = PipelineSession()

sft_trainer = SFTTrainer(
    model="huggingface-reasoning-qwen3-1-7b",
    training_type=TrainingType.LORA,
    training_dataset=f"{TRIAGE_DATA_PREFIX}/anycompany_triage_sft.jsonl",
    model_package_group="anycompany-triage-agent", 
    s3_output_path=f"{PIPELINE_OUTPUT_PREFIX}/sft-output",
    sagemaker_session=session,
)

step_sft_args = sft_trainer.train()   # Expect to be deferred inside pipeline-definition context, but automatically run

step_sft = TrainingStep(
    name="SFTTuneTriage",
    step_args=step_sft_args,
)
...
```

**Expected behavior**
The function is not automatically run, but defined as a pipeline in SageMaker Pipeline. 

**System information**
- **SageMaker Python SDK version**: 3.17.0

nayan3107:
…/RLAIF/RLVR)

Add @runnable_by_pipeline decorator and PascalCase request serialization so TrainingStep can consume step_args from V3 fine-tuning trainers.

Changes:
- Add @runnable_by_pipeline decorator on train() for all 4 trainers
- Build PascalCase serialized request (remove session/region, pop job name)
- Fix Tags to PascalCase (JumpStart returns lowercase key/value)
- Add source_code=None to BaseTrainer (required by get_code_hash)
- Add consumer tests (TrainingStep.arguments validation)
- Add regular session regression tests

Fixes: https://github.com/aws/sagemaker-python-sdk/issues/6163

*Issue #, if available:*

*Description of changes:*
## fix(train): complete PipelineSession support for V3 trainers

Follow-up to #6213. Completes PipelineSession integration so V3 fine-tuning 
trainers (SFTTrainer, DPOTrainer, RLAIFTrainer, RLVRTrainer) work with 
TrainingStep in SageMaker Pipelines.

Fixes https://github.com/aws/sagemaker-python-sdk/issues/6163

### Changes

- **`@runnable_by_pipeline` decorator** on `train()` for all 4 trainers — matches 
  ModelTrainer/Processor/Transformer pattern. Decorator captures function reference 
  + trainer instance so TrainingStep can re-execute during pipeline compilation.

- **PascalCase request serialization** inside `train()` body — removes non-serializable 
  `session`/`region`, pops `training_job_name` (engine generates at runtime), converts 
  snake_case keys to PascalCase via `to_pascal_case()` + `serialize()`.

- **Tags PascalCase fix** — `_get_jumpstart_tags()` returns lowercase `key`/`value` 
  dicts (not Tag objects), so `_serialize_dict()` doesn't auto-convert them. Manually 
  converts to `Key`/`Value` before serialization.

- **`source_code = None` in BaseTrainer** — `get_code_hash()` in pipeline compilation 
  accesses `model_trainer.source_code` directly. ModelTrainer defines this field; 
  BaseTrainer didn't. Fine-tuning trainers use built-in containers with no customer code.

### Testing

**Unit tests (12 new, all existing pass):**
- Producer tests: `train()` returns `_StepArguments` with func + trainer captured
- Consumer tests: `TrainingStep(step_args=...).arguments` produces valid PascalCase dict 
  with no session/region leaked, Tags in correct `Key`/`Value` format
- Regular session tests: `TrainingJob.create()` still called normally (no regression)

**Manual E2E (real AWS, us-west-2, account 243179453435):**
```python
session = PipelineSession()
trainer = SFTTrainer(model='meta-textgeneration-llama-3-2-1b-instruct', ...)
result = trainer.train()                    # _StepArguments (no job)
step = TrainingStep(name="SFT", step_args=result)
step.arguments                              # PascalCase dict ✅
pipeline = Pipeline(steps=[step], sagemaker_session=session)
pipeline.upsert(role_arn="...")             # CreatePipeline ✅
execution = pipeline.start()               # Execution started ✅
```

Pipeline created and execution `rg29qr88isq5` started successfully.


sagemaker-mlops/tests/integ/workflow/test_v3_trainer_pipeline.py
```diff
@@ -0,0 +1,399 @@
...
+from sagemaker.train.rlvr_trainer import RLVRTrainer
+from sagemaker.train.common import TrainingType
+from sagemaker.core.workflow.pipeline_context import PipelineSession, _StepArguments
+from sagemaker.core.helper.session_helper import Session, get_execution_role
+from sagemaker.mlops.workflow.steps import TrainingStep
+from sagemaker.mlops.workflow.pipeline import Pipeline
+
+MODEL_ID = "meta-textgeneration-llama-3-2-1b-instruct"
+
+_BASE_SFT_SAMPLES = [
```
rohangujarathi: nit: we could reuse some existing dataset or move this to a file

rohangujarathi (dismissed):

sagemaker-mlops/tests/integ/workflow/test_v3_trainer_pipeline.py
```diff
@@ -0,0 +1,399 @@
...
+def model_package_group(sagemaker_session):
+    """Create a temporary model package group and delete it after the test."""
+    name = f"integ-test-v3-trainer-{uuid.uuid4().hex[:8]}"
+    sagemaker_session.sagemaker_client.create_model_package_group(
+        ModelPackageGroupName=name,
+        ModelPackageGroupDescription="Integ test for V3 trainer pipeline",
+    )
+    yield name
+    try:
+        sagemaker_session.sagemaker_client.delete_model_package_group(
```
papriwal: Minor: since the `*_create_and_execute` tests register a model package into `model_package_group`, `delete_model_package_group` here may fail on a non-empty group (and the `except Exception: pass` would quietly hide it), potentially leaving the group and its model packages behind. Might be worth deleting the member model packages before the group so nothing accumulates in the CI account over time. Non-blocking.

papriwal (commented):
One small heads-up on the new integ test's teardown (`test_v3_trainer_pipeline.py`). Nothing blocking; the production PipelineSession change looks good.

rohangujarathi (dismissed):

nayan3107:
Fixed -
- Tags conversion for Tag objects. The old code assumed all tag entries were dicts and called t.get(...), which crashes on the pydantic Tag objects the public type signature (tags: Optional[List[Tag]]) invites. Now normalizes both forms in all four trainers (sft/dpo/rlaif/rlvr_trainer.py):
python

```
  {"Key": t.get("key", t.get("Key")), "Value": t.get("value", t.get("Value"))}
  if isinstance(t, dict)
  else {"Key": t.key, "Value": t.value}
```
  

- Missing trailing newline in sft_trainer.py (flake8 W292).
- Unit test coverage for Tag objects. New test test_train_pipeline_session_normalizes_tag_objects in test_sft_trainer.py — mixes a JumpStart dict tag with user-supplied Tag(key="env", value="prod") and asserts all three end up in [{Key, Value}] form. This test fails on the old code with AttributeError: 'Tag' object has no attribute 'get'.
- Model package group cleanup (@papriwal). Fixture now paginates list_model_packages, deletes each member package, then the group. Cleanup errors are printed instead of silently swallowed, so real leaks show up in CI output.
Sample training data moved to files (@rohangujarathi). Two 20-row files under sagemaker-mlops/tests/integ/data/v3_trainer/ (sft_train.jsonl, preference_train.jsonl); loaded and repeated 7× in-memory at fixture time to exceed the 128 batch size used by DPO/RLAIF/RLVR.

sagemaker-train/src/sagemaker/train/base_trainer.py
```diff
@@ -160,6 +160,7 @@ def __init__(
         self.base_model_name = base_model_name
         self.disable_output_compression = disable_output_compression
         self.notification_rule_arn = None
+        self.source_code = None
```
nargokul, line 163: Where is this used ?

sagemaker-train/src/sagemaker/train/rlvr_trainer.py
```diff
@@ -561,8 +564,23 @@ def train(self, training_dataset: Optional[Union[str, DataSet]] = None,
...
+            if "Tags" in pipeline_request and pipeline_request["Tags"]:
+                pipeline_request["Tags"] = [
+                    {"Key": t.get("key", t.get("Key")), "Value": t.get("value", t.get("Value"))}
+                    if isinstance(t, dict)
+                    else {"Key": t.key, "Value": t.value}
+                    for t in pipeline_request["Tags"]
+                ]
+            serialized_request = serialize(pipeline_request)
+            sagemaker_session._intercept_create_request(serialized_request, None, "train")
+            return
```
nargokul, line 583: Why are we changing the method signature here ?
Can this still return the sagemaker_session.context ?

nargokul (changes requested):

sagemaker-train/src/sagemaker/train/base_trainer.py
nayan3107: source_code = None in BaseTrainer — get_code_hash() in pipeline compilation
accesses model_trainer.source_code directly. ModelTrainer defines this field;
BaseTrainer didn't. Fine-tuning trainers use built-in containers with no customer code.

sagemaker-train/src/sagemaker/train/rlvr_trainer.py
nayan3107: Bare return matches every peer @runnable_by_pipeline method — ModelTrainer.train, Transformer.transform, Processor.process, Tuner.fit all end the PipelineSession branch with return (not return ...context).

Reason: with @runnable_by_pipeline, the customer's trainer.train() return value comes from the decorator, not the body. What the body returns only matters to execute_job_functions, which recurses if it's a _StepArguments:

python

> chained_args = step_args.func(*step_args.func_args, **step_args.func_kwargs)
> if isinstance(chained_args, _StepArguments):
>     execute_job_functions(chained_args)
> 

So returning sagemaker_session.context here would crash  after _intercept_create_request(..., "train") it's a _JobStepArguments (subclass of _StepArguments) with func=None, which triggers the isinstance branch → tries to call None(...) → TypeError. That recursion is for legitimate chained processors (PySpark → Processor) whose returns carry a real func.

The context still ends up in the pipeline definition, TrainingStep.arguments reads it via sagemaker_session.context.args after replaying the body, so no info is lost.

Context: 
https://github.com/aws/sagemaker-python-sdk/pull/6213
 had no decorator, so return sagemaker_session.context handed a _JobStepArguments (func=None) directly to the caller and TrainingStep(step_args=…) crashed at pipeline compile, one of the two bugs flagged in that review.

nargokul: runnable_by_pipeline also ends up calling the run_func in this case the train . 

This is a breaking change essentially and is not something the SDK can support with a minor version bump .

sagemaker-train/src/sagemaker/train/base_trainer.py
nargokul: So we would need to update get_code_hash with handling source_code for base trainer as well right as part of this PR ?

nayan3107: Its already handled.

sagemaker-train/src/sagemaker/train/rlvr_trainer.py
nayan3107: The decorator has an explicit if run_func.__name__ in ["register", "build"]: branch (sagemaker-core/.../workflow/pipeline_context.py:336) — that branch calls run_func immediately at customer-call time. For everything else, including train, the code path is different:

```
return _StepArguments(retrieve_caller_name(self_instance), run_func, *args, **kwargs)
```
That returns an envelope, does not invoke run_func. run_func for train gets invoked exactly once, later, by execute_job_functions when TrainingStep.arguments is called.

rohangujarathi (approved):

rsareddy0329 (approved):

nargokul (approved):

PR #6242, opened 2026-09-08
fix(jumpstart): add eusc-de-east-1 buckets and name JumpStart in the bucket error

evakravi:
## Problem

In `eusc-de-east-1`, the sample notebook cell `JumpStartModel(model_id="deepseek-llm-r1-distill-llama-8b", model_version="*", hub_name="SageMakerPublicHub")` fails with `ValueError: Unable to get content bucket for Neo in eusc-de-east-1 region.` The model does not use Neo.

The v2 region configuration has no `eusc-de-east-1` entry, so `get_jumpstart_content_bucket` raises for the region. Its error names Neo, a copy of the Neo accessor message. v3 got the region entry in #5615.

## Solution

Add the public and gated JumpStart buckets for `eusc-de-east-1` to the region configuration, with no Neo bucket. Name JumpStart in the `get_jumpstart_content_bucket` error. #6243 corrects the same error text in v3.

## Tests

Unit tests read the `eusc-de-east-1` public and gated buckets and match the JumpStart error text. The three tests fail on `origin/master-v2` and pass on this branch.

```text
SAGEMAKER_SUPPRESS_V2_WARNING=1 PYTHONPATH=src python -m pytest -s -vv tests/unit/sagemaker/jumpstart/test_utils.py::TestBucketUtils
```

Result: `12 passed in 1.93s`.

`integ-tests` on `master-v2` fails without this change. SageMaker rejects the `ml.t2.medium` fixture with `ValidationException: Deprecated instance type "T2"` since 2026-09-03, `test_jumpstart_gated_model` gets `InsufficientInstanceCapacity`, and the `build_jar` teardown deletes `hello-spark-java.jar` under xdist. Those tests are outside this change. The `codestyle-doc-tests` findings are in unchanged files.

## Merge Checklist

- [x] I read the contribution guide.
- [x] The change is backward compatible.
- [x] The commits use the repository commit-message format.
- [x] The change creates no S3 or STS client.
- [x] The change has regression coverage and adds no dependency.

### DIFF
--- a/src/sagemaker/jumpstart/utils.py
+++ b/src/sagemaker/jumpstart/utils.py
@@ -165,7 +165,7 @@ def get_jumpstart_content_bucket(
         except KeyError:
             formatted_launched_regions_str = get_jumpstart_launched_regions_message()
             raise ValueError(
-                f"Unable to get content bucket for Neo in {region} region. "
+                f"Unable to get content bucket for JumpStart in {region} region. "
                 f"{formatted_launched_regions_str}"
             )
 
--- a/src/sagemaker/jumpstart/region_config.json
+++ b/src/sagemaker/jumpstart/region_config.json
@@ -119,6 +119,10 @@
         "gated_content_bucket": "jumpstart-private-cache-prod-eu-west-3",
         "neo_content_bucket": "sagemaker-sd-models-prod-eu-west-3"
     },
+    "eusc-de-east-1": {
+        "content_bucket": "jumpstart-cache-prod-eusc-de-east-1",
+        "gated_content_bucket": "jumpstart-private-cache-prod-eusc-de-east-1"
+    },
     "il-central-1": {
         "content_bucket": "jumpstart-cache-prod-il-central-1",
         "gated_content_bucket": "jumpstart-private-cache-prod-il-central-1"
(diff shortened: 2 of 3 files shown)


src/sagemaker/jumpstart/utils.py
```diff
@@ -1631,18 +1631,15 @@ def _add_model_access_configs_to_model_data_sources(
     return acked_model_data_sources
 
 
-def get_draft_model_content_bucket(provider: Dict, region: str) -> str:
+def get_draft_model_content_bucket(provider: Dict, region: str) -> Optional[str]:
```
evakravi: why did we change this?

evakravi: The return type now matches the accessor behavior for a launched region without a Neo bucket.

The factory raises only when speculative-decoding metadata selects Neo and the accessor returns `None`.

JumpStart-backed metadata uses a JumpStart bucket. Models without this metadata do not call this function.

I added the contract documentation and live EU Sovereign Cloud integration coverage for the reported DeepSeek setup.

Narrohag (approved):

e-davidson (approved):

jam-jee (approved):

lucasjia-aws (approved):

PR #6246, opened 2026-09-09
feat(core): add InstancePreferences for multi-instance-type training and processing

papriwal:
> [!CAUTION]
> DO NOT MERGE

**Description**

Adds support for `InstancePreferences`, an ordered list of candidate instance types for training and processing jobs. When set, the service tries each candidate in list order and launches on the first type with available capacity, then reports the resolved choice back via `SelectedInstanceType` and `SelectedInstanceCount`.

This makes capacity-constrained jobs easier to schedule: instead of picking one instance type and retrying on `InsufficientCapacityError`, you supply a ranked list once.

**Training**

```python
from sagemaker.core.shapes import InstancePreference

compute = Compute(
    instance_preferences=[
        InstancePreference(instance_type="ml.p5.48xlarge", instance_count=2),
        InstancePreference(instance_type="ml.p4d.24xlarge", instance_count=4),
    ],
    volume_size_in_gb=100,
)
```

`instance_preferences` is mutually exclusive with `instance_type`, `instance_groups` and `instance_placement_config`. Each entry may carry its own `instance_count`, or you can leave them unset and use the uniform `ResourceConfig.InstanceCount` for whichever type wins. `training_plan_arns` is training-only and capped at one entry per preference.

**Processing**

`ProcessingClusterConfig` gains the same three members. `InstanceType` and `InstanceCount` are no longer required on that shape, since a preferences-only config supplies them per candidate — `VolumeSizeInGB` remains the only required member.

## What's in this change

| Area | Files | Notes |
|---|---|---|
| `sagemaker-core` src | 9 | training/processing configs and utils, Spark processor plumbing, plus regenerated `shapes.py`, `shape_dag.py` and the bundled service model |
| `sagemaker-core` tests | 5 | unit coverage for compute configs, processing and the service model; a new processing integ test |
| `sagemaker-train` | 4 | `model_trainer.py` validation, `defaults.py` handling, unit and integ tests |
| Docs | 2 | `docs/training/index.rst`, `docs/ml_ops/index.rst` |
| Examples | 2 | one training and one processing notebook |

Generated artifacts were produced by `sagemaker-core`'s own codegen from the bundled sample model, not hand-edited. New shapes: `InstancePreference`, `InstancePreferenceList`, `ProcessingInstancePreference`, `ProcessingInstancePreferenceList`, `TrainingPlanArnList`.

Unsupported combinations are rejected with a clear error rather than failing at the service: local mode, training recipes, and JumpStart training all reject `instance_preferences` up front.

## Notes for reviewers

**Requires botocore ≥ 1.43.90.** These shapes first ship in that release. On an older botocore the client rejects the new members with a `ParamValidationError`; only callers using this feature are affected. The dependency floor is bumped as part of the release, not in this PR.

**One unrelated model line moves.** The bundled `service-2.json` refresh also brings `MaxPendingTimeInSeconds.min` from `7200` to `1800`, matching botocore 1.43.90+. It shows up in the diff but is not part of this feature — the checked-in model was simply behind on that field.

**`IntPipeVar` on five members is not reproducible by codegen.** `shapes.py` widens `instance_count` on `InstancePreference`, `ProcessingInstancePreference` and `ProcessingClusterConfig`, plus `selected_instance_count` on `ResourceConfig` and `ProcessingClusterConfig`, to `IntPipeVar` so pipeline variables can be passed. `PIPE_VAR_OVERRIDES` in `tools/constants.py` is keyed by shape name and currently only covers `ResourceConfig`, so re-running codegen narrows these back to `Optional[int]`. Happy to add the missing override entries here or in a follow-up — worth deciding, since the periodic botocore sync regenerates this file. The two `selected_instance_count` fields are Describe-response only, so a pipeline variable arguably shouldn't apply to them at all.

## Testing

- Unit tests for training and processing config construction, serialization, and the mutually-exclusive validation paths
- A service-model test asserting the new shapes and the relaxed required list
- Integration tests for training and processing that assert the full round trip, including that the service echoes `SelectedInstanceType` / `SelectedInstanceCount`; they skip unless the required role, image and output-path environment variables are set
- Both notebooks run end to end

*Issue #, if available:*

*Description of changes:*


aviruthen (approved):

sagemaker-core/sample/sagemaker/2017-07-24/service-2.json
```diff
@@ -28787,6 +28787,33 @@
       },
       "documentation":"<p>Defines an instance group for heterogeneous cluster training. When requesting a training job using the <a href=\"https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html\">CreateTrainingJob</a> API, you can configure multiple instance groups .</p>"
     },
+    "InstancePreference":{
```
jam-jee, line 28790: not sure if we need to update service-2.json manually? It would be auto-updated based on boto spec?

deeppcs: This was just updated manually to enable local testing before Trebuchet release. This should be a no-op now that these fields are already released in the public spec?

deeppcs (approved):

baiyli (approved):

PR #6249, opened 2026-09-10
feat(core): add InstancePreferences for multi-instance-type training and processing

papriwal:
**Description**

Adds support for `InstancePreferences`, an ordered list of candidate instance types for training and processing jobs. When set, the service tries each candidate in list order and launches on the first type with available capacity, then reports the resolved choice back via `SelectedInstanceType` and `SelectedInstanceCount`.

This makes capacity-constrained jobs easier to schedule: instead of picking one instance type and retrying on `InsufficientCapacityError`, you supply a ranked list once.

**Training**

```python
from sagemaker.core.shapes import InstancePreference

compute = Compute(
    instance_preferences=[
        InstancePreference(instance_type="ml.p5.48xlarge", instance_count=2),
        InstancePreference(instance_type="ml.p4d.24xlarge", instance_count=4),
    ],
    volume_size_in_gb=100,
)
```

`instance_preferences` is mutually exclusive with `instance_type`, `instance_groups` and `instance_placement_config`. Each entry may carry its own `instance_count`, or you can leave them unset and use the uniform `ResourceConfig.InstanceCount` for whichever type wins. `training_plan_arns` is training-only and capped at one entry per preference.

**Processing**

`ProcessingClusterConfig` gains the same three members. `InstanceType` and `InstanceCount` are no longer required on that shape, since a preferences-only config supplies them per candidate — `VolumeSizeInGB` remains the only required member.

## What's in this change

| Area | Files | Notes |
|---|---|---|
| `sagemaker-core` src | 9 | training/processing configs and utils, Spark processor plumbing, plus regenerated `shapes.py`, `shape_dag.py` and the bundled service model |
| `sagemaker-core` tests | 5 | unit coverage for compute configs, processing and the service model; a new processing integ test |
| `sagemaker-train` | 4 | `model_trainer.py` validation, `defaults.py` handling, unit and integ tests |
| Docs | 2 | `docs/training/index.rst`, `docs/ml_ops/index.rst` |
| Examples | 2 | one training and one processing notebook |

Generated artifacts were produced by `sagemaker-core`'s own codegen from the bundled sample model, not hand-edited. New shapes: `InstancePreference`, `InstancePreferenceList`, `ProcessingInstancePreference`, `ProcessingInstancePreferenceList`, `TrainingPlanArnList`.

Unsupported combinations are rejected with a clear error rather than failing at the service: local mode, training recipes, and JumpStart training all reject `instance_preferences` up front.

## Notes for reviewers

**Requires botocore ≥ 1.43.90.** These shapes first ship in that release. On an older botocore the client rejects the new members with a `ParamValidationError`; only callers using this feature are affected. The dependency floor is bumped as part of the release, not in this PR.

**One unrelated model line moves.** The bundled `service-2.json` refresh also brings `MaxPendingTimeInSeconds.min` from `7200` to `1800`, matching botocore 1.43.90+. It shows up in the diff but is not part of this feature — the checked-in model was simply behind on that field.

**`IntPipeVar` on five members is not reproducible by codegen.** `shapes.py` widens `instance_count` on `InstancePreference`, `ProcessingInstancePreference` and `ProcessingClusterConfig`, plus `selected_instance_count` on `ResourceConfig` and `ProcessingClusterConfig`, to `IntPipeVar` so pipeline variables can be passed. `PIPE_VAR_OVERRIDES` in `tools/constants.py` is keyed by shape name and currently only covers `ResourceConfig`, so re-running codegen narrows these back to `Optional[int]`. Happy to add the missing override entries here or in a follow-up — worth deciding, since the periodic botocore sync regenerates this file. The two `selected_instance_count` fields are Describe-response only, so a pipeline variable arguably shouldn't apply to them at all.

## Testing

- Unit tests for training and processing config construction, serialization, and the mutually-exclusive validation paths
- A service-model test asserting the new shapes and the relaxed required list
- Integration tests for training and processing that assert the full round trip, including that the service echoes `SelectedInstanceType` / `SelectedInstanceCount`; they skip unless the required role, image and output-path environment variables are set
- Both notebooks run end to end

*Issue #, if available:*

*Description of changes:*


deeppcs (approved):

sagemaker-core/tests/integ/processing/test_instance_preferences_processing.py
```diff
@@ -0,0 +1,133 @@
...
+    t.strip()
+    for t in os.environ.get(
+        "PROCESSING_INSTANCE_PREFERENCES_TEST_INSTANCE_TYPES", "ml.m5.xlarge,ml.m5.large"
+    ).split(",")
+    if t.strip()
+]
+WAIT_TIMEOUT_SECONDS = 30 * 60
+POLL_SECONDS = 30
+
+pytestmark = pytest.mark.skipif(
```
papriwal: Could you confirm how the required role and image variables will be provided in the standard CI environment? As written, this module-level skipif causes test_processor_instance_preferences_e2e to be skipped whenever either variable is absent, so an integration run can be green without exercising the new processing path.

If this test requires dedicated resources, please coordinate with the maintainers to wire these variables into the canonical CI configuration while keeping concrete account-specific values out of the public repository. Alternatively, the test can derive suitable defaults from existing fixtures. Before merge, please provide CI output confirming that this test executed and passed rather than being skipped.

sagemaker-train/tests/integ/train/test_instance_preferences.py
```diff
@@ -0,0 +1,150 @@
...
+    t.strip()
+    for t in os.environ.get(
+        "INSTANCE_PREFERENCES_TEST_INSTANCE_TYPES", "ml.m5.xlarge,ml.m4.xlarge"
+    ).split(",")
+    if t.strip()
+]
+WAIT_TIMEOUT_SECONDS = 30 * 60
+POLL_SECONDS = 30
+
+pytestmark = pytest.mark.skipif(
```
papriwal: Could you also confirm how the required role, image, and S3 output variables will be supplied for this test in the standard CI environment? Without all three, the module-level marker skips test_model_trainer_instance_preferences_e2e, allowing the suite to pass without validating the new training path.

If these inputs depend on dedicated resources, please coordinate with the maintainers to configure them outside the public source tree. Otherwise, please derive defaults from the existing integration-test setup. Before merge, please provide CI output showing that this test ran and passed rather than being skipped.

sagemaker-core/tests/integ/processing/test_instance_preferences_processing.py
deeppcs: Good catch, the test was written before GA so it was gated like this. Posting fix PR to make it use existing CI resources

sagemaker-train/tests/integ/train/test_instance_preferences.py
deeppcs: Good catch, the test was written before GA so it was gated like this. Posting fix PR to make it use existing CI resources

v3-examples/training-examples/instance-preferences-example.ipynb
```diff
@@ -0,0 +1,225 @@
...
+    "role = get_execution_role()\n",
+    "region = sagemaker_session.boto_region_name\n",
+    "bucket = sagemaker_session.default_bucket()\n",
+    "\n",
+    "training_image = image_uris.retrieve(\n",
+    "    framework=\"pytorch\",\n",
+    "    region=region,\n",
+    "    version=\"2.0.0\",\n",
+    "    py_version=\"py310\",\n",
+    "    instance_type=\"ml.m5.xlarge\",\n",
```
papriwal: Step 1 resolves `training_image` with `instance_type="ml.m5.xlarge"`, which selects the CPU PyTorch image, but the `instance_preferences` in Step 2 are GPU types (`ml.p5.48xlarge` / `ml.p4d.24xlarge`). This example is about capacity-based instance selection and doesn't need to demonstrate CUDA utilization, so the resolver type may not matter functionally. For readers who copy this notebook as a starting point, though, would resolving the image from one of the p-family candidate types make the pairing between the resolved image and the requested GPU candidates less potentially misleading—or is the CPU resolver type here intentional?

papriwal (commented):
One documentation question about the training notebook example.

deeppcs:
All feature-related integ-tests PASSED - 

- `sagemaker-train` - Pass
```
[gw27] [ 91%] PASSED tests/integ/train/test_instance_preferences.py::test_instance_preferences_select_a_winner_and_complete
```

- `sagemaker-core` - Pass
```
[gw7] [ 35%] PASSED tests/integ/processing/test_instance_preferences_processing.py::test_processor_instance_preferences_e2e
```

- `fast-integ-tests` - All Pass
```
[gw6] [  1%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesRejected::test_managed_spot_with_preferences_is_rejected
[gw6] [  2%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesRejected::test_duplicate_types_are_rejected
[gw7] [  3%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesRejected::test_server_rejects_more_than_five_preferences
[gw4] [  6%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesAccepted::test_per_preference_counts_are_accepted_and_echoed
[gw4] [  7%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesRejected::test_instance_type_with_preferences_is_rejected
[gw5] [  8%] PASSED tests/integ/train/shallow/test_instance_preferences.py::TestInstancePreferencesAccepted::test_uniform_count_is_accepted_and_echoed
```

v3-examples/training-examples/instance-preferences-example.ipynb
deeppcs: Yes, if someone copies this verbatim, then it would result in wrong image usage. But the example in this and all other notebooks are geared towards understanding feature usage rather than actual training.

Can fix this in a follow up, I don't think this should be a blocker

deeppcs: Notebooks have also been approved by SA, PM and Doc writers

deeppcs: Notebooks have been fixed in new commit

baiyli (approved):

sagemaker-train/src/sagemaker/train/model_trainer.py
```diff
@@ -1569,6 +1581,11 @@ def from_jumpstart_config(
                 f"Training is not supported for the model ID: {jumpstart_config.model_id}.\n"
                 "Please check that the model ID is available for training."
             )
+        if compute and getattr(compute, "instance_preferences", None):
+            raise ValueError(
+                "JumpStart training does not support ``instance_preferences``. "
+                "Set a single ``instance_type`` in Compute for JumpStart models."
+            )
```
nargokul, line 1588: What about support in other trainers ? SFT, DPO etc also support serverful .

deeppcs: We will not be supporting these model customization based serverful Trainers during the initial launch. They might be covered in a follow up as it also needs some updates on Studio side and recipe rendering logic

Its also explicitly denied in the PR - `Training recipes do not support instance_preferences`

nargokul (approved):

PR #6250, opened 2026-09-10
fix(instance-preferences): run integ tests in CI, reject managed spot client-side, stabilize IntPipeVar codegen

deeppcs:
## Description

Addresses the review feedback on #6249. Targets `feature-smtj-instance-preferences-rebased` so #6249 updates in place.

**Integ tests now run in the standard CI account.** Both tests were gated on account-specific environment variables that no CI configuration sets, so every run skipped them and the suite passed without exercising the feature. Now:

- Training coverage lives in `sagemaker-train/tests/integ/train/shallow/test_instance_preferences.py`, which `fast-integ-tests` runs on every pull request: two accepted cases (uniform and per-preference counts — submit, Describe echo, stop) and four rejected cases (`instance_type`, managed spot, and duplicate types client-side; the five-entry cap server-side). The environment-gated deep test is removed.
- The processing test uses the suite's `SageMakerRole` and resolves its image via `image_uris`, like the other `sagemaker-core` integ tests.
- Candidate types remain overridable via `*_TEST_INSTANCE_TYPES` for accounts whose defaults lack quota. No account-specific values remain in the repository.

**Managed spot is rejected client-side.** The documentation already lists it as mutually exclusive with `instance_preferences`, but `validate_instance_preferences` only checked `instance_type`, `instance_groups`, and `instance_placement_config`. Both copies now reject an explicit `enable_managed_spot_training=True`; `False`/unset stay valid.

**`IntPipeVar` is codegen-stable.** The widening on the five instance-preferences count members was a hand edit that `generate_shapes` could not reproduce — only `ResourceConfig.InstanceCount` was in `PIPE_VAR_OVERRIDES`, so the next regeneration would have narrowed the rest to `Optional[int]`. All five are registered, and a unit test asserts the extractor emits `IntPipeVar` for each, so a dropped override fails loudly instead of silently.

## Testing

In us-west-2, with no environment variables beyond region:

- New shallow tests: 6/6 pass. Under the exact CI command (`pytest tests/integ/train/shallow -n 8 -m "not gpu_intensive and not us_east_1"`) all six pass alongside the existing suite.
- Processing integ test: pass, winner resolved.
- Unit: 196 `sagemaker-core` (compute configs, processing, codec, service model, shapes extractor), 111 `sagemaker-train` (model_trainer, defaults). The new extractor test was mutation-checked: removing one override fails exactly that member's case.

The `fast-integ-tests` run on this PR is the CI execution proof requested on #6249.


deeppcs:
Superseded by the master-targeted PR carrying the same two commits, since PRs into feature branches do not trigger the integration tests.

deeppcs:
Superseding PR: #6251.

deeppcs:
Reopened — this remains the PR to merge into the feature branch. #6251 exists only to run the master CI suite against the same tree; it is not intended to merge.

sagemaker-core/tests/integ/processing/test_instance_preferences_processing.py
```diff
@@ -56,35 +54,24 @@
...
+    )
 
 
 def test_processor_instance_preferences_e2e():
     """Create a real processing job with InstancePreferences and verify the winner."""
     from sagemaker.core.helper.session_helper import Session
     from sagemaker.core.processing import Processor
 
-    client = _sagemaker_client()
+    client = boto3.client("sagemaker", region_name=REGION)
```
papriwal, line 84: This test now launches a real processing job on every CI run but never stops it. The wait loop breaks as soon as the job is `InProgress` with a `SelectedInstanceType`, so the job keeps running to its 1,800-second `max_runtime_in_seconds` limit—consuming quota and risking `ResourceLimitExceeded` when runs overlap.

Please stop the job in guaranteed cleanup (for example, a `try/finally` calling `client.stop_processing_job`) after the contract assertions, tolerating a job that is already terminal or was never created. This would mirror the submit-then-stop lifecycle used by the training coverage.

sagemaker-train/tests/integ/train/shallow/test_instance_preferences.py
```diff
@@ -0,0 +1,166 @@
...
+        ]
+        trainer = _trainer(
+            sagemaker_session,
+            unique_name("shallow-ip-six"),
+            Compute(
+                instance_preferences=[InstancePreference(instance_type=t) for t in six],
+                instance_count=1,
+            ),
+        )
+        assert_rejected(trainer, ("InstancePreferences", "5", "Member must have length"))
```
papriwal: The `"5"` token lets this negative test pass for the wrong reason. `assert_rejected` succeeds when *any* token is a substring of the error message (`any(token in message for token in expected_tokens)`), so an unrelated rejection mentioning an instance type such as `ml.m5.large` would satisfy `"5"`. The other two tokens already pin the intended five-entry-cap error. Please drop the broad token:

```python
assert_rejected(
    trainer,
    ("InstancePreferences", "Member must have length"),
)
```

sagemaker-core/tests/integ/processing/test_instance_preferences_processing.py
deeppcs: Thanks for flagging this, fixed in new commit

sagemaker-train/tests/integ/train/shallow/test_instance_preferences.py
deeppcs: Thanks for flagging this, fixed in new commit

jam-jee (approved):

papriwal (approved):

papriwal (commented):
Merging this one, will check if the CI passes in the feature branch's PR.

PR #6264, opened 2026-09-11
fix: restore model customization reuse state

papriwal:
*Issue #, if available:*

*Description of changes:*

## What problem does this solve?

`ModelBuilder` performs Model reuse and endpoint reuse independently. When `build(reuse_resources=True)` found an existing package-backed Model but deployment could not reuse an endpoint, the SDK continued through the endpoint-creation path without rebuilding the Model.

Some deployment state normally populated while building a customization Model—such as recipe-derived compute requirements and the LoRA adapter artifact location—was therefore unavailable. This could cause the fallback deployment to fail or reach resource creation without the required configuration.

The existing integration coverage could also discover unrelated resources left by another test run. That made the reuse scenario non-deterministic and could hide regressions that created duplicate resources.

## Why is this change needed?

Reusing a Model should not require its endpoint to be reusable. If endpoint reuse misses, `ModelBuilder` must create the endpoint using the same customization metadata that a fresh build would provide, without creating a second Model.

Explicitly supplied resource requirements must remain authoritative. Cached recipe settings should only be restored when the caller did not provide requirements, while LoRA deployments must still have a valid adapter artifact location.

The integration scenario also needs to prove that the exact Model and endpoint created by the test are reused, rather than accepting arbitrary resources from a shared test environment.

## What changed?

- Track whether the current build reused an existing Model and reset that state between build attempts.
- Restore package-backed customization deployment state only when a reused Model must enter the endpoint-creation path.
- Preserve explicitly supplied resource requirements and copy count.
- Resolve LoRA adapter artifacts consistently for supported training-job, trainer, and model-package sources.
- Fail before endpoint resource creation when required customization state cannot be resolved.
- Keep healthy endpoint reuse on the existing fast path without unnecessary state restoration.
- Consolidate the scheduled reuse integration coverage around uniquely named, test-owned resources.
- Require exact Model and endpoint discovery before exercising the public reuse flow.
- Guard the integration scenario against unexpected duplicate Model or endpoint creation and clean up only resources owned by that test.

This change does not add or modify public API parameters and introduces no new dependencies.

## Testing

- Added unit coverage for reused-Model marker lifecycle and reset behavior.
- Added coverage for exact Model and endpoint reuse without state preparation.
- Added LoRA and non-LoRA coverage for endpoint creation after Model reuse.
- Added coverage for preserving explicit resource requirements and copy count.
- Added coverage for supported LoRA artifact paths and unsupported source rejection before writes.
- Ran the focused ModelBuilder, inference-configuration, model-reuse, recipe-selection, and instance-selection unit suites.
- Verified Python compilation and diff whitespace checks.
- Verified integration-test collection, scheduled-test selection, and removal of the non-hermetic reuse tests.


zhaoqizqwang (approved):

sagemaker-serve/tests/integ/test_model_customization_deployment.py
```diff
@@ -108,92 +116,208 @@ def test_build_from_training_job(self, training_job_name, sagemaker_session):
...
-        assert endpoint.endpoint_arn is not None
-        assert endpoint.endpoint_status == "InService"
-
-        # Verify model-source tag is present on the endpoint for reuse discovery.
-        sm_client = boto3.client("sagemaker", region_name=AWS_REGION)
-        endpoint_tags = sm_client.list_tags(ResourceArn=endpoint.endpoint_arn).get("Tags", [])
-        assert MODEL_SOURCE_TAG_KEY in {t["Key"] for t in endpoint_tags}, (
-            f"Endpoint {endpoint.endpoint_arn} missing model-source tag for reuse"
-        )
+            with patch.object(
```
aviruthen, line 143: Should we be patching this method on integ tests?

sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -4328,11 +4383,11 @@ def build(
                 configuration chain. (Default: None).
             region (str, optional): The AWS region for deployment. If specified and different
                 from the current region, a new session will be created. (Default: None).
-            reuse_resources (bool, optional): If True, checks for an existing endpoint built
-                from the same model source (with matching deployment configuration) before
-                creating anything. On a match, build() creates no new resources and sets
-                ``built_model`` to the existing Model backing that endpoint; the subsequent
-                deploy() returns the existing endpoint. (Default: False).
+            reuse_resources (bool, optional): If True, checks for an existing Model built
```
aviruthen, line 4386: At a high level, it makes sense to reuse Model with build and reuse Endpoint with deploy. Just want to confirm that this is new behavior and should be listed as a feature...? To some extent this is a breaking change on the reuse_resources parameter since it was documented incorrectly

sagemaker-serve/tests/integ/test_model_customization_deployment.py
papriwal: Yes, this patch is intentional to isolate the reuse identity for this test. The training-job fixture is shared, so using its natural source identity can discover resources from a previous or concurrent run. We patch only source identity generation; Model and endpoint creation, persisted tags, resource discovery, the public reuse calls, and invocation still use the real implementations. This keeps the integration scenario deterministic without deleting or accepting resources owned by another run.

sagemaker-serve/src/sagemaker/serve/model_builder.py
papriwal: Confirmed that independent Model and endpoint reuse is existing behavior, not a new feature. Before this PR, build(reuse_resources=True) already called _find_reusable_model(), and the implementation explicitly stated that endpoint reuse was resolved separately by deploy(). The previous parameter docstring incorrectly described build as performing endpoint lookup. This PR fixes the missing customization state exposed by the existing Model-hit/endpoint-miss lifecycle and aligns the docstring with the actual contract. It does not change the reuse API or split previously coupled runtime behavior, so fix: is the appropriate classification.

aviruthen: I see, so it was a docstring issue. Thanks for the clarification!

sagemaker-serve/tests/integ/test_model_customization_deployment.py
aviruthen: Makes sense!

aviruthen (approved):

Issue #5986, opened 2026-07-05
SageMakerClient ignores the passed boto3 session for the "sagemaker" client since 2.13.0 — all resource API calls sign with default-chain credentials


brifordwylie:
**TL;DR** — Since sagemaker-core 2.13.0, `SageMakerClient` ignores the `session=` the caller passed and builds its main `sagemaker` client from a fresh default-chain session. Every V3 resource API call (`FeatureGroup.create`, `Model.create`, `describe`/`delete`/`list`, …) is silently signed with ambient credentials instead of the caller's. 

Besides breaking explicit-session workflows, this is a quiet identity mix-up with a security edge: code written to run under a scoped or assumed role can instead execute under a less-privileged (or worse a more-privileged) ambient identity, with no error or warning. Last good version: 2.12.0. Workaround: pin `sagemaker-core>=2.12,<2.13`.

**PySDK Version**
- [ ] PySDK V2 (2.x)
- [x] PySDK V3 (3.x)

**Describe the bug**
`SageMakerClient.__init__` (`sagemaker-core/src/sagemaker/core/utils/utils.py`) contains a block introduced by #5919 (merged 2026-06-03, released as sagemaker-core 2.13.0) that loads a custom service model for pre-GA Job APIs through a brand-new botocore session:

```python
# TODO: Remove post-launch. This loads a custom botocore service model
# (from the 'sample/' directory) that includes pre-GA Job APIs ...
bc_session = bc_session_mod.get_session()
...
custom_session = Session(botocore_session=bc_session, region_name=env_region)

self.sagemaker_client = custom_session.client("sagemaker", ...)   # <-- passed `session` ignored
```

`custom_session` resolves credentials from the default provider chain (env vars / `AWS_PROFILE` / IMDS), discarding the passed session's credentials. The other clients built in the same constructor (`sagemaker-runtime`, `sagemaker-featurestore-runtime`, `sagemaker-metrics`) still correctly use the passed `session`.

Since every resource class routes through `Base.get_sagemaker_client(session=...)` → `SageMakerClient(...).get_client("sagemaker")`, all V3 resource API calls are affected.

**To reproduce**
Minimal proof, no AWS resources created — the returned client does not hold the passed session's credentials:

```python
import boto3
from sagemaker.core.utils.utils import SageMakerClient

# Any session whose credentials differ from the default provider chain,
# e.g. an assumed role:
sts = boto3.client("sts")
creds = sts.assume_role(
[...]

PR #6273, opened 2026-09-14
fix: SageMakerClient honors the passed boto3 session and config

jam-jee:
## Problem

`SageMakerClient` in `sagemaker.core.utils.utils` was a process-wide singleton. The first instance created was returned to every later caller and the `session=` passed on those later calls was silently ignored. If anything built a client with default-chain credentials before the user's own session was seen (a bare `SageMakerClient()` call inside the SDK, a second profile, an assumed role), every subsequent control-plane call was signed with the wrong identity. That is the remaining half of #5986 after the #5919 regression was reverted, and it is what produces the cross-account `ValidationException: RoleArn: Cross-account pass role is not allowed` in #6069 when `ModelTrainer(sagemaker_session=...)` runs under a profile that differs from the default.

Separately, the `config` argument was computed and then thrown away (`self.config` was reassigned to a bare user-agent `Config`), so caller timeouts and retry settings never reached the boto3 clients.

## Why it matters

Any multi-account or assumed-role workflow on V3 is affected, and it fails silently: calls succeed under an identity the caller did not choose. This is a credential mix-up rather than a plain bug, which is why #5986 was filed as a security issue. The `config` loss means callers who construct `SageMakerClient(config=...)` directly could not tune retries or timeouts; note that the generated resource layer's `Base.get_sagemaker_client()` does not forward a `config` yet, so reaching it from resource calls is a separate template change.

## Fix (symptoms -> root cause -> change)

Symptom: a client for session A is handed to a caller who passed session B. Root cause: `SingletonMeta.__call__` keys the cache on the class alone, so arguments are only consulted on the very first construction. Change: give `SageMakerClient` its own metaclass, `_ClientCacheMeta`, that keys cached instances on the constructor arguments (session identity, `region_name`, config identity).

- An explicit `session` (or `region_name` / `config`) always yields a client built from exactly those arguments. Sessions and configs are keyed by identity because credentials live on the session object; the cached instance retains both, so the ids cannot be recycled while the entry is alive.
- `SageMakerClient()` with no arguments still returns a single process default, and the first instance created in the process still becomes that default. This keeps the configure-once pattern (`SageMakerClient(session=...)` up front, bare calls afterwards) and the generated `stop()` methods working unchanged.
- Keyed entries are a small LRU (32) so a service that mints a fresh session per request does not retain every session and its four boto3 clients forever. The default entry is never evicted. Construction runs under a lock so concurrent first callers share one instance.
- The caller's `config` is now merged into the client config, with the SDK user-agent suffix appended to any `user_agent_extra` the caller set rather than replacing it.
- `SingletonMeta` is untouched; `CloudWatchLogsClient` in `utils/logs.py` still uses it.

Not changed here: `HubContent.load()` in the generated `resources.py` calls `SageMakerClient(..., service_name="sagemaker").client`, which is broken on master too (`__init__` never accepted `service_name`, and the attribute is `sagemaker_client`). It fails on both master and this branch; the correct pattern `Base.get_sagemaker_client(...)` is used everywhere else. Fixing the template is a separate change.

## Tests

New `sagemaker-core/tests/unit/test_sagemaker_client.py` (13 tests, all offline with static-credential `boto3.Session` objects):

- `test_explicit_session_is_honored_after_another_client_exists`: the #5986/#6069 regression. Two sessions with different access keys; each of the four service clients on the second instance signs with the second session's key. Fails on master.
- `test_same_session_returns_cached_instance`, `test_same_session_different_region_gets_distinct_client`, `test_distinct_config_objects_get_distinct_clients`: cache key semantics.
- `test_bare_call_returns_first_created_instance_as_default`, `test_bare_call_default_is_not_replaced_by_later_explicit_session`, `test_bare_call_creates_default_from_default_chain`: the default-instance contract.
- `test_config_argument_is_applied_and_user_agent_suffix_appended`, `test_default_config_keeps_retry_settings`: the `config` argument reaches the clients.
- `test_keyed_cache_is_bounded_and_default_survives_eviction`, `test_recently_used_entry_is_kept_over_stale_ones`: LRU bound and ordering (both fail if eviction or the LRU touch is removed).
- `test_concurrent_first_calls_share_one_instance`: 8 threads racing the first construction get one instance.
- `test_reset_clears_default_and_keyed_entries`.

Existing suites run locally: `tests/unit/generated`, `tests/unit/session`, `tests/unit/helper` (383 passed, 6 skipped) plus `tests/unit/test_common_utils.py`; the tests that patch `sagemaker.core.utils.utils.SageMakerClient` replace the class outright and are unaffected by the metaclass change.

## Manual verification

N/A: the failure mode is fully reproducible offline (credential on the request signer), and the new unit tests assert on exactly that. No AWS calls are needed to observe or verify the fix.

---
X-AI-Prompt: Fix the SageMakerClient singleton so an explicitly passed boto3 session is honored (#5986/#6069)
X-AI-Tool: Kiro


brifordwylie:
@jam-jee thanks so much for looking into this issue, I'm sure your solution is better/broader but can you take a quick peek at my PR just to make sure we don't miss anything: https://github.com/aws/sagemaker-python-sdk/pull/6017/changes. (EDIT) Okay, I see now.. https://github.com/aws/sagemaker-python-sdk/pull/5919 was reverted and part of my PR was removing parts of that.

Thanks again for being responsive to this, glad to see this one getting addressed. :)

rsareddy0329 (approved):

PR #6275, opened 2026-09-15
fix(train): enforce S3 ownership on ai_registry default bucket

rsareddy0329:
The ai_registry and finetune_utils modules derive a predictable default bucket name (sagemaker-{region}-{account_id}) and read from / write to it without verifying ownership, unlike sagemaker-core which enforces ExpectedBucketOwner. Because S3 bucket names are globally unique, another account could pre-create that name and the SDK would silently use it.

Add the ownership guard for the SDK-derived default bucket (explicitly provided buckets are left untouched):

- air_hub.upload_to_s3 / download_from_s3 pass ExpectedBucketOwner via ExtraArgs when the target is the derived default bucket.
- _create_mlflow_app (and _create_mlflow_app_as_upgrade) verify ownership and pass ExpectedBucketOwner on the S3 probe/writes before registering the MLflow ArtifactStoreUri.
- _validate_s3_path_exists verifies ownership of the derived default bucket before it is created or used as OutputDataConfig.

Add unit tests asserting the guard is applied for the default bucket, left off for explicit buckets, and fails safe on a foreign-owned bucket.

*Issue #, if available:*

*Description of changes:*


sagemaker-train/src/sagemaker/train/common_utils/finetune_utils.py
```diff
@@ -422,6 +423,42 @@ def _wait_for_mlflow_app_ready_boto(sm_client, arn: str, timeout: int = 600) ->
...
+    """
+    if bucket_name != _default_bucket_name(region, account_id):
+        return
+    try:
+        s3_client.head_bucket(Bucket=bucket_name, ExpectedBucketOwner=account_id)
+    except ClientError as e:
+        error_code = str(e.response.get("Error", {}).get("Code", ""))
+        if error_code in ("404", "NoSuchBucket", "NotFound"):
+            return  # Bucket does not exist yet; the caller creates it in-account.
+        if error_code in ("403", "AccessDenied"):
```
papriwal, line 453: **Suggestion (non-blocking):** Consider making this error message neutral and omitting the account ID. A 403 from `HeadBucket(..., ExpectedBucketOwner=...)` can indicate either an ownership mismatch or missing `s3:ListBucket` permission, so it does not establish who owns the bucket. For example:

```python
raise ValueError(
    f"Refusing to use default bucket '{bucket_name}': ownership or access "
    f"could not be verified. Confirm that the bucket belongs to your AWS "
    f"account and that the caller has s3:ListBucket permission, or provide "
    f"an explicit bucket you own."
)
```

This avoids exposing account identifiers or directing users toward an ownership diagnosis that may be incorrect.

papriwal (approved):

jam-jee:
nit : Can we add a integ test in `sagemaker-train/tests/integ/ai_registry/test_air_hub.py` covering these changes. (positive and negative)

jam-jee (approved):

PR #6279, opened 2026-09-15
Fix LineIterator hanging forever when the stream ends without a trailing newline

### LINKED ISSUE #6278: LineIterator hangs forever if the streamed response ends without a trailing newline
**`LineIterator.__next__` hangs forever if the last chunk of a streamed response has no trailing newline.**

`sagemaker-core/src/sagemaker/core/iterators.py`:

```python
except StopIteration:
    if self.read_pos < self.buffer.getbuffer().nbytes:
        continue
    raise
```

`continue` re-reads the same unterminated line, fails the same check, calls `next()` on the now-exhausted `byte_iterator` again, gets `StopIteration` again, forever. Nothing in that branch changes `read_pos` or the buffer.

Repro:

```python
from sagemaker.core.iterators import LineIterator

chunks = [
    {"PayloadPart": {"Bytes": b'{"outputs": [" a"]}\n'}},
    {"PayloadPart": {"Bytes": b'{"outputs": [" b"]}'}},
]

for line in LineIterator(iter(chunks)):
    print(line)
```

Prints the first line then hangs.

Will open a PR.

RudraDudhat2509:
Fixes #6278

Returns the leftover partial line once instead of `continue`, so the next call correctly hits `read_pos == buffer size` and raises `StopIteration` normally instead of looping forever.

Added `test_line_iterator_no_trailing_newline_at_end`. `test_line_iterator_incomplete_line_at_end` doesn't actually cover this, it only tests a stream that already ends with `\n`.

Verified against the pre-fix code: hangs, confirmed by killing it after 15s with no output. Full `test_iterators.py` 21 passed after the fix.


RudraDudhat2509:
Gentle ping

PR #6281, opened 2026-09-16
Checks to use the correct ami when on cu130 to avoid crashes at launch

dwarez:
*Description of changes:*
Currently, some instances crash at startup because of a mismatch between the CUDA version of the engine container and the one present in the instance. This will result in a run that crashes even before any logs can be produced. Currently the fix is to specify the correct `inference_ami_version` when invoking `ModelBuilder.deploy`, however this can be managed by the SDK and should not be a concern of the user.


sagemaker-serve/src/sagemaker/serve/model_builder.py
```diff
@@ -4188,6 +4188,11 @@ def _deploy(self, **kwargs):
         if self.mode == Mode.LOCAL_CONTAINER:
             endpoint = self._deploy_local_endpoint(**kwargs)
         elif self.mode == Mode.SAGEMAKER_ENDPOINT:
+            if "container_timeout_in_seconds" in kwargs:
```
rsareddy0329: is this change required for AMI fix?

dwarez: no not really, reverting it and in case we can make a different PR for that

sagemaker-serve/src/sagemaker/serve/model_builder_servers.py
```diff
@@ -358,16 +372,28 @@ def _build_for_hf_server(self, model_server: ModelServer) -> Model:
...
+                self.env_vars.setdefault("SM_VLLM_MODEL", self.model)
 
             if self.env_vars.get("HUGGING_FACE_HUB_TOKEN"):
                 self.env_vars["HF_TOKEN"] = self.env_vars.get("HUGGING_FACE_HUB_TOKEN")
 
             self.s3_upload_path = None
 
         self._auto_detect_image_uri()
+        if (
+            model_server == ModelServer.VLLM
```
ehcalabres: I think we can safely remove the vLLM check and apply this by default to any `image_uri` and `instance_type` combination, regarding the inference engine used, as it will fix all of them when no AMI is specified

sagemaker-serve/tests/unit/test_model_builder_servers_hf_model_id.py
```diff
@@ -292,6 +292,59 @@ def test_sets_default_hf_model_id_when_not_provided(
     )
 
 
+class TestBuildForVllmCompatibility:
```
ehcalabres: We can change this test name to make it independent from the inference engine used, as the fix won't be vLLM specific

ehcalabres (approved):
LGTM now! 🚀

rsareddy0329 (approved):

PR #6286, opened 2026-09-17
Feat: finetuningoptions validation telemetry

rsareddy0329:
*Issue #, if available:*
* Instrument FineTuningOptions validation failures.

*Description of changes:*
Instruments FineTuningOptions.__setattr__ so client-side hyperparameter validation failures are captured in the SDK dev-experience logs, enabling detection/measurement of this failure class:
- setting an option name that isn't a valid/overridable hyperparameter (AttributeError), and
- setting a value outside its spec — type/range/enum (ValueError).

Emission is failure-only via sagemaker.core.telemetry helper _emit_failure_telemetry: a single MODEL_CUSTOMIZATION FAILURE event is sent on the failure path, success path would not be captured as it records as multiple for one trainer call(per number of overridable params in recipe). The helper also adds the TelemetryOptOut config, mirrors the decorator's env/platform payload fields.


jam-jee (dismissed):

jam-jee:
nit: can you add details to PR description.

jam-jee (approved):

PR #6289, opened 2026-09-17
change: update image_uri_config for DLC frameworks and me-west-1 coverage

Yadan-Wei:
Promotes the latest generated image_uri_config to master.

Additive change across 13 config files — no version, version_alias, or per-version registry entry present in master is removed (verified).

- New serving-framework versions: vllm (1→8), sglang (1→8), vllm-server (2→6), vllm-omni (2→5), sglang-server (2→5), ray-serve (2→4), llama-cpp / llama-cpp-arm64 (2→3), whisperx (1→2).
- Adds me-west-1 region coverage to in-window framework versions (pytorch, pytorch-amzn2023, tensorflow, and the serving frameworks above); older out-of-window versions are unchanged.

The line-level deletions in the diff are JSON key re-sorting only, not content removal.


Yadan-Wei:
Superseded by https://github.com/aws/sagemaker-python-sdk/pull/6290, which includes the same config promotion plus the whisperx alias fix and the image_uris test alignment needed for CI to pass.

